PR #55, opened 2026-02-05
feat: Add ArgoCD Application manifest for nebari-operator deployment

viniciusdc:
## Summary

Adds ArgoCD Application manifest to deploy the NebariApp Operator (formerly NIC Operator) as part of the Nebari foundational infrastructure. This enables GitOps-based deployment and management of the operator across Nebari clusters.

### Verify ArgoCD Application

```bash
# Check Application status
kubectl get application nebari-operator -n argocd

# View sync status
argocd app get nebari-operator

# Check Application health
argocd app status nebari-operator
```

### Verify Operator Deployment

```bash
# Check operator pods
kubectl get pods -n nebari-operator-system

# Check operator logs
kubectl logs -n nebari-operator-system -l control-plane=controller-manager -f

# Verify CRDs installed
kubectl get crd nebariapps.reconcilers.nebari.dev

# Test NebariApp creation
kubectl apply -f - <<EOF
apiVersion: reconcilers.nebari.dev/v1
kind: NebariApp
metadata:
  name: test-app
  namespace: default
spec:
  hostname: test.nebari.local
  service:
    name: test-service
    port: 8080
EOF
```

### Verify Metrics

```bash
# Port-forward to metrics endpoint
kubectl port-forward -n nebari-operator-system \
  deployment/nebari-operator-controller-manager 8080:8080

# Check metrics
curl http://localhost:8080/metrics
```

### DIFF
--- a/pkg/argocd/templates/manifests/nebari-operator/kustomization.yaml
+++ b/pkg/argocd/templates/manifests/nebari-operator/kustomization.yaml
@@ -4,11 +4,11 @@ kind: Kustomization
 namespace: nebari-operator-system
 
 resources:
-  - https://github.com/nebari-dev/nebari-operator.git/config/default?ref=v0.1.0-alpha.1
+  - https://github.com/nebari-dev/nebari-operator.git/config/default?ref=v0.1.0-alpha.2
 
 images:
   - name: quay.io/nebari/nebari-operator
-    newTag: 0.1.0-alpha.1
+    newTag: 0.1.0-alpha.2
 
 patches:
   - path: deployment-patch.yaml


dcmcand:
Note: the kustomization.yaml resource URL needed to be corrected. Two issues with the original:

1. **`dist/install.yaml` doesn't exist** in the nebari-operator repo — the operator uses standard kubebuilder layout with manifests under `config/default/`
2. **`raw.githubusercontent.com` URLs don't work as kustomize resources** — kustomize tries to `git fetch` from them and fails

The correct kustomize remote resource format is:
```yaml
resources:
  - https://github.com/nebari-dev/nebari-operator.git/config/default?ref=main
```

instead of:
```yaml
resources:
  - https://raw.githubusercontent.com/nebari-dev/nebari-operator/main/dist/install.yaml
```

pkg/argocd/templates/manifests/nebari-operator/deployment-patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+kind: Deployment
+metadata:
+  name: nebari-operator-controller-manager
+  namespace: nebari-operator-system
+spec:
+  template:
+    spec:
+      containers:
+        - name: manager
+          image: quay.io/nebari/nebari-operator:v0.1.0-alpha.1
```
dcmcand: The image tag v0.1.0-alpha.1 appears in both the kustomization.yaml resource URL and the deployment patch. When upgrading, both must be updated together. A more idiomatic approach would be to use kustomize's images transformer in kustomization.yaml:
```
  images:
    - name: quay.io/nebari/nebari-operator
      newTag: v0.1.0-alpha.1
```
  This centralizes version management and avoids the deployment patch needing to carry the image at all.

tylerpotts: centralized as you suggested

PR #104, opened 2026-02-16
Write tofu binary to working directory instead of shared cache

### LINKED ISSUE #77: Cannot run multiple nic deployments in parallel from the same machine
## Description

When running two or more deployments from the same machine, only one of them will run successfully while the others will throw an error like this:
```  
{"time":"2026-02-05T19:09:24.017507+01:00","level":"ERROR","msg":"Command execution failed","error":"signal: killed\n"}
```

I suspect the root of the issue is that currently, each deployment will try to overwrite the binary, even though it is being used while other deployment:
https://github.com/nebari-dev/nebari-infrastructure-core/blob/ee6536aa6d74696d1a04bb1ffe07b744b772d050/pkg/tofu/tofu.go#L111-L126

## Use Case

Although probably not a very common scenario, two or more deployments could be ran from the same machine.

marcelovilla:
## Closes

Closes #77

## Description

Moved the extracted OpenTofu binary from a shared cache path (`~/.cache/nic/tofu/tofu`) to the per-deployment temporary working directory. This prevents race conditions and version conflicts when multiple deployments run concurrently or when different NIC versions pin different OpenTofu versions. Downloaded archives are still cached in `~/.cache/nic/tofu/` via tofudl.Mirror, so there's no redundant network downloads. 

Also improves error cleanup by using two defers—one immediately after extracting the templates and one after setting up the cache directories—to ensure no resources leak on failure.

## How to Test

Run multiple `nic` deployments with the `--dry-run` flag locally and check if it succeeds.

### DIFF
--- a/pkg/tofu/tofu.go
+++ b/pkg/tofu/tofu.go
@@ -261,7 +261,7 @@ func Setup(ctx context.Context, templates fs.FS, tfvars any) (te *TerraformExecu
 		}
 	}()
 
-	downloader := &tofuDownloader{cacheDir: cacheDir, version: DefaultVersion}
+	downloader := &tofuDownloader{cacheDir: cacheDir, version: Version}
 	execPath, err := downloadExecutable(ctx, appFs, workingDir, downloader)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get executable: %w", err)


dcmcand:
@marcelovilla please fix merge conflicts so I can approve

PR #120, opened 2026-02-20
Rename DefaultVersion to TofuVersion for clarity

marcelovilla:
## Description

This is a very small PR renaming the `DefaultVersion` variable in the `tofu` package to `TofuVersion`. This makes it clearer as we're not supporting overriding this variable in any way and our intent is to keep it pinned to a specific version to ensure reproducibility. Thus, the name `DefaultVersion` is misleading.

## How to Test


dcmcand (approved):
🚀

PR #122, opened 2026-02-20
Fix double SIGINT on Ctrl+C by wrapping tofu context per platform

### LINKED ISSUE #114: Ctrl+C during deploy/destroy causes double SIGINT to tofu, risking state drift
## Description

When the user presses Ctrl+C during `nic deploy` or `nic destroy`, tofu receives two SIGINT signals instead of one. This causes tofu to abort immediately instead of shutting down gracefully, which can leave the state file out of sync with actual cloud resources.

The first SIGINT is delivered by the OS to the entire process group (both the Go parent and the tofu child). The second comes from terraform-exec: Go's signal handler cancels the context, terraform-exec sees the cancelled context and sends another signal to the child process. Tofu interprets two SIGINTs as "abort immediately," skipping its graceful shutdown (finish in-flight operations, save state).

The result might be state drift—resources may exist in the cloud but not in state, or vice versa, requiring manual `tofu import` or cleanup.

## Steps to Reproduce

1. Run `nic deploy -f config.yaml` (or `nic destroy`)
2. Wait for tofu to start applying/destroying resources
3. Press Ctrl+C once
4. Observe tofu aborts immediately instead of finishing the in-flight operation

## Error Output

```
^C

{"time":"2026-02-19T09:49:36.218626+01:00","level":"WARN","msg":"Received interrupt signal, initiating graceful shutdown","signal":"interrupt"}

Interrupt received.
Please wait for OpenTofu to exit or data loss may occur.
Gracefully shutting down...

Stopping operation...

Two interrupts received. Exiting immediately. Note that data loss may have
occurred.
```
## Environment

- OpenTofu 1.11.2

marcelovilla:
## Closes

Closes #114

## Description

On macOS, tofu shares the parent's process group, so Ctrl+C delivers SIGINT to both processes. terraform-exec then sends a second SIGINT when it observes the cancelled context, causing tofu to abort immediately instead of shutting down gracefully (risking state drift). On Linux this is not an issue because terraform-exec sets `Setpgid: true`, placing tofu in its own process group isolated from the terminal.

This PR introduces `signalSafeContext` in the `tofu` package, compiled per platform via build tags (`context_linux.go` / `context_default.go`), and wraps it into `TerraformExecutor`'s methods (`Init`, `Plan`, `Apply`, `Destroy`, `Output`) so the fix is applied automatically and callers don't need to handle it.

## How to Test

From the `main` branch, run either `nic destroy` or `nic deploy` and cancel the command once tofu is running by pressing Ctrl+C. If running on macOS, you'll see that two interrupts were received:

```
^C{"time":"2026-02-19T10:50:14.294058+01:00","level":"WARN","msg":"Received interrupt signal, initiating graceful shutdown","signal":"interrupt"}
Stopping operation...

Interrupt received.
Please wait for OpenTofu to exit or data loss may occur.
Gracefully shutting down...


Two interrupts received. Exiting immediately. Note that data loss may have 
occurred.
```

Then from this branch (don't forget to rebuild), run the same command and cancel it again once tofu is running by pressing Ctrl+C. The output should now show that OpenTofu shut down gracefully with a single interrupt:

```
^C{"time":"...","level":"WARN","msg":"Received interrupt signal, initiating graceful shutdown","signal":"interrupt"}

Interrupt received.
Please wait for OpenTofu to exit or data loss may occur.
Gracefully shutting down...
```

I tested this on both macOS and a Linux VM and it works as expected. On Linux, the behavior is unchanged since terraform-exec already isolates the process group, but on macOS it now shuts down gracefully instead of aborting immediately.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,8 @@ go 1.25.1
 require (
 	github.com/aws/aws-sdk-go-v2 v1.41.1
 	github.com/aws/aws-sdk-go-v2/config v1.31.20
+	github.com/aws/aws-sdk-go-v2/service/ec2 v1.286.0
+	github.com/aws/aws-sdk-go-v2/service/elasticloadbalancing v1.33.19
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.95.1
 	github.com/aws/aws-sdk-go-v2/service/sts v1.40.2
 	github.com/cloudflare/cloudflare-go/v4 v4.6.0
--- a/go.sum
+++ b/go.sum
@@ -56,6 +56,10 @@ github.com/aws/aws-sdk-go-v2/internal/ini v1.8.4 h1:WKuaxf++XKWlHWu9ECbMlha8WOEG
 github.com/aws/aws-sdk-go-v2/internal/ini v1.8.4/go.mod h1:ZWy7j6v1vWGmPReu0iSGvRiise4YI5SkR3OHKTZ6Wuc=
 github.com/aws/aws-sdk-go-v2/internal/v4a v1.4.17 h1:JqcdRG//czea7Ppjb+g/n4o8i/R50aTBHkA7vu0lK+k=
 github.com/aws/aws-sdk-go-v2/internal/v4a v1.4.17/go.mod h1:CO+WeGmIdj/MlPel2KwID9Gt7CNq4M65HUfBW97liM0=
+github.com/aws/aws-sdk-go-v2/service/ec2 v1.286.0 h1:GgLc+o2oD2sXxlEwGUCCWz/1v3Wa8dN9RRebcIFXeOo=
+github.com/aws/aws-sdk-go-v2/service/ec2 v1.286.0/go.mod h1:Uy+C+Sc58jozdoL1McQr8bDsEvNFx+/nBY+vpO1HVUY=
+github.com/aws/aws-sdk-go-v2/service/elasticloadbalancing v1.33.19 h1:ybEda2mkkX2o8NadXZBtcO9tgmW9cTQgeVSjypNsAy0=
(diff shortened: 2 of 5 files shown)


pkg/tofu/tofu.go
```diff
@@ -30,6 +30,31 @@ func (te *TerraformExecutor) Cleanup() error {
...
+	return te.Terraform.Apply(signalSafeContext(ctx), opts...)
+}
+
+// Destroy wraps tfexec.Terraform.Destroy with a signal-safe context.
+func (te *TerraformExecutor) Destroy(ctx context.Context, opts ...tfexec.DestroyOption) error {
+	return te.Terraform.Destroy(signalSafeContext(ctx), opts...)
+}
+
+// Output wraps tfexec.Terraform.Output with a signal-safe context.
+func (te *TerraformExecutor) Output(ctx context.Context) (map[string]tfexec.OutputMeta, error) {
```
dcmcand: Output wrapper drops the variadic `opts ...tfexec.OutputOption` parameter.
             The upstream signature is:
```go
func (tf *Terraform) Output(ctx context.Context, opts ...OutputOption) (map[string]OutputMeta, error)
```
The wrapper is:
```go
func (te *TerraformExecutor) Output(ctx context.Context)(map[string]tfexec.OutputMeta, error)
```
Since TerraformExecutor embeds *tfexec.Terraform, this shadows the promoted method.
If a future caller tries te.Output(ctx, someOption), it won't compile.Suggestion:
```go
func (te *TerraformExecutor) Output(ctx context.Context, opts ...tfexec.OutputOption) (map[string]tfexec.OutputMeta, error) {
                   return te.Terraform.Output(signalSafeContext(ctx), opts...)
}
```

No current callers pass options, so this is non-blocking, but it's a trivial fix to preserve the full API.

PR #136, opened 2026-03-04
Add local filesystem GitOps support for Kind development

tylerpotts:
## Summary

- Add `file://` URL support so ArgoCD can read manifests from local directories instead of requiring a remote git repo
- Auto-mount local gitops directories into Kind clusters (both the node and the ArgoCD repo-server pod)
- Smart `localkind-up` Makefile target that parses the config YAML and handles three modes: explicit `file://` repo (mount), auto-generated `/tmp` directory (mount), or remote git repo (no mount)
- Write the NIC config file (`nic-config.yaml`) into the gitops repository on every deploy
- Fix cert-manager CrashLoopBackOff by swapping sync waves so envoy-gateway (which provides Gateway API CRDs) installs before cert-manager

### Git config (`pkg/git/config.go`, `pkg/git/client_impl.go`)
- `IsLocalPath()` / `GetLocalPath()` methods to detect and extract `file://` URLs
- `Validate()` skips auth checks for local paths, validates directory exists instead
- `NewClient()` uses local path directly instead of creating temp directories
- `CommitAndPush()` commits locally without pushing for `file://` repos
- `Cleanup()` never deletes local paths (only temp dirs)

### Deploy workflow (`cmd/nic/deploy.go`)
- `getOrCreateGitConfig()` auto-creates `/tmp/nebari-gitops-{project_name}` when no `git_repository` is configured
- Config YAML is written to the gitops repo as `nic-config.yaml` on every deploy

### ArgoCD Helm values (`pkg/argocd/install.go`)
- `Install()` accepts resolved `gitConfig` parameter (supports auto-generated configs)
- `addLocalGitopsMount()` injects `hostPath` volume + volumeMount into the repo-server pod via Helm values, so ArgoCD can access `file://` repos inside the container

### ArgoCD bootstrap (`pkg/argocd/bootstrap.go`, `pkg/argocd/foundational.go`)
- Updated to accept `gitConfig *git.Config` parameter instead of reading from `cfg.GitRepository` directly

### Sync wave fix (`pkg/argocd/templates/apps/`)
- envoy-gateway: wave 2 → 1 (installs Gateway API CRDs first)
- cert-manager: wave 1 → 2 (finds CRDs on startup)

### Makefile
- `LOCAL_CONFIG` variable (overridable) for specifying config file
- `localkind-up` parses config to determine mount strategy:
  - `file:///` URL → mounts that path
  - No git_repository → mounts auto-generated `/tmp/nebari-gitops-{project_name}`
  - Remote URL → no mount needed

### Tests
- `TestAddLocalGitopsMount` — volume/volumeMount structure validation
- `TestAddLocalGitopsMountPreservesExistingKeys` — existing Helm values not clobbered
- `TestSyncWaveOrdering` — correct wave numbers on templates
- `TestEnvoyGatewayBeforeCertManager` — ordering invariant
- Existing tests for `IsLocalPath`, `GetLocalPath`, `Validate` with local paths

### Docs
- `docs/local-kind-development.md` — quick start, gitops modes, teardown, service access, troubleshooting

## Test plan
- [x] `go test -short ./pkg/argocd/ ./cmd/nic/` passes
- [x] `make localkind-up` with no git_repository → auto-generates and mounts `/tmp` dir
- [x] `make localkind-up` with `file:///` URL → mounts explicit path
- [x] ArgoCD repo-server pod can access mounted path
- [x] cert-manager starts without Gateway API CRD errors


cmd/nic/deploy.go
```diff
@@ -183,7 +213,9 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 				},
 			}
```
dcmcand: **Blocker:** `getOrCreateGitConfig(cfg)` is called a second time here, shadowing the `gitConfig` variable from line 170. When `cfg.GitRepository != nil`, line 170 correctly uses the real config, but this second call creates a new struct pointing at `/tmp/nebari-gitops-{name}` - so `InstallFoundationalServices` will always use the auto-generated local path even when a remote git repository is configured.

Fix: remove this second call and reuse the `gitConfig` from line 170.

dcmcand (changes requested):
Thanks for getting on this so quickly! Really nice work overall - the `file://` scheme approach is clean and the test coverage is solid. There are a few changes needed before we can merge, then it should be :rocket:

**Highlights:**
- Using `file://` as the local-path signal is self-documenting and elegant
- `TestNewClientCleansUpOnAuthError` catching temp dir leaks is great
- The sync wave ordering tests enforce business rules, not just code paths
- `generateSecurePassword` accepting `io.Reader` is textbook dependency injection

pkg/git/config.go
```diff
@@ -77,6 +92,20 @@ func (c *Config) Validate() error {
...
+// IsLocalPath returns true if the URL uses the file:// protocol (local filesystem).
+func (c *Config) IsLocalPath() bool {
+	return strings.HasPrefix(c.URL, "file://")
+}
+
+// GetLocalPath returns the filesystem path from a file:// URL.
+// Returns empty string if not a local path.
+func (c *Config) GetLocalPath() string {
+	if !c.IsLocalPath() {
+		return ""
```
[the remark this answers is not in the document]
tylerpotts: Fixed: Added `filepath.Clean()` to sanitize the extracted path and verify it's absolute. Now returns empty string for non-absolute paths, preventing directory traversal attacks like `file:///tmp/nebari/../etc/shadow`.

tylerpotts:
Thanks for the thorough review! All inline comments have been addressed (see resolved threads), and here's how the general feedback was handled:

**Questions:**

- **CommitAndPush on empty local repo:** Added `TestClientCommitLocalOnFreshRepo` which exercises the exact path — `PlainInit` with zero commits, write a file, then `CommitAndPush`. Confirmed go-git handles first commits on orphan branches correctly. The test passes.

- **--regen-apps on localkind-up:** Made it configurable via a `REGEN_APPS` Make variable that defaults to off. Developers can opt in with `REGEN_APPS=--regen-apps make localkind-up` when they want a fresh start.

**Style changes:**

- **getOrCreateGitConfig:** Split into a pure `defaultGitConfig(projectName)` (no I/O, easily testable) and `getOrCreateGitConfig` which calls it and handles the `os.MkdirAll` + logging.

- **bootstrapGitOps config copy:** Extracted into `writeConfigToRepo(configFile, workDir)` helper to keep the orchestrator thin.

- **local-config.yaml indentation:** Fixed the Option 2 comment block to use consistent `#   ` indentation.

Commits: c5dca40, 2e94d41

Makefile
```diff
@@ -114,14 +114,44 @@ localstack-logs: ## Show LocalStack logs
...
+	elif [ -z "$$GITOPS_URL" ]; then \
+		LOCAL_PATH="/tmp/nebari-gitops-$$PROJECT_NAME"; \
+		echo "No git_repository configured, mounting auto-generated dir: $$LOCAL_PATH"; \
+	else \
+		LOCAL_PATH=""; \
+		echo "Remote git repo detected ($$GITOPS_URL), no mount needed"; \
+	fi; \
+	if [ -n "$$LOCAL_PATH" ]; then \
+		mkdir -p "$$LOCAL_PATH"; \
+		echo 'kind: Cluster' > /tmp/kind-config.yaml; \
```
dcmcand: **Stylistic:** Building YAML with 9 separate `echo >>` statements is fragile and hard to read. Consider a heredoc:

```makefile
cat > /tmp/kind-config.yaml <<KINDEOF
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: nebari-local
nodes:
- role: control-plane
  extraMounts:
  - hostPath: "$$LOCAL_PATH"
    containerPath: "$$LOCAL_PATH"
    readOnly: true
KINDEOF
```

Also, `/tmp/kind-config.yaml` is a predictable path that could collide with concurrent runs. `mktemp` would be safer.

examples/local-config.yaml
```diff
@@ -15,9 +15,22 @@ local:
...
-  auth:
-    ssh_key_env: GIT_SSH_PRIVATE_KEY  # Environment variable containing SSH private key
+#
+# OPTION 1: Auto-generated local directory (zero configuration for development)
+# Simply omit the git_repository section and NIC will auto-create /tmp/nebari-gitops-{project_name}
+#
+# OPTION 2: Explicit local directory (full control over location)
+# git_repository:
+#   # Local path to git repository (must be an absolute path)
+#   url: "file:///Users/tylerman/gh/NIC-local-gitops-repo.git"
```
dcmcand: **Stylistic:** This contains a contributor-specific path. Should use a generic placeholder:

```yaml
#   url: "file:///home/user/my-gitops-repo"
```

pkg/argocd/writer_test.go
```diff
@@ -104,3 +104,60 @@ type nopWriteCloser struct {
...
+		t.Fatalf("%s has no sync-wave annotation", appName)
+		return ""
+	}
+
+	envoyWave := getSyncWave("envoy-gateway")
+	certWave := getSyncWave("cert-manager")
+
+	// envoy-gateway must come before cert-manager (lower wave number)
+	// because cert-manager needs Gateway API CRDs that envoy-gateway installs
+	if envoyWave >= certWave {
```
dcmcand: **Stylistic:** `envoyWave >= certWave` compares full annotation strings lexicographically. This works for single-digit wave numbers (`"1"` < `"2"`) but would give wrong results for multi-digit numbers (`"9"` > `"10"` lexicographically). Consider parsing the actual wave numbers as integers for a more robust comparison.

pkg/git/client_impl.go
```diff
@@ -356,6 +464,59 @@ func (c *ClientImpl) CommitAndPush(ctx context.Context, message string) error {
 	return nil
 }
 
+// commitLocal commits changes to a local repository without pushing.
+func (c *ClientImpl) commitLocal(ctx context.Context, message string) error {
```
dcmcand: **Stylistic:** `commitLocal` duplicates most of the staging/status-check/commit logic from the remote branch of `CommitAndPush` (lines 414-465). Consider extracting a shared `stageAndCommit` helper to reduce duplication.

docs/local-kind-development.md
```diff
@@ -0,0 +1,77 @@
...
+make localkind-up
+```
+
+This single command:
+1. Builds the `nic` binary
+2. Creates a `kind` Docker network (`192.168.1.0/24`) for MetalLB
+3. Creates a Kind cluster named `nebari-local` with appropriate volume mounts
+4. Deploys Nebari (ArgoCD, cert-manager, envoy-gateway, MetalLB, Keycloak, etc.)
+
+## GitOps Repository Modes
```
dcmcand: **Stylistic:** The docs should clarify that `file://` repos require the path to be accessible from the cluster node. This works for Kind (via extraMounts), k3s, and bare metal, but not for cloud providers where the Kubernetes nodes don't have access to your local filesystem.

Suggested addition to this section:

> **Note:** `file://` repos only work when the cluster nodes can access the local path (Kind, k3s, bare metal). For cloud providers (AWS, GCP, Azure), use a remote git repository.

cmd/nic/deploy.go
```diff
@@ -53,6 +55,40 @@ func init() {
 	}
 }
 
+// defaultGitConfig returns a default local git configuration for development workflows.
+// This is a pure function with no side effects — directory creation happens separately.
+func defaultGitConfig(projectName string) *git.Config {
+	localPath := fmt.Sprintf("/tmp/nebari-gitops-%s", projectName)
```
dcmcand: **Question:** `getOrCreateGitConfig` runs for every deploy regardless of provider. This means a deploy to AWS/GCP without `git_repository` configured will auto-create `/tmp/nebari-gitops-{name}` and attempt local gitops bootstrapping. Is this the intended behavior for cloud providers, or should it be gated to `provider: local` (or providers where hostPath is available)?

pkg/git/config.go
```diff
@@ -77,6 +93,28 @@ func (c *Config) Validate() error {
...
 
+// IsLocalPath returns true if the URL uses the file:// protocol (local filesystem).
+func (c *Config) IsLocalPath() bool {
+	return strings.HasPrefix(c.URL, "file://")
+}
+
+// GetLocalPath returns the filesystem path from a file:// URL.
+// Returns empty string if not a local path.
+// The path is cleaned to prevent directory traversal attacks.
+func (c *Config) GetLocalPath() string {
```
dcmcand: **Question:** `GetLocalPath()` returns empty string for invalid `file://` URLs (relative paths), but callers like `getOrCreateGitConfig` don't check for empty string before calling `MkdirAll("")`. This would silently succeed without creating anything useful. Should this return an error instead, or should callers check the return value?

pkg/git/config.go
```diff
@@ -77,6 +93,28 @@ func (c *Config) Validate() error {
 	return nil
 }
 
+// IsLocalPath returns true if the URL uses the file:// protocol (local filesystem).
+func (c *Config) IsLocalPath() bool {
```
dcmcand: Nice - clean separation between `IsLocalPath()` and `GetLocalPath()`, and the path sanitization with `filepath.Clean` + absolute path validation in `GetLocalPath` is solid security practice.

pkg/git/client_impl.go
```diff
@@ -396,6 +557,8 @@ func (c *ClientImpl) WriteBootstrapMarker(ctx context.Context) error {
 }
 
 // Cleanup removes temporary resources.
+// Only removes temp directories created for remote repositories.
+// Local file:// paths are never removed.
```
dcmcand: Good detail - `Cleanup()` correctly skips deletion for local paths by checking `tempDir != ""`. Simple and effective, no chance of accidentally deleting the user's local gitops directory.

dcmcand (changes requested):

Makefile
```diff
@@ -114,14 +114,44 @@ localstack-logs: ## Show LocalStack logs
...
+
+localkind-up: build ## Create local kind cluster and deploy Nebari (mounts file:// gitops repos automatically)
 	@echo "Setting up local kind cluster..."
 	@which kind > /dev/null || (echo "Error: kind is not installed" && exit 1)
+	@which yq > /dev/null || (echo "Error: yq is not installed. Please install and try again" && exit 1)
 	@which docker > /dev/null || (echo "Error: Docker is not installed or not running" && exit 1)
 	-docker network create --subnet=192.168.1.0/24 --gateway=192.168.1.1 kind
-	-kind create cluster --name nebari-local
+	@GITOPS_URL=$$(yq '.git_repository.url // ""' $(LOCAL_CONFIG)); \
+	PROJECT_NAME=$$(awk '/^project_name:/ { gsub(/.*project_name:[[:space:]]*/, ""); gsub(/"/, ""); print; exit }' $(LOCAL_CONFIG)); \
```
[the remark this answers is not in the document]
tylerpotts: Fixed: Now using `yq '.project_name // ""'` for PROJECT_NAME extraction, consistent with GITOPS_URL.

Makefile
tylerpotts: Ended up replacing with printf which works correctly in Make recipes. Heredoc doesn't work because each line is interpreted as a directive.

dcmcand (dismissed):
## Re-Review Summary

All 9 previously reported blockers are properly fixed. The code is well-tested (40+ new test cases passing), follows project conventions (OTel instrumentation, slog at app layer, table-driven tests), and the security posture is solid (path traversal prevention, credential scrubbing, read-only mounts).

Nice work addressing all the feedback, Tyler. This is in good shape for merge.

### Suggested follow-up issue

**`scrubSensitiveFields` scoped too broadly** - The current implementation strips any `auth:` key at any depth in the YAML, not just under `git_repository`. Today this is fine since no other config sections use `auth:`, but if one is added later (e.g., `keycloak.auth`, `dns.auth`), it would be silently scrubbed. A follow-up could either scope the scrubbing to the `git_repository` section by tracking indentation depth, or switch to proper YAML parsing with `yaml.v3` to surgically remove only `git_repository.auth` and `git_repository.argocd_auth`.

dcmcand (changes requested):

cmd/nic/deploy_test.go
```diff
@@ -60,3 +60,103 @@ func TestGenerateSecurePasswordFallback(t *testing.T) {
...
+					t.Errorf("scrubSensitiveFields() should NOT contain %q, got:\n%s", s, result)
+				}
+			}
+		})
+	}
+}
+
+func containsString(haystack, needle string) bool {
+	return bytes.Contains([]byte(haystack), []byte(needle))
+}
```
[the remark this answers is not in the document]
tylerpotts: Fixed: Now using `strings.Contains(haystack, needle)` directly instead of the unnecessary bytes conversion.

pkg/argocd/writer_test.go
```diff
@@ -674,3 +675,73 @@ type nopWriteCloser struct {
...
+					num, err := strconv.Atoi(numStr)
+					if err != nil {
+						t.Fatalf("%s has invalid sync-wave value %q: %v", appName, numStr, err)
+					}
+					return num
+				}
+			}
+		}
+		t.Fatalf("%s has no sync-wave annotation", appName)
+		return 0
```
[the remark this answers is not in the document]
tylerpotts: Thanks\!

cmd/nic/deploy.go
```diff
@@ -49,6 +52,50 @@ func init() {
...
+// For local provider without explicit git_repository config, this auto-creates /tmp/nebari-gitops-{project_name}.
+// For cloud providers, explicit git_repository config is required (returns error if not configured).
+func getOrCreateGitConfig(cfg *config.NebariConfig) (*git.Config, error) {
+	if cfg.GitRepository != nil {
+		return cfg.GitRepository, nil
+	}
+
+	// Only auto-create local gitops for the local provider
+	// Cloud providers without explicit git_repository config skip GitOps bootstrapping
+	if cfg.Cluster.ProviderName() != "local" {
```
dcmcand: Branching on `ProviderName() != "local"` is the provider-name switch that CLAUDE.md flags as a code smell under "Abstraction Boundaries." Adding another provider where nodes can see the host filesystem (k3s, kind in CI, microk8s) would silently skip the auto-create path here.

Suggest threading this through `provider.InfraSettings` instead: add `SupportsLocalGitOps bool`, set it to `true` in `pkg/provider/local/provider.go`, and gate on the capability rather than the name. That's the same shape `NeedsMetalLB` and `KeycloakBasePath` already use.

cmd/nic/deploy.go
```diff
@@ -365,20 +440,95 @@ func bootstrapGitOps(ctx context.Context, cfg *config.NebariConfig, regenApps bo
...
+	}
+	slog.Info("Wrote NIC config to repository (auth fields scrubbed)", "path", configDest)
 	return nil
 }
 
+// scrubSensitiveFields removes the auth block from git_repository config to prevent
+// accidentally committing credentials. Only env var names should be in configs,
+// but this provides defense in depth.
+// Only scrubs auth blocks that are nested under git_repository, preserving other auth blocks.
+func scrubSensitiveFields(configBytes []byte) []byte {
```
dcmcand: A hand-rolled indent-based YAML parser in the CLI layer is going to struggle with valid YAML that doesn't match the single example shape: flow style (`auth: {ssh_key_env: X}`), tab-indented blocks, block scalars, and the `ArgoCDAuth` `auth:` nested under `git_repository` will all confuse the indent tracker. It also won't catch credentials embedded in URLs like `https://user:token@host/repo`, which the validation path does allow.

This belongs on the type that owns the schema. Something like `(c *git.Config) RedactedYAML() ([]byte, error)` using the YAML library already in use gives structural guarantees instead of indent-matching, and keeps the redaction rules co-located with the fields they cover.

pkg/argocd/install.go
```diff
@@ -315,3 +329,58 @@ func waitForArgoCDReady(ctx context.Context, client *kubernetes.Clientset, names
 
 	return waitForArgoCDReadyWithLister(ctx, listDeployments, requiredDeployments, timeout)
 }
+
+// addLocalGitopsMount adds a hostPath volume and volumeMount to the ArgoCD
+// repo-server Helm values so it can access a local file:// git repository.
+func addLocalGitopsMount(values map[string]any, localPath string) {
```
dcmcand: Neither `addLocalGitopsMount` nor `appendToSlice` below starts an OpenTelemetry span, and neither takes `ctx`. CLAUDE.md requires every function in `pkg/` to be instrumented; propagation from the parent `Install` span is broken for this call. Same pattern as the other helpers in this file should work:

```go
func addLocalGitopsMount(ctx context.Context, values map[string]any, localPath string) {
    tracer := otel.Tracer(tracerName)
    _, span := tracer.Start(ctx, "argocd.addLocalGitopsMount")
    defer span.End()
    // ...
}
```

pkg/git/client_impl.go
```diff
@@ -304,28 +410,77 @@ func (c *ClientImpl) CommitAndPush(ctx context.Context, message string) error {
...
+		return err
+	}
+
+	_, err := c.stageAndCommit(ctx, span, message)
+	return err
+}
+
+// stageAndCommit stages all changes and commits them. Returns true if a commit was made,
+// false if there were no changes to commit.
+func (c *ClientImpl) stageAndCommit(ctx context.Context, span trace.Span, message string) (bool, error) {
```
dcmcand: `stageAndCommit` takes a `trace.Span` as a parameter instead of starting its own child span from `ctx`. This is the only place in the codebase that threads a span through a signature; everywhere else each function opens its own span. Accepting the caller's span breaks per-function timing in traces (the parent's `span.End` time now includes this function's work) and leaks an OTel type through internal plumbing.

Preferred shape:

```go
func (c *ClientImpl) stageAndCommit(ctx context.Context, message string) (bool, error) {
    tracer := otel.Tracer(tracerName)
    ctx, span := tracer.Start(ctx, "git.stageAndCommit")
    defer span.End()
    // ...
}
```

Callers pass `ctx` as they already do.

dcmcand (commented):
## Review summary

Nice work on the local dev loop. Four design-gating blockers flagged inline - all fixable without redesign. The questions below can be handled in follow-ups if you prefer.

### What's good

- The `defaultGitConfig` / `getOrCreateGitConfig` split separates pure computation from I/O cleanly, which makes the pure half independently testable.
- `appendToSlice` handles both `[]map[string]any` and `[]any` from YAML parsing, which is exactly the right paranoia for that layer.
- `TestEnvoyGatewayBeforeCertManager` parses sync-wave values as integers before comparing. The comment explaining why lexicographic comparison would break on multi-digit waves saved a future bug.
- OTel instrumentation is present on most new functions in `pkg/git/`, and no logging crept into `pkg/` code.
- The cert-manager / envoy-gateway wave swap is the right fix - cert-manager's `config.enableGatewayAPI: true` needs the Gateway API CRDs at startup, and those come from envoy-gateway. Worth updating `docs/design-doc/implementation/10-foundational-software.md §9.2` to reflect the corrected ordering in a follow-up.

### Questions

**Q1: Makefile re-implements the Go mount-path logic.** `localkind-up` parses the config via shell + yq and hardcodes `/tmp/nebari-gitops-$PROJECT_NAME` for the auto-generated case, but `getOrCreateGitConfig` uses `os.TempDir()`. On macOS `os.TempDir()` returns `/var/folders/.../T/`, so the Makefile will mount `/tmp/...` into Kind while the Go binary creates and uses a path under `/var/folders/`. ArgoCD ends up pointed at an empty or stale directory. Commit `496a4e2a` already caught the same category of bug once; the Makefile still has it.

A subcommand like `nic print-gitops-path -f <config>` that the Makefile can call would keep the logic in one place.

**Q2: Should `pkg/argocd` know about `file://` semantics?** `Install()` now reaches into `cfg.GitRepository.IsLocalPath()` and `GetLocalPath()`. Have you considered passing the resolved host path through `provider.InfraSettings` (something like `LocalGitOpsHostPath string`) so `pkg/argocd` stays agnostic of URL schemes? Open to the counter-argument, but the current shape pushes `file://` awareness into a second package.

**Q3: `GetLocalPath` godoc overstates the guarantee.** The comment says it "prevents directory traversal attacks," but `filepath.Clean` only normalizes `..`; it doesn't constrain the result to a base directory. `file:///tmp/nebari/../etc/shadow` still resolves to `/tmp/etc/shadow`, which the tests document. Fine for a dev-only tool, but the godoc implies a security property that isn't there. Soften the wording or add the base-dir constraint.

cmd/nic/deploy.go
tylerpotts: Fixed: Added `SupportsLocalGitOps bool` to `provider.InfraSettings` struct and set it to `true` in the local provider. `getOrCreateGitConfig` now takes `supportsLocalGitOps` parameter instead of checking provider name directly. This follows the same pattern as `NeedsMetalLB` and `KeycloakBasePath`.

cmd/nic/deploy.go
tylerpotts: Fixed: Replaced hand-rolled indent parser with proper `gopkg.in/yaml.v3` parsing. The function now unmarshals to `map[string]any`, deletes auth fields from `git_repository`, adds a `_auth_scrubbed` marker, and re-marshals. This handles all valid YAML formats (flow style, tabs, block scalars) and only touches fields under `git_repository`.

pkg/argocd/install.go
tylerpotts: Fixed: Added OTel span to `addLocalGitopsMount` with `local_path` attribute. The function now takes `ctx context.Context` as the first parameter for span propagation, following the standard pattern used elsewhere in pkg/argocd.

pkg/git/client_impl.go
tylerpotts: Fixed: Refactored `stageAndCommit` to create its own child span from ctx instead of accepting a span parameter. The signature is now `stageAndCommit(ctx context.Context, message string)` which follows the standard pattern used everywhere else in the codebase. Removed the unused `trace` import.

viniciusdc:
As a side note, this has been running marvelously trough the https://github.com/nebari-dev/action-nebari-sandbox

pkg/git/config.go
```diff
@@ -77,6 +96,28 @@ func (c *Config) Validate() error {
...
+	}
+	path := strings.TrimPrefix(c.URL, "file://")
+	// Clean the path to resolve ".." and other traversal attempts
+	path = filepath.Clean(path)
+	// Ensure the path is absolute to prevent relative path attacks
+	if !filepath.IsAbs(path) {
+		return "", fmt.Errorf("file:// URL must use absolute path, got: %s", path)
+	}
+	return path, nil
+}
```
dcmcand, line 119: Solid: `filepath.Clean` plus an explicit `IsAbs` check, and the `(string, error)` return forces callers to handle invalid paths. Good defense against the traversal case from the last review.

pkg/git/config.go
```diff
@@ -146,6 +187,19 @@ func (a *AuthConfig) GetToken() (string, error) {
...
+// The auth block is replaced with a placeholder, suitable for writing to
+// the gitops repo without leaking credentials.
+func (c *Config) RedactedCopy() *Config {
+	return &Config{
+		URL:    c.URL,
+		Branch: c.Branch,
+		Path:   c.Path,
+		// Auth and ArgoCDAuth are intentionally omitted (zero values)
+		// to prevent accidentally committing credentials
+	}
```
dcmcand, line 200: Nice defense in depth - even if a user hardcodes credentials in their config, `RedactedCopy()` keeps them out of the gitops repo.

cmd/nic/deploy.go
```diff
@@ -368,20 +444,76 @@ func bootstrapGitOps(ctx context.Context, cfg *config.NebariConfig, regenApps bo
...
+		delete(gitRepo, "argocd_auth")
+		// Add a comment placeholder (YAML doesn't preserve comments, so add as a field)
+		gitRepo["_auth_scrubbed"] = "credentials removed for security"
+	}
+
+	scrubbed, err := yaml.Marshal(doc)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal scrubbed config: %w", err)
+	}
+	return scrubbed, nil
```
dcmcand, line 514: Clean approach: parsing the YAML and deleting the auth keys is much more robust than regex-scrubbing, and the `_auth_scrubbed` marker is a nice human-readable hint.

pkg/argocd/install.go
```diff
@@ -315,3 +329,63 @@ func waitForArgoCDReady(ctx context.Context, client *kubernetes.Clientset, names
...
+	}
+	repoServer["volumes"] = appendToSlice(repoServer["volumes"], newVolume)
+
+	// Append to existing volumeMounts
+	newMount := map[string]any{
+		"name":      "local-gitops",
+		"mountPath": localPath,
+		"readOnly":  true,
+	}
+	repoServer["volumeMounts"] = appendToSlice(repoServer["volumeMounts"], newMount)
```
dcmcand, line 363: Good - `readOnly: true` plus the `appendToSlice` helper means existing volumes/volumeMounts from `DefaultConfig()` (or future additions) won't be silently clobbered.

pkg/argocd/install.go
```diff
@@ -315,3 +329,63 @@ func waitForArgoCDReady(ctx context.Context, client *kubernetes.Clientset, names
...
+			if m, ok := item.(map[string]any); ok {
+				result = append(result, m)
+			}
+		}
+		return append(result, newItem)
+	}
+
+	// Fallback: just return new item
+	return []map[string]any{newItem}
+}
```
dcmcand, line 391: Nice helper - handling both the typed `[]map[string]any` and untyped `[]any` shapes that come out of YAML parsing avoids a class of subtle bugs.

pkg/git/client_impl.go
```diff
@@ -338,22 +496,11 @@ func (c *ClientImpl) CommitAndPush(ctx context.Context, message string) error {
...
-	})
-	if err != nil {
-		span.RecordError(err)
-		return fmt.Errorf("failed to push: %w", err)
-	}
-
-	return nil
+	span.SetAttributes(attribute.Bool("git.committed", true))
+	return true, nil
 }
```
dcmcand, line 504: Clean refactor - extracting `stageAndCommit` removes the duplication between local and remote commit paths cleanly.

pkg/git/client_impl.go
```diff
@@ -396,6 +543,8 @@ func (c *ClientImpl) WriteBootstrapMarker(ctx context.Context) error {
...
 
 // Cleanup removes temporary resources.
+// Only removes temp directories created for remote repositories.
+// Local file:// paths are never removed.
 func (c *ClientImpl) Cleanup() error {
 	if c.tempDir != "" {
```
dcmcand, line 549: Good - the `tempDir != ""` guard means a developer's local gitops directory will never be deleted by `Cleanup()`.

pkg/provider/provider.go
```diff
@@ -57,6 +57,11 @@ type InfraSettings struct {
 	// Empty when EFS is not enabled. Software packs can use this for ReadWriteMany
 	// volumes (e.g., shared model storage for LLM serving).
 	EFSStorageClass string
+
+	// SupportsLocalGitOps indicates whether this provider can use local file:// git repos.
+	// True for providers where cluster nodes can access host filesystem paths (local, kind, k3s).
+	// Cloud providers (AWS, GCP, Azure) return false - their nodes can't see the dev machine's FS.
+	SupportsLocalGitOps bool
```
dcmcand, line 64: This is the right architectural seam - capability flag on `InfraSettings` instead of a provider-name switch in the CLI, matching the convention documented in CLAUDE.md.

dcmcand (approved):
Re-review: all four design-gating blockers from the previous review have been properly addressed, tests pass, and the architectural integration via `InfraSettings.SupportsLocalGitOps` matches the pattern documented in CLAUDE.md. Nice work on the iteration.

### Resolved from prior review

- Path traversal in `GetLocalPath()` - now uses `filepath.Clean` + `filepath.IsAbs`, returns `(string, error)`
- Duplicate `getOrCreateGitConfig` call collapsed to a single call site
- `readOnly: true` on the repo-server volumeMount and the Kind `extraMounts`
- Volumes/volumeMounts no longer clobbered (`appendToSlice` handles both YAML-parsed shapes)
- Credentials scrubbed from `nic-config.yaml` before it lands in the gitops repo
- Makefile uses `yq` for proper YAML path extraction instead of fragile awk regex
- Sync wave comparison is numeric (`strconv.Atoi`) instead of lexicographic

### A couple of follow-up notes (non-blocking)

These are easy follow-ups, not approval-gating:

1. **Silent skip on cloud providers without `git_repository`** (`cmd/nic/deploy.go:78-82`): currently logs at Info and returns `(nil, nil)`. A user who forgot to configure `git_repository` for AWS/GCP would see `Deployment completed successfully` despite ArgoCD apps never being bootstrapped. Consider escalating to `Warn` so it's harder to miss.
2. **`_auth_scrubbed` placeholder key**: `scrubSensitiveFields` injects a phantom field that isn't part of the `NebariConfig` schema. Fine for human reference, but worth keeping in mind if a future feature re-parses `nic-config.yaml` from the gitops repo.
3. **`/tmp/nebari-gitops-{name}` on shared hosts**: the directory itself is 0750, but `/tmp` allows enumeration. Worth a one-line caveat in the docs for anyone running on a multi-user box.

Approving - happy to see these handled here or in a follow-up.

viniciusdc:
Since this has both approvals and the review suggestions have been addressed I am merging this

PR #144, opened 2026-03-10
Modifications and override functionality for otel collector

tylerpotts:
## Summary

- **OpenTelemetry Collector: overridable values** — Moves the OTel collector Helm values from inline in the ArgoCD Application template to external files (`values.yaml` + `overrides.yaml`) managed in the GitOps repo. Uses a new `_nooverwrite_` file convention so `overrides.yaml` is created on first deploy but preserved on subsequent runs, allowing users to customize exporters, pipelines, and other collector config without losing changes.
- **Multi-source ArgoCD pattern** — Converts the OTel collector app from single-source inline values to ArgoCD multi-source, referencing both the Helm chart and a `$values` source from the GitOps repo with `valueFiles` (base values first, then overrides for precedence).
- **`_nooverwrite_` template convention** — Extends `WriteAllToGit` to recognize `_nooverwrite_`-prefixed template files: written to the destination (with the prefix stripped) only if the file does not already exist. Uses `embed all:templates` to include these underscore-prefixed files.
- **Keycloak URL handling** — Adds `KeycloakIssuerURL` (external public URL for token `iss` claim validation) and `KeycloakAdminSecretNamespace` to template data. Fixes `deployment-patch.yaml` to use the pre-built `KeycloakServiceURL` instead of appending a hardcoded `/auth` suffix.
- **OTel cluster role** — Adds `nodes/metrics` and `nodes/proxy` resources plus `/metrics/cadvisor` non-resource URL to the collector's ClusterRole for cAdvisor scraping.
- **Nebari operator version bump**

## Test plan

- [x] `golangci-lint run` passes
- [x] `go test -short ./pkg/argocd/...` passes
- [x] Table-driven tests for `_nooverwrite_` (file absent creates, file present preserves)
- [x] Tests verify `values.yaml` precedes `overrides.yaml` in `valueFiles` ordering
- [x] Tests verify Keycloak URL rendering with and without `/auth` base path
- [ ] Deploy to a test cluster and verify OTel collector syncs with multi-source config
- [ ] Verify `overrides.yaml` survives a re-deploy

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 <p align="center">
   <a href="https://nebari.dev">
-    <img src="docs/assets/nebari-logo.svg" alt="Nebari" width="400">
+    <picture>
+      <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/nebari-dev/nebari-design/main/logo-mark/horizontal/standard/Nebari-Logo-Horizontal-Lockup-White-text.png">
+      <source media="(prefers-color-scheme: light)" srcset="https://raw.githubusercontent.com/nebari-dev/nebari-design/main/logo-mark/horizontal/standard/Nebari-Logo-Horizontal-Lockup.png">
+      <img alt="Nebari" src="docs/Nebari-Logo-Horizontal-Lockup.png" width="300">
+    </picture>
   </a>
 </p>
 
--- a/pkg/argocd/templates/manifests/nebari-operator/deployment-patch.yaml
+++ b/pkg/argocd/templates/manifests/nebari-operator/deployment-patch.yaml
@@ -24,6 +24,8 @@ spec:
               value: "{{ .CertificateIssuer }}"
             - name: KEYCLOAK_ISSUER_CONTEXT_PATH
               value: "{{ .KeycloakBasePath }}"
+            - name: KEYCLOAK_EXTERNAL_URL
+              value: "https://keycloak.{{ .Domain }}{{ .KeycloakBasePath }}"
           resources:
             requests:
               cpu: 100m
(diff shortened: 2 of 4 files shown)


pkg/argocd/templates/manifests/opentelemetry-collector/values.yaml
```diff
@@ -0,0 +1,109 @@
...
+                regex: ([^:]+)(?::\d+)?;(\d+)
+                replacement: $1:$2
+                target_label: __address__
+          - job_name: 'kubernetes-nodes-cadvisor'
+            kubernetes_sd_configs:
+              - role: node
+            scheme: https
+            tls_config:
+              ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+              insecure_skip_verify: true
```
dcmcand: **Stylistic:** `insecure_skip_verify: true` makes the `ca_file` on line 53 a no-op since cert verification is disabled entirely. This is a common pattern in Kubernetes monitoring setups because kubelet certs often have mismatched SANs, but a comment explaining *why* it's needed would help:

```yaml
# insecure_skip_verify is needed because kubelet serving certs
# may not include the API server proxy hostname in their SANs
insecure_skip_verify: true
```

pkg/argocd/templates/manifests/opentelemetry-collector/_nooverwrite_overrides.yaml
```diff
@@ -0,0 +1,5 @@
+# OpenTelemetry Collector overrides
+# Software packs can add exporters and pipeline configuration here.
+# Values in this file are deep-merged with the base config.
+# Note: array values (like pipeline exporter lists) are replaced, not appended.
+{}
```
dcmcand: **Stylistic:** Some Helm versions handle an explicit empty map `{}` differently from a comment-only file during value merging. A safer no-op would be to drop the `{}` and leave just the comments:

```yaml
# OpenTelemetry Collector overrides
# Software packs can add exporters and pipeline configuration here.
# Values in this file are deep-merged with the base config.
# Note: array values (like pipeline exporter lists) are replaced, not appended.
```

PR #201, opened 2026-03-26
feat(local): make StorageClass, MetalLB, and HTTPSPort configurable

### LINKED ISSUE #200: Local provider: make StorageClass and MetalLB configurable
## Problem

The local provider hardcodes infrastructure settings that don't work across different local Kubernetes distributions:

```go
// pkg/provider/local/provider.go:282
StorageClass:       "standard",
NeedsMetalLB:       true,
MetalLBAddressPool: "192.168.1.100-192.168.1.110",
```

- **StorageClass `standard`** doesn't exist on k3s (which uses `local-path`). This causes PVCs like PostgreSQL for Keycloak to get stuck in `Pending`.
- **MetalLB is always enabled**, but k3s ships with ServiceLB (formerly Klipper) and doesn't need MetalLB. Deploying both can conflict.
- **MetalLB address pool `192.168.1.x`** doesn't work on kind, where the Docker bridge network uses `172.18.0.0/16`. Services get IPs that are unreachable from the host.

The local provider is used for both kind and k3s (and potentially other distributions), but the hardcoded defaults only work for a narrow case.

## Proposed Solution

Make these settings configurable in the YAML config under the `local` provider block:

```yaml
provider: local
local:
  storage_class: local-path        # default: "standard"
  metallb:
    enabled: false                  # default: true
    address_pool: 172.18.255.100-172.18.255.110  # default: "192.168.1.100-192.168.1.110"
```

The `InfraSettings()` method would read these from the provider config instead of returning hardcoded values.

### Sensible defaults by distribution

| Setting | kind | k3s |
|---------|------|-----|
| `storage_class` | `standard` | `local-path` |
| `metallb.enabled` | `true` | `false` |
| `metallb.address_pool` | (must match Docker network) | n/a |

## Related

- #188 — Add `existing` cluster provider (proposes `storage_class` field)

dcmcand:
## Summary

- Add `storage_class`, `https_port`, and `metallb` (enabled + address_pool) as optional fields in the local provider config
- `InfraSettings()` reads these from config with current hardcoded values as defaults
- Fixes local provider only working for kind clusters - k3s users can now set `storage_class: local-path` and `metallb.enabled: false`

## Closes

Closes #200

### Config (`pkg/provider/local/config.go`)
- Added `StorageClass`, `HTTPSPort` fields to `Config`
- Added `MetalLBConfig` struct with `Enabled` (`*bool` to distinguish unset from false) and `AddressPool`

### Implementation (`pkg/provider/local/provider.go`)
- `InfraSettings()` now parses the local config block and overrides defaults for any set fields
- Parse errors silently return defaults (same pattern as Hetzner provider)

### Tests (`pkg/provider/local/provider_test.go`)
- 8 table-driven test cases: no config, empty config, each field individually, full override, unmarshal error

### Examples (`examples/local-config.yaml`)
- Added commented examples showing k3s and custom MetalLB pool configurations

## Example usage

k3s cluster:
```yaml
local:
  storage_class: local-path
  https_port: 8443
  metallb:
    enabled: false
```

Custom MetalLB pool (kind on non-default Docker network):
```yaml
local:
  metallb:
    address_pool: 172.18.255.100-172.18.255.110
```

## Test plan

- [x] `go test ./pkg/provider/local/ -v` passes (8/8 test cases)
- [x] `golangci-lint run ./pkg/provider/local/` passes
- [x] Full project test suite passes


viniciusdc:
Hey @dcmcand — could you rebase this against main and resolve the conflicts when you get a chance? Three refactors landed after this was opened (#190, #208, #211) that changed the `Provider.InfraSettings` signature to take `*config.ClusterConfig` and moved provider config under a nested `cluster:` block. The diff itself is small (mostly mirroring the pattern already in `Summary()`), so it should be a fairly mechanical rebase. Happy to review once it's green.

dcmcand:
@viniciusdc it is rebased now

viniciusdc (commented):
Looked through the diff and nebari-dev/action-nebari-sandbox#15 — a few things worth discussing before merging.

**Two questions:**

`pkg/provider/local/provider.go` — the comment says "InfraSettings is called after Validate() has confirmed the config is parseable", but that ordering isn't enforced anywhere in the code. If someone calls `InfraSettings` without going through `Validate` (in a test, or a future codepath), they'll silently get defaults with no indication something went wrong. Is there a lifecycle guarantee that makes this safe? Or should we at least log the error instead of swallowing it?

Also on `provider.go` — why does `UnmarshalProviderConfig` take a context at all? For a pure config unmarshal `context.Background()` is fine, but if the function does any I/O the context matters and should be threaded through. If it's purely structural (the interface requires it), a quick comment there would save future readers the same question.

**Two small things:**

`provider_test.go` — the test covers `metallb.enabled: false` and the default (`true` implicitly), but not `enabled: true` explicitly. That's the case where the pointer is non-nil but `*enabled == true` — small gap in the pointer semantics coverage.

`provider.go` — `cfg.ProviderConfig()` is called without guarding against a nil `cfg`. Current callers never pass nil (and the `"no local config block"` test case passes a non-nil `cfg` with `Providers: nil`, which the `rawCfg == nil` guard handles correctly), but the function signature allows it. Worth a nil check at the top for robustness.

**What looks good:**

`Enabled *bool` for MetalLB is the right design — cleanly distinguishes "not set" (default true) from "explicitly false" without overloading the zero value. `omitempty` on a pointer works exactly as expected here.

The 8 table-driven test cases are solid: defaults, each field in isolation, full override, and the unmarshal error path. I also noticed the assertions verify the fields this PR doesn't touch (`LoadBalancerAnnotations`, `KeycloakBasePath`, `SupportsLocalGitOps`) haven't regressed — good discipline.

The integration evidence from nebari-dev/action-nebari-sandbox#15 is the best part: it removes the `standard` StorageClass workaround from `create-cluster.sh` — the one with the TODO pointing directly at this PR. That deletion is the cleanest possible proof this solves the underlying problem. The workaround only existed because this config field didn't. Full `platform` profile runs end-to-end against this branch in CI.

`examples/local-config.yaml` is clear and matches the actual struct. The k3s and custom MetalLB pool examples cover exactly the two cases that motivated this.

pkg/provider/local/provider.go
```diff
@@ -267,11 +267,42 @@ func (p *Provider) Summary(clusterConfig *config.ClusterConfig) map[string]strin
 }
 
 // InfraSettings returns local provider Kubernetes infrastructure settings.
-func (p *Provider) InfraSettings(_ *config.ClusterConfig) provider.InfraSettings {
-	return provider.InfraSettings{
+// Values are read from the local provider config block, falling back to defaults.
+// Parse errors are intentionally ignored: InfraSettings is called after Validate()
+// has confirmed the config is parseable. If it somehow fails (e.g., nil config in
+// tests), we return valid defaults.
```
viniciusdc, line 273: The comment says InfraSettings is called after Validate() has confirmed the config is parseable, but that ordering isn't enforced anywhere in the code. If someone calls `InfraSettings` without going through `Validate` (in a test, or a future codepath), they'll silently get defaults with no indication something went wrong. Is there a lifecycle guarantee that makes this safe? Or should we at least log the error instead of swallowing it?

pkg/provider/local/provider.go
```diff
@@ -267,11 +267,42 @@ func (p *Provider) Summary(clusterConfig *config.ClusterConfig) map[string]strin
...
 		SupportsLocalGitOps: true,
 	}
+
+	rawCfg := cfg.ProviderConfig()
+	if rawCfg == nil {
+		return settings
+	}
+
+	var localCfg Config
+	if err := config.UnmarshalProviderConfig(context.Background(), rawCfg, &localCfg); err != nil {
```
viniciusdc, line 288: Why does `UnmarshalProviderConfig` take a context at all? For a pure config unmarshal `context.Background()` is fine, but if the function does any I/O the context matters and should be threaded through. If it's purely structural (the interface requires it), a quick comment here would save future readers the same question.

pkg/provider/local/provider.go
```diff
@@ -267,11 +267,42 @@ func (p *Provider) Summary(clusterConfig *config.ClusterConfig) map[string]strin
...
+// tests), we return valid defaults.
+func (p *Provider) InfraSettings(cfg *config.ClusterConfig) provider.InfraSettings {
+	settings := provider.InfraSettings{
 		StorageClass:        "standard",
 		NeedsMetalLB:        true,
 		MetalLBAddressPool:  "192.168.1.100-192.168.1.110",
 		SupportsLocalGitOps: true,
 	}
+
+	rawCfg := cfg.ProviderConfig()
```
viniciusdc, line 282: `cfg.ProviderConfig()` is called without guarding against a nil `cfg`. Current callers never pass nil (the `"no local config block"` test passes a non-nil `cfg` with `Providers: nil`, which the `rawCfg == nil` guard on the next line handles correctly), but the function signature allows it. Worth a nil check at the top for robustness.

pkg/provider/local/provider_test.go
```diff
@@ -12,29 +12,135 @@ var _ provider.Provider = (*Provider)(nil)
...
+			providerConfig: map[string]any{
+				"local": map[string]any{
+					"metallb": map[string]any{"enabled": false},
+				},
+			},
+			wantSC:        "standard",
+			wantMetalLB:   false,
+			wantPool:      "192.168.1.100-192.168.1.110",
+			wantHTTPSPort: 0,
+		},
```
viniciusdc, line 61: The test covers `enabled: false` and the default (`true` implicitly), but not `enabled: true` explicitly — that's the case where the pointer is non-nil but `*enabled == true`. Small gap in the pointer semantics coverage.

pkg/provider/local/config.go
```diff
@@ -1,8 +1,22 @@
...
+	HTTPSPort        int                          `yaml:"https_port,omitempty"`
+	MetalLB          *MetalLBConfig               `yaml:"metallb,omitempty"`
 	AdditionalFields map[string]any               `yaml:",inline"`
 }
+
+// MetalLBConfig holds MetalLB-specific settings for the local provider.
+type MetalLBConfig struct {
+	// Enabled controls whether MetalLB is deployed. Default: true.
+	// Use a pointer to distinguish "not set" (default true) from "explicitly false".
+	Enabled *bool `yaml:"enabled,omitempty"`
```
viniciusdc, line 17: `Enabled *bool` is the right call here — cleanly distinguishes "not set" (default true) from "explicitly false" without overloading the zero value. `omitempty` on a pointer works exactly as expected.

viniciusdc (commented):
Full `platform` profile runs end-to-end against this branch in CI nebari-dev/action-nebari-sandbox#15

viniciusdc (approved):

PR #206, opened 2026-03-30
fix: Resolve k3s versions from hetzner-k3s binary instead of GitHub API

### LINKED ISSUE #205: k3s version resolver picks versions unsupported by hetzner-k3s
## Bug

`resolveK3sVersion()` resolves a short Kubernetes version (e.g. `1.34`) to a full k3s release tag by querying the **GitHub k3s releases API**. However, hetzner-k3s maintains its own curated list of supported versions, which can lag behind GitHub.

This causes a deploy failure when the GitHub API returns a newer patch release (e.g. `v1.34.6+k3s1`) that hetzner-k3s doesn't support yet (latest is `v1.34.5+k3s1`):

```
[Configuration]  - K3s version is not valid, run `hetzner-k3s releases` to see available versions
```

## Root Cause

`pkg/provider/hetzner/k3s.go` queries `https://api.github.com/repos/k3s-io/k3s/releases` and returns the first stable match. It does not validate the result against what `hetzner-k3s releases` actually supports.

## Fix

Replace the GitHub API call with output from `hetzner-k3s releases`, which is the authoritative list of supported versions. The binary is already downloaded before version resolution happens.

tylerpotts:
## Summary

- Replace GitHub API call in `resolveK3sVersion()` with `hetzner-k3s releases` output
- The GitHub API can return k3s versions newer than what hetzner-k3s supports, causing deploy failures with `K3s version is not valid`
- The hetzner-k3s binary is already downloaded before version resolution, so we use it as the authoritative source

## Test plan

- [x] `TestResolveK3sVersion` — all existing cases pass with fake binary
- [x] `TestResolveK3sVersion_BinaryError` — verifies error on missing binary
- [x] `golangci-lint run` — clean
- [x] Manual: `./nic deploy -f examples/hetzner-tyler-config.yaml` with `kubernetes_version: "1.34"` resolves to `v1.34.5+k3s1` (latest supported by hetzner-k3s)

Fixes #205


pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +44,47 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
+// getHetznerK3sReleases runs `hetzner-k3s releases` and returns the list of
+// supported version tags. Results are returned in the order the binary outputs
+// them (newest first).
+func getHetznerK3sReleases(ctx context.Context, binaryPath string) ([]string, error) {
+	cmd := exec.CommandContext(ctx, binaryPath, "releases")
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	if err := cmd.Run(); err != nil {
```
dcmcand: **Question**: The old code had a 30-second `context.WithTimeout`. If `hetzner-k3s releases` makes a network call internally and hangs, there's no deadline here. Worth wrapping with a timeout?

pkg/provider/hetzner/k3s_test.go
```diff
@@ -2,25 +2,32 @@ package hetzner
 
 import (
 	"context"
-	"encoding/json"
-	"net/http"
-	"net/http/httptest"
+	"os"
+	"path/filepath"
 	"strings"
```
dcmcand: Nice approach with the fake binary - it exercises the full `exec.CommandContext` path. (Though if `resolveK3sVersion` is refactored to accept `[]string`, this helper becomes unnecessary and the tests get even simpler.)

dcmcand (changes requested):
Good fix - using `hetzner-k3s releases` as the source of truth directly eliminates the version mismatch bug (#205). Removing the GitHub API dependency also drops the rate limiting, pagination, and User-Agent concerns. The test cases are well-chosen and port over cleanly.

One blocker around separating I/O from logic, and a few questions below.

pkg/provider/hetzner/k3s.go
```diff
@@ -1,37 +1,27 @@
...
+// to a full k3s release tag (e.g., "v1.32.12+k3s1") by matching against the
+// provided list of available releases. This is a pure function - the caller is
+// responsible for fetching the releases list from the hetzner-k3s binary.
 //
-// Note: only the 100 most recent releases are fetched (no pagination).
-// Older Kubernetes versions may not resolve if there are 100+ newer releases.
-// Use an explicit k3s version string (e.g., "v1.28.5+k3s1") for older versions.
-func resolveK3sVersion(ctx context.Context, version string, apiURL string) (string, error) {
+// If version already contains "+k3s", it's returned as-is.
+// releases should be ordered newest-first for correct resolution.
```
dcmcand: **Blocker**

This doc comment is incorrect: `hetzner-k3s releases` output is oldest-first, not newest-first.

I confirmed empirically against the v2.4.6 binary. Output starts at `v0.1.0-rc1` and ends at `v1.35.3+k3s1`. Upstream source in `src/k3s.cr`:

```crystal
private def self.fetch_all_releases_from_github : Array(String)
    ...
    releases.reverse   # GitHub /tags is newest-first; reverse -> oldest-first
```

Cross-check in `src/kubernetes/script/master_generator.cr`: `selected_k3s_index >= k3s_1_23_6_index` gates `--flannel-backend=wireguard-native` (k3s 1.23.6+). For that `>=` comparison to select newer versions, newer releases must have a HIGHER index, which only holds if the list is oldest-first.

**Impact**: for `kubernetes_version: "1.34"`, the loop below returns the first prefix match on the oldest-first list, which is `v1.34.1+k3s1` (no stable `v1.34.0+k3s1` exists), not `v1.34.6+k3s1`. Silent downgrade onto an outdated patch.

I was not able to reproduce the PR description's manual test result of `"1.34" -> v1.34.5+k3s1`. Tracing the code against live binary output gives `v1.34.1+k3s1`. Could you share the config you used? A full patch like `"1.34.5"` would resolve correctly, but `"1.34"` does not.

**Fix**: iterate `releases` in reverse, or sort semver-descending before matching. Please also update the test fixtures in `k3s_test.go`: they are currently newest-first, which hides this bug.

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +44,52 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
-		}
-		if strings.HasPrefix(r.TagName, matchPrefix) {
-			span.SetAttributes(attribute.String("resolved_version", r.TagName))
-			return r.TagName, nil
+	var releases []string
+	for _, line := range strings.Split(stdout.String(), "\n") {
+		tag := strings.TrimSpace(line)
+		if tag != "" {
+			releases = append(releases, tag)
 		}
```
dcmcand, line 146: **Blocker**

This accepts every non-empty line as a release tag. Running `hetzner-k3s releases` against v2.4.6 actually produces:

```
╭─────────────────────────────────────────╮
│            hetzner-k3s                  │
│   Production-ready K8s on Hetzner       │
╰─────────────────────────────────────────╯

Version: 2.4.6

Available k3s releases:
v0.1.0-rc1
...
```

All banner lines, `Version: 2.4.6`, and `Available k3s releases:` get appended into `releases`. It only "works" today because the downstream `HasPrefix(tag, "v1.XX.")` filter rejects them. Any future formatting change in hetzner-k3s silently breaks this.

**Fix**: validate each line against a version pattern before appending, for example a regex like `^v\d+\.\d+\.\d+(-[a-z]+\d*)?\+k3s\d+$`. Consider erroring if zero lines match so format drift fails loudly instead of silently returning no results.

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +44,52 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
+	return strings.Contains(tag, "-rc") ||
+		strings.Contains(tag, "-alpha") ||
+		strings.Contains(tag, "-beta")
+}
+
+// getHetznerK3sReleases runs `hetzner-k3s releases` and returns the list of
+// supported version tags. Results are returned in the order the binary outputs
+// them (newest first). A 30-second timeout is applied to prevent hanging if
+// the binary makes network calls internally.
+func getHetznerK3sReleases(ctx context.Context, binaryPath string) ([]string, error) {
```
dcmcand, line 118: **Question**

`CLAUDE.md`:
> All new Go functions must be instrumented. Use `tracer.Start(ctx, "functionName")` at the beginning.

This function is new but has no span. Worth adding one with `binary.path`, `release.count`, and duration as attributes? Would help a lot when debugging binary-exec hangs or empty-output bugs in production.

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +44,52 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
dcmcand: **Question**

Is 30s enough on a cold cache? hetzner-k3s caches releases for 7 days in `~/.hetzner-k3s/k3s-releases.yaml`. On a cache miss it paginates through all k3s tags (currently 70+ pages at 100 per page) via `Crest.get`. The previous code fetched a single page, so 30s was generous. Here it could be tight on slow connections or when GitHub is rate-limiting. Did you measure the cache-miss case?

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +44,52 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
+func isPrerelease(tag string) bool {
+	return strings.Contains(tag, "-rc") ||
+		strings.Contains(tag, "-alpha") ||
+		strings.Contains(tag, "-beta")
+}
+
+// getHetznerK3sReleases runs `hetzner-k3s releases` and returns the list of
+// supported version tags. Results are returned in the order the binary outputs
+// them (newest first). A 30-second timeout is applied to prevent hanging if
+// the binary makes network calls internally.
```
dcmcand: **Question**

Running `hetzner-k3s releases` creates and updates `~/.hetzner-k3s/k3s-releases.yaml` (7-day cache). Version resolution now has a filesystem side effect in the user's home directory that did not exist before. Worth calling this out in the function doc, and maybe considering whether it should respect the existing `cacheDir` convention via `HOME` override or similar. Not blocking, but surprising.

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +44,52 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
 
-	const maxAPIResponseSize = 10 * 1024 * 1024 // 10 MB
-	var releases []ghRelease
-	if err := json.NewDecoder(io.LimitReader(resp.Body, maxAPIResponseSize)).Decode(&releases); err != nil {
-		return "", fmt.Errorf("failed to decode k3s releases: %w", err)
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("failed to get hetzner-k3s releases: %w: %s", err, stderr.String())
```
dcmcand: **Question**

stderr from the binary may contain ANSI escape codes and trailing newlines. Consider `strings.TrimSpace(stderr.String())` or `%q` to quote-escape control chars so the wrapped error is readable in log aggregators.

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +44,52 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
 	}
-	req.Header.Set("Accept", "application/vnd.github.v3+json")
-	req.Header.Set("User-Agent", userAgent)
 
-	resp, err := http.DefaultClient.Do(req)
-	if err != nil {
-		return "", fmt.Errorf("failed to fetch k3s releases: %w", err)
-	}
-	defer resp.Body.Close() //nolint:errcheck // Best-effort close on read-only response
+	return "", fmt.Errorf("no supported k3s release found for kubernetes version %q (run 'hetzner-k3s releases' to see available versions)", version)
```
dcmcand, line 72: **Stylistic**

Since the full releases list is already in memory when we hit this path, the error could be more actionable: list the minor versions we did see, for example `"available minor versions: 1.31, 1.32, 1.33, 1.34, 1.35"`. Saves the user running the binary again. Low priority.

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +44,52 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
-		err := fmt.Errorf("GitHub API returned status %d", resp.StatusCode)
-		span.RecordError(err)
-		return "", err
-	}
+// isPrerelease returns true if the version tag indicates a pre-release version.
+// Checks for common pre-release suffixes: -rc, -alpha, -beta.
+func isPrerelease(tag string) bool {
+	return strings.Contains(tag, "-rc") ||
+		strings.Contains(tag, "-alpha") ||
+		strings.Contains(tag, "-beta")
```
dcmcand: **Stylistic**

`strings.Contains` is loose. Current k3s tag shapes like `v1.32.12-rc1+k3s1` are fine, but a tighter check (regex `-(rc|alpha|beta)\d*\+k3s` or validating the suffix sits between the patch and `+k3s`) would be more defensive against surprising tag shapes. Low priority.

pkg/provider/hetzner/k3s_test.go
```diff
@@ -2,26 +2,17 @@ package hetzner
...
 
 func TestResolveK3sVersion(t *testing.T) {
-	releases := []ghRelease{
-		{TagName: "v1.32.12+k3s1", Prerelease: false},
-		{TagName: "v1.32.12-rc1+k3s1", Prerelease: true},
-		{TagName: "v1.32.11+k3s3", Prerelease: false},
-		{TagName: "v1.31.5+k3s1", Prerelease: false},
+	releases := []string{
+		"v1.32.12+k3s1",
+		"v1.32.12-rc1+k3s1",
```
dcmcand: **Stylistic**

These fixtures are newest-first, which is the opposite of what the real binary outputs and hides the ordering bug flagged on `k3s.go`. Consider adding a fixture that mirrors real `hetzner-k3s releases` output (oldest-first, banner header present) so unit tests exercise realistic conditions and catch format regressions.

pkg/provider/hetzner/k3s.go
```diff
@@ -1,37 +1,27 @@
...
+// provided list of available releases. This is a pure function - the caller is
+// responsible for fetching the releases list from the hetzner-k3s binary.
 //
-// Note: only the 100 most recent releases are fetched (no pagination).
-// Older Kubernetes versions may not resolve if there are 100+ newer releases.
-// Use an explicit k3s version string (e.g., "v1.28.5+k3s1") for older versions.
-func resolveK3sVersion(ctx context.Context, version string, apiURL string) (string, error) {
+// If version already contains "+k3s", it's returned as-is.
+// releases should be ordered newest-first for correct resolution.
+func resolveK3sVersion(ctx context.Context, version string, releases []string) (string, error) {
```
dcmcand, line 28: **Kudos**

Splitting I/O (`getHetznerK3sReleases`) from the pure resolution logic here is a clean response to dcmcand's feedback. Easy to test, easy to reason about. Good shape.

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +44,52 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
+		strings.Contains(tag, "-beta")
+}
+
+// getHetznerK3sReleases runs `hetzner-k3s releases` and returns the list of
+// supported version tags. Results are returned in the order the binary outputs
+// them (newest first). A 30-second timeout is applied to prevent hanging if
+// the binary makes network calls internally.
+func getHetznerK3sReleases(ctx context.Context, binaryPath string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
```
dcmcand, line 126: **Kudos**

`context.WithTimeout` + `exec.CommandContext` is the right pattern. Cancellation propagates correctly and the deferred cancel avoids leaking the timer.

dcmcand (changes requested):
Thanks for tackling this. The intent and overall structure are solid: using hetzner-k3s's own view of supported releases is the right fix for #205, and the I/O-vs-logic separation is cleaner than the original.

Two blockers need addressing before this can merge:

1. The "newest-first" ordering assumption is inverted. `hetzner-k3s releases` output is oldest-first. I verified this empirically against the v2.4.6 binary and traced it to `releases.reverse` in upstream `src/k3s.cr`. With the current code, `kubernetes_version: "1.34"` resolves to `v1.34.1+k3s1`, not `v1.34.6+k3s1`: a silent downgrade. I could not reproduce the PR's manual test claim.

2. The line parser in `getHetznerK3sReleases` accepts every non-empty line as a release, including the banner, `Version: X.Y.Z`, and the `Available k3s releases:` heading. It works today only because the downstream prefix filter catches the noise. Fragile to any future format change in hetzner-k3s.

Rest are questions, stylistic, and a couple of kudos on the cleanup.

pkg/provider/hetzner/k3s.go
tylerpotts: Fixed: Corrected the doc comment - releases are now documented as oldest-first (matching actual `hetzner-k3s releases` output). The loop now iterates in reverse (`for i := len(releases) - 1; i >= 0; i--`) to find the newest matching stable release. Test fixtures also updated to oldest-first order to match realistic conditions.

pkg/provider/hetzner/k3s.go
tylerpotts: Fixed: Added `k3sVersionPattern` regex (`^v\d+\.\d+\.\d+(-(?:rc|alpha|beta)\d*)?\+k3s\d+$`) to validate each line against a strict version pattern. Banner lines, headers, and malformed versions are now filtered out. Also added an error if zero valid lines match, so format drift fails loudly. Added `TestK3sVersionPattern` covering banner lines, valid tags, and edge cases.

pkg/provider/hetzner/k3s.go
tylerpotts: Fixed: Added OTel span to `getHetznerK3sReleases` with `binary.path` and `release.count` attributes. Errors are now recorded via `span.RecordError()` for better observability in production.

pkg/provider/hetzner/k3s.go
tylerpotts: Good point - increased timeout to 90 seconds to account for cold cache pagination through ~70 pages. Updated the doc comment to explain why: "A 90-second timeout is applied - the binary paginates through all k3s tags on a cache miss (~70 pages), which can be slow."

pkg/provider/hetzner/k3s.go
tylerpotts: Good catch - documented the side effect in the function doc: "Note: This function has a side effect - hetzner-k3s creates/updates a cache file at ~/.hetzner-k3s/k3s-releases.yaml (7-day TTL). The cache path is controlled by the binary, not by this code." Considered HOME override but decided against adding complexity since the binary controls the path internally.

pkg/provider/hetzner/k3s.go
tylerpotts: Fixed: Now using `strings.TrimSpace(stderr.String())` and `%q` formatting to quote-escape the stderr in error messages. Control characters and ANSI codes will be visible as escape sequences in logs.

pkg/provider/hetzner/k3s.go
tylerpotts: Fixed: Added `extractAvailableMinors()` helper that builds a list of unique minor versions from the releases. Error message now shows: "no supported k3s release found for kubernetes version \"1.99\" (available minor versions: 1.31, 1.32, 1.33)". Added test coverage for the helper.

pkg/provider/hetzner/k3s.go
tylerpotts: Fixed: Replaced `strings.Contains` with a strict regex pattern ``prereleasePattern = regexp.MustCompile(`-(?:rc|alpha|beta)\d*\+k3s`)``. This only matches pre-release suffixes between the patch version and `+k3s` suffix, avoiding false positives from hypothetical tags with "rc" elsewhere. Added test case to verify.

pkg/provider/hetzner/k3s_test.go
tylerpotts: Fixed: Updated test fixtures to oldest-first order matching real `hetzner-k3s releases` output. Added `TestK3sVersionPattern` with realistic banner/header lines from actual binary output to ensure the regex filtering catches format variations.

pkg/provider/hetzner/k3s.go
tylerpotts: Thanks! Glad the separation landed well.

pkg/provider/hetzner/k3s.go
tylerpotts: Thanks for the validation!

pkg/provider/hetzner/k3s.go
```diff
@@ -1,37 +1,33 @@
...
+// responsible for fetching the releases list from the hetzner-k3s binary.
 //
-// Note: only the 100 most recent releases are fetched (no pagination).
-// Older Kubernetes versions may not resolve if there are 100+ newer releases.
-// Use an explicit k3s version string (e.g., "v1.28.5+k3s1") for older versions.
-func resolveK3sVersion(ctx context.Context, version string, apiURL string) (string, error) {
+// If version already contains "+k3s", it's returned as-is.
+// releases are expected in oldest-first order (as output by `hetzner-k3s releases`);
+// the function iterates in reverse to find the newest matching stable release.
+func resolveK3sVersion(ctx context.Context, version string, releases []string) (string, error) {
```
dcmcand, line 28: Pure-function signature here is the key improvement from the prior round - caller owns I/O, this is testable without any binary or network. Nice.

pkg/provider/hetzner/k3s.go
```diff
@@ -1,37 +1,33 @@
...
-const k3sReleasesURL = "https://api.github.com/repos/k3s-io/k3s/releases?per_page=100"
-
-// ghRelease is the subset of GitHub release API response we need.
-type ghRelease struct {
-	TagName    string `json:"tag_name"`
-	Prerelease bool   `json:"prerelease"`
-}
+// k3sVersionPattern matches valid k3s release tags.
+// Examples: v1.32.12+k3s1, v1.32.12-rc1+k3s1, v1.32.12-alpha1+k3s1
+var k3sVersionPattern = regexp.MustCompile(`^v\d+\.\d+\.\d+(-(?:rc|alpha|beta)\d*)?\+k3s\d+$`)
```
dcmcand, line 18: Strict anchored regex is the right call - `TestK3sVersionPattern` below shows it correctly rejects the Unicode banner lines the binary emits, which is easy to get wrong with a looser pattern.

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +50,109 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
 	}
-	defer resp.Body.Close() //nolint:errcheck // Best-effort close on read-only response
+	return minors
+}
 
-	if resp.StatusCode != http.StatusOK {
-		err := fmt.Errorf("GitHub API returned status %d", resp.StatusCode)
+// getHetznerK3sReleases runs `hetzner-k3s releases` and returns the list of
+// supported version tags. Results are returned in the order the binary outputs
+// them (oldest-first). A 90-second timeout is applied - the binary paginates
```
dcmcand, line 112: Appreciate calling out the cache side effect and TTL in the doc comment - that's exactly the kind of thing that bites someone reading this in two years.

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +50,109 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
+	tracer := otel.Tracer("nebari-infrastructure-core")
+	ctx, span := tracer.Start(ctx, "hetzner.getHetznerK3sReleases")
+	defer span.End()
+
+	span.SetAttributes(attribute.String("binary.path", binaryPath))
+
+	ctx, cancel := context.WithTimeout(ctx, 90*time.Second)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, binaryPath, "releases")
```
dcmcand, line 128: `binaryPath` is trusted here because it comes from the provider's own install/download step, not user config. Might be worth a one-line comment asserting that invariant so a future refactor doesn't accidentally route a config-sourced path through here without re-validating.

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +50,109 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	if err := cmd.Run(); err != nil {
+		stderrStr := strings.TrimSpace(stderr.String())
 		span.RecordError(err)
-		return "", err
+		return nil, fmt.Errorf("failed to get hetzner-k3s releases: %w: %q", err, stderrStr)
```
dcmcand, line 136: Minor: when `stderrStr` is empty this formats as `...: ""`. Could conditionally drop the `: %q` segment or rephrase to `"failed to get hetzner-k3s releases (stderr: %q): %w"`. Not blocking.

pkg/provider/hetzner/k3s.go
```diff
@@ -58,45 +50,109 @@ func resolveK3sVersion(ctx context.Context, version string, apiURL string) (stri
...
-			continue
-		}
-		if strings.HasPrefix(r.TagName, matchPrefix) {
-			span.SetAttributes(attribute.String("resolved_version", r.TagName))
-			return r.TagName, nil
-		}
+	span.SetAttributes(attribute.Int("release.count", len(releases)))
+
+	if len(releases) == 0 {
+		err := fmt.Errorf("hetzner-k3s returned no valid release tags (output format may have changed)")
```
dcmcand, line 152: When this fires, the operator will want the raw output to debug. Consider `span.SetAttributes(attribute.String("raw_output_preview", firstN(stdout.String(), 500)))` before returning so the span carries the offending output.

pkg/provider/hetzner/k3s_test.go
```diff
@@ -73,30 +66,103 @@ func TestResolveK3sVersion(t *testing.T) {
 	}
 }
 
-func TestResolveK3sVersion_APIError(t *testing.T) {
+func TestResolveK3sVersion_SkipsAllPrereleaseTypes(t *testing.T) {
```
dcmcand, line 69: Good explicit coverage of alpha/beta/rc skipping in oldest-first order. This is the test that catches a regression of the prior ordering bug.

dcmcand (approved):
Nice iteration on this. The pure/IO split is clean and the ordering fix is well-guarded by tests.

**Highlights**
- `resolveK3sVersion([]string)` is now a pure function, fully unit-tested, and `getHetznerK3sReleases` holds the exec concern. Good separation.
- `k3sVersionPattern` is strict enough to filter banner/header noise from the binary output, and `TestK3sVersionPattern` covers realistic Unicode banners.
- The oldest-first ordering contract is documented in the function comment and mirrored in the test fixture, which will prevent the prior-round regression.
- 90s timeout with documented rationale (cold-cache pagination) is a thoughtful bound.
- Error messages include available minor versions and captured stderr, which will save operator time.
- OTel instrumentation is correct on both functions, and no `slog` leaked into `pkg/`.

**Non-blocking follow-ups to consider** (all safe to defer)
1. `getHetznerK3sReleases` and `runHetznerK3s` in `cluster.go` are two independent `exec.CommandContext` sites in the same package. Factoring a shared `runHetznerK3sOutput(ctx, binaryPath, args...)` helper would centralize stderr capture, timeouts, and unlock a mock seam. Pre-existing, not introduced here.
2. No unit test for the parsing side of `getHetznerK3sReleases`. Splitting out a `parseReleases(stdout []byte) []string` would let you exercise banner-filtering and the "no valid tags" branch without shelling out.
3. On the "no valid release tags" error path, consider stashing a short preview of the raw output as a span attribute for post-mortem visibility.

Approving - the fix is correct and the tests are in the right shape.

PR #233, opened 2026-04-08
Add Existing Cluster Provider

### LINKED ISSUE #188: Add `existing` cluster provider to support existing clusters outside of AWS
### Problem

Deploying to an existing cluster is currently wired to the AWS provider. The logic to skip infrastructure deployment and read `kube_context` lives inside the AWS provider
https://github.com/nebari-dev/nebari-infrastructure-core/blob/d33ed2e3df6d4144f49f26ef1a7c819ce6b7000a/pkg/provider/aws/provider.go#L114-L122

This means there's no way to deploy to an existing cluster outside of AWS.

The current config for existing clusters looks like:

```yaml
provider: aws
kube_context: "arn:aws:eks:us-west-2:123456789012:cluster/my-nebari-aws"
```

This forces users to pick `aws` as the provider even though no AWS infrastructure is being provisioned. The `kube_context` is a top-level field but only relevant to this mode, and provider-specific settings like storage class are silently inferred from the provider choice.

### Prerequisite

This depends on #186 (refactor cloud provider config to nested `cluster: <provider>:` format), which introduces the `ClusterConfig` pattern that the `existing` provider would use.

### Proposal

Add an `existing` cluster provider that handles pre-existing clusters as a first-class concept:

```yaml
cluster:
  existing:
    kube_context: "arn:aws:eks:us-west-2:123456789012:cluster/my-cluster"
    storage_class: longhorn
```

### Use case

`kube_context` is just a name in `~/.kube/config` and it works for any Kubernetes cluster regardless of where it runs (AWS EKS, Azure AKS, GCP GKE, bare-metal, kind, k3s). This would allow users to use NIC to deploy the foundational software to any existing k8s cluster.

marcelovilla:
## Closes

Closes #188 

## Description

This PR introduces an `existing` cluster provider, allowing users to deploy Nebari on top of an existing cluster.

The provider specific config block looks like this:

```yaml
cluster:
  existing:
    # kubeconfig is the path to the kubeconfig file. May be absolute or relative and
    # tilde (~) is not expanded. If not specified, defaults to the KUBECONFIG environment
    # variable or $HOME/.kube/config.
    kubeconfig: path/to/kubeconfig
    context: "my-cluster"
    storage_class: gp2
    load_balancer_annotations:
```

It follows the way tools like `kubectl` allow users to specify their cluster access information by exposing two fields: `kubeconfig`, containing the file with the kubeconfig information and `context`, specifying the name of the context to connect to. The former is optional and defaults to the `KUBECONFIG` env var or `~/.kube/config`, while the latter is required to avoid people accidentally deploying to their default context.

There are two additional optional fields exposed: `storage_class` and `load_balancer_annotations`, which might need to be specified depending on where the existing cluster is running.

## How to Test

1. Deploy a k8s cluster anywhere using other tool than NIC. In my case, I deployed to Hetzner using https://vitobotta.github.io/hetzner-k3s/
2. Create an `existing.yaml` file with a config similar to this. Make sure to update the `kubeconfig` and `context` files with the relevant file and context to connect to the newly created cluster. Also, update the `storage_class` and remove the `load_balancer_annotations` if the k8s cluster is not running on Hetzner.
  
    ```yaml
    project_name: test-existing
    domain: existing.domain.com
    
    cluster:
      existing:
        # Both kubeconfig and context are optional:
        #   - kubeconfig defaults to KUBECONFIG env or ~/.kube/config
        #   - context defaults to current-context from the kubeconfig file
        kubeconfig: ./kubeconfig
        context: "test-existing"
        storage_class: hcloud-volumes
        load_balancer_annotations:
          load-balancer.hetzner.cloud/location: hel1
    
    
    git_repository:
      url: "git@github.com:foobar/test-existing.git"
      branch: main
      path: "clusters/test-existing"
      auth:
        ssh_key_env: GIT_SSH_PRIVATE_KEY
    ```

3. Build the `nic` binary from this branch by running the following command:

    ```bash
    make build
    ````

4. Deploy Nebari in top of the existing cluster by running the following command:

    ```bash
    ./nic deploy -f existing.yaml
    ```


marcelovilla:
When running `nic destroy` on an existing cluster, it should remove everything it installed (i.e., ArgoCD, foundational apps, secrets, etc...). This is not implemented in this PR as the `argocd` pkg does not expose that functionality yet,and furthermore, it shouldn't be called from the cluster provider but rather from the destroy orchestrator (i.e. `cmd/destroy`). I created a follow-up issue to track this: https://github.com/nebari-dev/nebari-infrastructure-core/issues/240

examples/existing-config.yaml
```diff
@@ -0,0 +1,23 @@
+project_name: my-nebari-existing
+domain: nebari.example.com
+
+cluster:
+  existing:
+    # Kubeconfig defaults to KUBECONFIG env or ~/.kube/config
+    kubeconfig: ~/.kube/config
```
dcmcand: **Blocker:** literal `~/.kube/config` won't be expanded by `clientcmd.LoadFromFile`. Verified:

```
LoadFromFile(~/.kube/config) err: open ~/.kube/config: no such file or directory
```

Users copying the example verbatim will hit `failed to load kubeconfig from ~/.kube/config: ... no such file or directory`, which is confusing because the comment one line above explicitly says "Kubeconfig defaults to KUBECONFIG env or ~/.kube/config".

Fix options:
1. Expand `~` in `Config.GetKubeconfigPath()` using `os.UserHomeDir()` — preserves the example as written and matches what users intuitively expect.
2. Drop the explicit `kubeconfig:` line from this example (the empty default already resolves to `KUBECONFIG` env → `~/.kube/config` correctly).

I'd lean toward option 1 since the inline comment promises `~` works.

pkg/provider/existing/provider.go
```diff
@@ -0,0 +1,239 @@
...
+	if cfg.Context == "" {
+		return "", "", fmt.Errorf("context is required for existing cluster provider")
+	}
+
+	path, err := cfg.GetKubeconfigPath()
+	if err != nil {
+		return "", "", err
+	}
+
+	kubeconfigData, err := clientcmd.LoadFromFile(path)
```
dcmcand: **Question:** any reason to call `clientcmd.LoadFromFile(path)` directly here instead of `kubeconfig.LoadFromPath(path)`? `GetKubeconfig` (line 166) uses the helper, so the two paths are inconsistent. Bonus: `pkg/kubeconfig/kubeconfig.go` already has a `ValidateContext(name)` function — it just needs to take a path arg to be reusable here, which would let `Validate` and `GetKubeconfig` share the load+lookup logic instead of `loadAndValidateContext` re-implementing it.

pkg/provider/existing/provider.go
```diff
@@ -0,0 +1,239 @@
...
+	defer span.End()
+
+	existingCfg, err := extractConfig(ctx, clusterConfig)
+	if err != nil {
+		span.RecordError(err)
+		return nil, err
+	}
+
+	if existingCfg.Context == "" {
+		err := fmt.Errorf("context is required for existing cluster provider")
```
dcmcand: **Stylistic:** `err :=` here shadows the `err` declared on line 148. Confirmed via `shadow`:

```
pkg/provider/existing/provider.go:155:3: declaration of "err" shadows declaration at line 148
```

`err = fmt.Errorf(...)` would avoid the shadow. Also, since `Validate` already enforces non-empty `Context`, this check is defensive redundancy in the call ordering — fine to keep, but worth noting.

pkg/provider/existing/provider.go
```diff
@@ -0,0 +1,239 @@
...
+		return nil, fmt.Errorf("failed to load kubeconfig from %s: %w", path, err)
+	}
+
+	span.SetAttributes(
+		attribute.String("provider", ProviderName),
+		attribute.String("kubeconfig_path", path),
+		attribute.String("kube_context", existingCfg.Context),
+	)
+
+	filtered, err := kubeconfig.FilterByContext(kubeconfigData, existingCfg.Context)
```
dcmcand, line 177: Nice — using `FilterByContext` to scope the returned kubeconfig to only the selected context (and its referenced cluster + user) avoids leaking unrelated credentials from a multi-context kubeconfig into downstream consumers. Good defensive choice.

pkg/provider/existing/config.go
```diff
@@ -0,0 +1,45 @@
...
+// Config represents configuration for connecting to a pre-existing Kubernetes cluster.
+type Config struct {
+	// Kubeconfig is the path to the kubeconfig file.
+	// Defaults to KUBECONFIG env or ~/.kube/config when empty.
+	Kubeconfig string `yaml:"kubeconfig,omitempty"`
+
+	// Context is the name of the context entry in the kubeconfig file.
+	// Required — must be explicitly set to avoid accidentally deploying
+	// to the wrong cluster.
+	Context string `yaml:"context"`
```
dcmcand, line 16: Appreciate the explicit "required" rationale in the comment — `context` defaulting to `current-context` would be a footgun (deploy lands wherever your kubectl was last pointed). Forcing the user to spell it out is the right call.

pkg/provider/existing/provider_test.go
```diff
@@ -0,0 +1,248 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/config"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/provider"
+)
+
+// Compile-time interface compliance check
+var _ provider.Provider = (*Provider)(nil)
```
dcmcand, line 15: Good — `var _ provider.Provider = (*Provider)(nil)` catches interface drift at compile time. Worth standardizing across the other provider packages too if not already done.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -12,21 +12,25 @@ import (
 
 // GetPath returns the path to the kubeconfig file.
 // It checks the KUBECONFIG environment variable first, then falls back to ~/.kube/config.
-func GetPath() string {
+func GetPath() (string, error) {
 	if kubeconfigEnv := os.Getenv("KUBECONFIG"); kubeconfigEnv != "" {
```
dcmcand, line 16: Good change — the previous version silently fell back to a relative `.kube/config` path on `os.UserHomeDir()` failure, which would mostly produce confusing "file not found" errors at the call site. Returning the error here makes the failure surface where it can be reported.

dcmcand (commented):
Nice cleanup of the provider interface. Narrowing `Provider` methods to take `projectName` + `ClusterConfig` instead of the full `NebariConfig` is a real improvement — providers no longer see DNS, certs, or git settings, and the new `existing` provider drops in cleanly as a result. Tests are well-structured (table-driven, compile-time interface assertion, happy + error paths covered).

One blocker on the example config (literal `~` won't expand) and one design question about destroy semantics. Inline comments below.

**Issue #188 resolution:** Fully resolved (pending the example fix). The proposed YAML shape, `kubeconfig` + `context` semantics, and `storage_class` placement all match the issue.

### General feedback (not tied to a line)

**Destroy semantics for existing clusters.** `nic destroy -f existing.yaml` today: shows the standard `⚠️ WARNING: You are about to destroy the following infrastructure ... ❌ This will permanently delete all resources and data` confirmation, user types `yes`, the no-op `Destroy` returns, and the CLI prints `Destruction completed successfully`. But `deploy` installed ArgoCD + Keycloak + foundational services onto the user's cluster, and nothing in the destroy path uninstalls them — they stay behind. Two issues worth thinking about:

1. The warning text is misleading for this provider — nothing of the cluster is actually being destroyed.
2. The foundational software lifecycle becomes asymmetric: install via `deploy`, no removal path via `destroy`.

Reasonable to defer to a follow-up issue (this PR is already large), but worth deciding the intended behavior. Two paths I see: (a) tailor `confirmDestruction` output for existing clusters so users know the cluster itself is untouched, or (b) have `Destroy` actually uninstall what `Deploy` installed.

**Silent error swallowing in `Summary` / `InfraSettings`.** Both swallow `UnmarshalProviderConfig` errors and return defaults. Consistent with AWS/other providers so not a regression, just noting that misconfig will silently degrade. A debug-level slog or span error would help diagnose. Out of scope for this PR.

examples/existing-config.yaml
marcelovilla: Good catch, thanks @dcmcand! This got me digging a little bit deeper into the `~` expansion and before making any decision, I wanted to see if we were supporting it in any other place in our config. Hetzner exposes a way of specifying the ssh keys and our example also use `~` but it does not handle expansion, leading to an error (I opened https://github.com/nebari-dev/nebari-infrastructure-core/issues/257 to track that). I decided to go with your second option for now as I'm unsure whether we'll want to support path expansion and I think a comment showing it defaults to `$HOME/.kube/config` (just like [kubectl does](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/)) makes things clearer.

pkg/provider/existing/provider.go
marcelovilla: Thanks for calling this out. You're right it's inconsistent and I've addressed this in 275549b319a94b323d1acf29df6c0b39f602fe92

pkg/provider/existing/provider.go
marcelovilla: Good catch. I addressed this in fd09c971390be79133d72b2b6bc2522cba054ec5 by removing the check.

> Also, since Validate already enforces non-empty Context, this check is defensive redundancy in the call ordering

This is actually not true because the deploy cmd won't call validate on the provider but I still removed it because:

1) Trying t filter the context will error out saying the empty context was not found
2) Once we enforce config-based validation, that field will be required and it needs to be non-empty

marcelovilla:
Thanks for the review @dcmcand, this is ready for a second pass.

Regarding your concerns about the destroy semantics for existing clusters, I purposefully left that out of this PR and opened https://github.com/nebari-dev/nebari-infrastructure-core/issues/240 to discuss what destroy should look like on an existing provider and how we would implement that.

pkg/provider/existing/provider.go
```diff
@@ -0,0 +1,226 @@
...
```
dcmcand, line 177: **Question:** `GetKubeconfig` doesn't re-validate that `existingCfg.Context` is non-empty before passing it to `FilterByContext`. All current callers (`cmd/nic/deploy.go`, `cmd/nic/kubeconfig.go`, `pkg/argocd/foundational.go`, `pkg/argocd/install.go`) go through `Validate` first, so this is safe in practice. But if someone ever calls `GetKubeconfig` without `Validate` the user will see `context "" not found in kubeconfig...`, which is a confusing error for a usage bug. Intentional to keep `GetKubeconfig` dependency-free, or worth a one-line guard that mirrors `loadAndValidateContext`?

pkg/kubeconfig/kubeconfig.go
```diff
@@ -12,21 +12,25 @@ import (
 
 // GetPath returns the path to the kubeconfig file.
 // It checks the KUBECONFIG environment variable first, then falls back to ~/.kube/config.
-func GetPath() string {
+func GetPath() (string, error) {
```
dcmcand, line 15: **Question:** The signature change to `GetPath()` (and the related change to `ValidateContext()` below) is a breaking API change to a shared helper package. Only `pkg/provider/local/provider.go` is updated in this PR to match, which is sufficient for build correctness, but these look like two independently useful changes smuggled inside a feature PR. Was splitting into a precursor PR considered, or is bundling them here deliberate? No action needed either way, just flagging.

examples/existing-config.yaml
```diff
@@ -0,0 +1,25 @@
+project_name: my-nebari-existing
+domain: nebari.example.com
+
+cluster:
+  existing:
+    # kubeconfig is the path to the kubeconfig file. May be absolute or relative and
+    # tilde (~) is not expanded. If not specified, defaults to the KUBECONFIG environment
+    # variable or $HOME/.kube/config.
+    kubeconfig: path/to/kubeconfig
```
dcmcand, line 9: **Question:** The example uses a literal relative path and the comment above says "tilde (~) is not expanded", but the PR description example uses `kubeconfig: ~/.kube/config` which would NOT be expanded. The PR framing ("just like kubectl") implies tilde works. Should the PR-description example be updated to match this comment, or should the code path expand `~`?

pkg/provider/existing/provider.go
```diff
@@ -0,0 +1,226 @@
...
+	if rawCfg == nil {
+		return result
+	}
+
+	var existingCfg Config
+	if err := config.UnmarshalProviderConfig(context.Background(), rawCfg, &existingCfg); err != nil {
+		return result
+	}
+
+	result["Mode"] = "existing-cluster"
```
dcmcand: **Stylistic:** Other providers' `Summary()` methods use `"Provider"` as the key (e.g., `pkg/provider/hetzner/provider.go:361` uses `"Provider": "Hetzner Cloud (k3s)"`). Using `"Mode"` here means the destroy-confirmation output is slightly inconsistent across providers. Not blocking.

pkg/provider/existing/provider.go
```diff
@@ -0,0 +1,226 @@
...
+		span.RecordError(err)
+		return nil, fmt.Errorf("failed to unmarshal existing cluster config: %w", err)
+	}
+
+	return &existingCfg, nil
+}
+
+// loadAndValidateContext loads the kubeconfig file and validates that the
+// required context exists. Returns the kubeconfig path and context name.
+func loadAndValidateContext(cfg *Config) (string, string, error) {
```
dcmcand: **Stylistic:** `loadAndValidateContext` does real I/O (two filesystem reads via `kubeconfig.ValidateContext` -> `LoadFromPath`) but doesn't accept `ctx` or create a span. CLAUDE.md says all new functions must be instrumented, and existing precedent treats trivial private helpers as exempt - this one sits closer to the 'real function' end of that spectrum. Cheap to add `ctx context.Context` + `tracer.Start`, matching `extractConfig()` just above. Not blocking.

pkg/provider/existing/provider.go
```diff
@@ -0,0 +1,226 @@
...
+	defer span.End()
+
+	span.SetAttributes(
+		attribute.String("provider", ProviderName),
+		attribute.String("project_name", projectName),
+	)
+
+	existingCfg, err := extractConfig(ctx, clusterConfig)
+	if err != nil {
+		span.RecordError(err)
```
dcmcand, line 99: **Stylistic:** `span.RecordError(err)` is called here and also inside `extractConfig()`, so the same error ends up on the span twice (same pattern in the block just below for `loadAndValidateContext`). Other providers in the codebase have this same double-record pattern, so flagging for awareness rather than change. Not blocking.

dcmcand (dismissed):
Nice work on this one, Marcelo. The design matches what #188 asks for and the cleanup of `NebariConfig.KubeContext` / `IsExistingCluster()` avoids leaving a deprecated shim everywhere. Build, `go vet`, and tests in the affected packages all pass locally. A few positives worth calling out:

- `var _ provider.Provider = (*Provider)(nil)` compile-time check in the test file, matching the pattern in other providers.
- `GetPath()` now returns `(string, error)` and propagates the home-dir lookup failure instead of silently falling back to a relative `.kube/config` - that's a real bug fix hidden in the refactor.
- `FilterByContext` before emitting kubeconfig bytes is the right call: deploy-time code gets a single-context kubeconfig rather than the user's full credentials file.
- Registration follows the existing explicit-registry pattern (no blank imports).

I've left a few questions and stylistic notes inline. **None of the stylistic comments are blockers** - feel free to fold them in here or open follow-up issues, whichever you prefer.

A few general notes not tied to a specific line:

1. **PR description drift:** The description says `context` "defaults to current-context from the kubeconfig file" but the implementation (and the `missing context` test case) rejects an empty context. The strict stance is defensible per the safety argument in the `Config` struct comment; worth updating the PR body (and release notes when cut) so users aren't surprised.
2. **Silent config migration:** Users with an existing top-level `kube_context: <arn>` in their config will silently have that field ignored after this change - `NebariConfig.KubeContext` is gone and no validation catches unknown top-level yaml keys. Worth a migration note in the PR description / release notes.
3. **Test coverage gaps (non-blocking):** No test for `GetKubeconfig` with a nonexistent kubeconfig path, `Summary` / `InfraSettings` with `nil` ProviderConfig (the code handles it but doesn't exercise it), or `GetKubeconfig` when the context exists but its referenced cluster/user entry is missing (`FilterByContext` currently silent-skips those and returns a credential-less kubeconfig - could be worth a defensive test or an explicit error).
4. **Supply chain:** No new dependencies, no new `init()` functions, no `os/exec`, no embedded files or build tags. File list matches the stated scope. Clean.

Approving.

pkg/provider/existing/provider.go
marcelovilla: Actually, `cmd/nic/deploy.go` does not call validate first and you will run into this error when trying to deploy. Removing it was intentional and I [commented my reasoning behind this previously](https://github.com/nebari-dev/nebari-infrastructure-core/pull/233#discussion_r3099081069).

examples/existing-config.yaml
marcelovilla: Updated the PR description to match the example and the implementation. As [I already commented](https://github.com/nebari-dev/nebari-infrastructure-core/pull/233#discussion_r3092070454), I don't think we should support expanding `~` at this moment.

pkg/provider/existing/provider.go
marcelovilla: Thanks for calling this out. Upon further investigation, it seems we're not following any convention across providers: AWS, Hetzner, and the existing provider all have different things and Hetzner is the only one specifying a provider. I changed it to `result["Provider"]` in 18b1c13c76e8551d6184889fb086b6a7f42cf761 to be consistent with Hetzner but I think at some point we'll want to standardize this a bit.

pkg/provider/existing/provider.go
marcelovilla: Addressed in 3a4a316e3d79d4ee00c3bc1e76e1ac1370ed4469

pkg/provider/existing/provider.go
marcelovilla: Good point and something we should standardize across providers in the future.

pkg/kubeconfig/kubeconfig.go
marcelovilla: None of the other providers use the `GetPath` function so there was no need to update it outside of the `local` provider. I did not consider splitting this into a precursor PR as it was a relatively minor change affecting the implementation of the new existing provider.

marcelovilla:
@dcmcand I addressed all your new inline comments and here are my answers to your general comments:

> **PR description drift**

I just updated the PR description

> **Silent config migration**

We're not supporting existing cluster at this point as we're constantly introducing breaking changes. Once we have a stable release, I agree we should take care of migration paths but right now it seems premature.

> **Test coverage gaps (non-blocking)**

I think you raise a good point here. I'm not sure if it should be responsibility of the `kubeconfig` package or the provider itself to check this, but I opened https://github.com/nebari-dev/nebari-infrastructure-core/issues/263 as a follow-up.

dcmcand (approved):
🚀 lgtm @marcelovilla

PR #234, opened 2026-04-08
ArgoCD SSO via Keycloak OIDC

dcmcand:
## Summary

Closes #227

- Configure ArgoCD with Keycloak OIDC SSO as part of initial deployment
- Create `argocd-admins` (full admin) and `argocd-viewers` (read-only) Keycloak groups mapped to ArgoCD RBAC roles
- Pre-generate OIDC client secret in Go, inject into ArgoCD Helm values and store as K8s secret for realm-setup job
- Extend realm-setup job to create ArgoCD OIDC client, groups, and add realm admin to argocd-admins
- Update post-deploy instructions with SSO login info and break-glass admin fallback

## How it works

1. `deploy.go` generates a client secret upfront
2. ArgoCD is installed via Helm with OIDC config (`ConfigWithOIDC()`) referencing Keycloak's issuer URL and the client secret
3. The secret is also stored as `argocd-oidc-client-secret` in the keycloak namespace
4. When Keycloak deploys (sync wave 4), the PostSync realm-setup job reads the secret and creates the matching OIDC client, groups, and group membership
5. ArgoCD's OIDC discovery is lazy, so SSO starts working once Keycloak is ready

## Test plan

- [x] Unit tests pass: `go test ./... -v`
- [x] Linter clean: `golangci-lint run`
- [x] Deploy to local cluster, verify ArgoCD shows "Log in via Keycloak" button
- [x] Log in as realm admin, verify admin access (argocd-admins group)
- [x] Create user in argocd-viewers group, verify read-only access
- [x] Create user not in either group, verify access denied
- [x] Re-run deploy (idempotency), verify no errors


viniciusdc (approved):
Tested this end-to-end on a fresh kind cluster from the PR branch. Walked through every item in the test plan and the contract holds.

Captura de Tela 2026-05-11 às 12 59 35

The load-bearing claim (pre-generate the OIDC client secret once at deploy time, distribute to both Keycloak and ArgoCD) works correctly. Across all three sites the value stays byte-identical:

- `argocd-oidc-client-secret` in `keycloak` ns
- `oidc.keycloak.clientSecret` in `argocd-secret`
- The actual `argocd` client secret stored in Keycloak (verified via `kcadm get clients/<id>/client-secret`)

<details>
<summary> list of created mock users </summary>

  Username: admin
  Group: argocd-admins
  Expected in ArgoCD: full admin (can sync, edit, delete apps)
  ────────────────────────────────────────
  Username: argocd-viewer
  Group: argocd-viewers
  Expected in ArgoCD: read-only (sees apps but can't sync or edit)
  ────────────────────────────────────────
  Username: argocd-nobody
  Group: (none)
  Expected in ArgoCD: logs in successfully but lands on a page with no apps visible / "permission denied" (see image bellow)

</details>

Captura de Tela 2026-05-11 às 13 12 46

RBAC enforcement is correct server-side. From `argocd-viewer` (member of `argocd-viewers` only):

```
Unable to sync cert-manager: permission denied: applications, sync,
  foundational/cert-manager, sub: c757be83-6b36-4633-a3af-3c6de016e31c

Unable to delete application: permission denied: applications, delete,
  foundational/cert-manager, sub: c757be83-6b36-4633-a3af-3c6de016e31c
```

Re-ran `nic deploy` to validate idempotency. Keycloak pod kept its uptime (0 restarts) and the secret values stayed byte-identical across both runs. `createSecret` being no-op-if-exists is doing the right thing.

One caveat worth flagging for future testers, not a PR finding:

> The ArgoCD UI renders all action buttons (Sync, Delete, etc.) regardless of the logged-in user's role. The confirmation modal is purely client-side. Authorization is enforced server-side, at the moment you actually click "Confirm Delete." A readonly user would get an HTTP 403 at that point. This is a long-standing UX quirk in ArgoCD.

I lost a bit of time the first round chasing what looked like an RBAC bug before realizing the buttons just always render. Worth a line in the post-deploy instructions or test plan.

LGTM, thanks @dcmcand!

PR #236, opened 2026-04-09
feat: create EFS StorageClass when EFS is enabled

dcmcand:
## Summary

Closes #235

- Add `StorageClassName` field to `EFSConfig` (default: `efs-sc`) and create a Kubernetes StorageClass backed by the EFS CSI driver (`efs.csi.aws.com`) after `tf.Apply()` completes
- Expose `EFSStorageClass` in `InfraSettings` so software packs can discover and use it for ReadWriteMany volumes (e.g., shared LLM model weights)
- Follow the existing Longhorn pattern: separate `WithClient` function for testability, idempotent create-or-update, OpenTelemetry instrumented
- Bump terraform-aws-eks-cluster to v0.2.0, which installs the EFS CSI driver addon

## Test plan

- [x] Table-driven unit tests for `EFSStorageClassName()` (nil, empty, custom)
- [x] Table-driven unit tests for `createEFSStorageClassWithClient()` (create, update, custom name)
- [x] All existing tests pass
- [x] `golangci-lint run` clean
- [x] `go vet` clean
- [ ] Manual: deploy with EFS enabled, verify StorageClass created with correct `fileSystemId`
- [ ] Manual: deploy twice, verify idempotent update

### DIFF
--- a/pkg/provider/aws/config.go
+++ b/pkg/provider/aws/config.go
@@ -75,7 +75,8 @@ type EFSConfig struct {
 const defaultEFSStorageClassName = "efs-sc"
 
 // EFSStorageClassName returns the StorageClass name for EFS volumes.
-// Defaults to "efs-sc" when not set.
+// Returns the default "efs-sc" when EFS is not configured or when
+// StorageClassName is empty.
 func (c *Config) EFSStorageClassName() string {
 	if c.EFS == nil || c.EFS.StorageClassName == "" {
 		return defaultEFSStorageClassName
--- a/pkg/provider/aws/config.go
+++ b/pkg/provider/aws/config.go
@@ -69,6 +69,18 @@ type EFSConfig struct {
 	ProvisionedThroughput int    `yaml:"provisioned_throughput_mibps,omitempty"`
 	Encrypted             bool   `yaml:"encrypted,omitempty"` // default: true
 	KMSKeyArn             string `yaml:"kms_key_arn,omitempty"`
+	StorageClassName      string `yaml:"storage_class_name,omitempty"` // default: efs-sc
+}
+
+const defaultEFSStorageClassName = "efs-sc"
+
+// EFSStorageClassName returns the StorageClass name for EFS volumes.
+// Defaults to "efs-sc" when not set.
+func (c *Config) EFSStorageClassName() string {
+	if c.EFS == nil || c.EFS.StorageClassName == "" {
+		return defaultEFSStorageClassName
+	}
+	return c.EFS.StorageClassName
 }
 
 type LonghornConfig struct {
--- a/pkg/provider/aws/efs.go
+++ b/pkg/provider/aws/efs.go
@@ -1,0 +1,112 @@
+package aws
+
+import (
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/status"
+)
+
+const (
+	efsCSIProvisioner = "efs.csi.aws.com"
+)
+
+// createEFSStorageClass creates or updates a Kubernetes StorageClass for EFS
+// dynamic provisioning using access points. This requires the EFS CSI driver
+// to be installed on the cluster (handled by the Terraform EKS module).
+func createEFSStorageClass(ctx context.Context, kubeconfigBytes []byte, cfg *Config, efsID string) error {
+	tracer := otel.Tracer("nebari-infrastructure-core")
+	ctx, span := tracer.Start(ctx, "aws.createEFSStorageClass")
+	defer span.End()
+
+	storageClassName := cfg.EFSStorageClassName()
+
+	span.SetAttributes(
+		attribute.String("storage_class_name", storageClassName),
+		attribute.String("efs_id", efsID),
+	)
+
+	client, err := newEFSK8sClient(kubeconfigBytes)
+	if err != nil {
+		span.RecordError(err)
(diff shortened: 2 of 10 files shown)


pkg/provider/aws/config.go
```diff
@@ -69,6 +69,19 @@ type EFSConfig struct {
...
+const defaultEFSStorageClassName = "efs-sc"
+
+// EFSStorageClassName returns the StorageClass name for EFS volumes.
+// Returns the default "efs-sc" when EFS is not configured or when
+// StorageClassName is empty.
+func (c *Config) EFSStorageClassName() string {
+	if c.EFS == nil || c.EFS.StorageClassName == "" {
+		return defaultEFSStorageClassName
+	}
+	return c.EFS.StorageClassName
```
marcelovilla, line 84: The `StorageClassName` field doesn't seem necessary here. We create the StorageClass ourselves and expose it via `InfraSettings.EFSStorageClass`. A configurable name would make sense if we were referencing a pre-existing StorageClass, but since we own the whole lifecycle, can't we just hardcode `efs-sc`?

dcmcand:
CI `drift` check is currently failing because the OpenTofu registry hasn't synced `nebari-dev/eks-cluster/aws` v0.2.0 yet. The tag was published to GitHub today (2026-04-13) and the registry's docs metadata already lists it, but the `/v1/modules/.../versions` API still only returns 0.1.0, so `tofu init` can't resolve the `version = "0.2.0"` constraint.

No code change needed - re-running CI once the registry finishes syncing should clear the check.

pkg/provider/aws/k8s.go
```diff
@@ -0,0 +1,19 @@
...
+// newK8sClient creates a Kubernetes clientset from kubeconfig bytes.
+// Used by post-deploy steps (Longhorn, EFS StorageClass) that need to
+// interact with the cluster after terraform provisioning.
+func newK8sClient(kubeconfigBytes []byte) (*kubernetes.Clientset, error) {
+	restConfig, err := clientcmd.RESTConfigFromKubeConfig(kubeconfigBytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse kubeconfig: %w", err)
+	}
+	return kubernetes.NewForConfig(restConfig)
+}
```
marcelovilla, line 19: Small nit and non-blocker: we have four similar implementations for creating a k8s client for kubeconfig bytes.

- This one
- https://github.com/nebari-dev/nebari-infrastructure-core/blob/ecb57d1aa3611edc96dafb8f184446b40fd0ec38/pkg/argocd/foundational.go#L160-L166
- https://github.com/nebari-dev/nebari-infrastructure-core/blob/ecb57d1aa3611edc96dafb8f184446b40fd0ec38/cmd/nic/deploy.go#L247-L251
- https://github.com/nebari-dev/nebari-infrastructure-core/blob/ecb57d1aa3611edc96dafb8f184446b40fd0ec38/pkg/argocd/applications.go#L30-L40

I know this PR at least avoids having two more inside the AWS provider and consolidates it into one. I wonder if we should have a helper function outside of the provider so all pkgs can use it.

pkg/provider/aws/efs.go
```diff
@@ -0,0 +1,112 @@
...
+		status.Send(ctx, status.NewUpdate(status.LevelInfo, "Updating existing EFS StorageClass").
+			WithResource("efs-storageclass").
+			WithAction("updating").
+			WithMetadata("name", storageClassName))
+		sc.ResourceVersion = existing.ResourceVersion
+		if _, err := client.StorageV1().StorageClasses().Update(ctx, sc, metav1.UpdateOptions{}); err != nil {
+			span.RecordError(err)
+			return fmt.Errorf("failed to update EFS StorageClass: %w", err)
+		}
+	}
```
marcelovilla, line 104: Storage class fields, such as `parameters`, seem to be immutable: https://github.com/kubernetes/kubernetes/blob/a3895062546711ae2666848a7930ae3b4ffae220/pkg/apis/storage/validation/validation.go#L65.

Although it seems like an unlikely case, if the filesystem id changes (maybe because users removed it from the config and then add it again), the update will be rejected by the API. Maybe we should compare the parameters and do a delete-then-create when they differ.

This is definitely not a blocker.

pkg/provider/aws/config.go
dcmcand: ~~if someone wants to use a different efs storage class than the one we provision, i think it makes sense to let them.~~

pkg/provider/aws/k8s.go
dcmcand: Agreed, good catch - thanks for flagging it. `pkg/kubeconfig` already exists and is the natural home for a shared `NewClientset(kubeconfigBytes []byte)` helper. Keeping that out of this PR to keep the diff focused on EFS, and tracking it in #247 as a follow-up.

pkg/provider/aws/config.go
marcelovilla: But if they want to use a different efs storag class, they wouldn't pass the `efs` block on their config file, which means we wouldn't create the storage class anyways.

pkg/provider/aws/efs.go
dcmcand: Good catch, thanks for digging up the validation source. You're right that the unconditional `Update()` will be rejected by the API if `fileSystemId` (or any other immutable field) changes. Tracking the fix in #248 - there are some non-trivial design questions around whether to no-op, fail loudly, or gate destructive replacement behind a flag, so I'd rather work it out in a focused PR than expand this one.

pkg/provider/aws/config.go
dcmcand: ~~@marcelovilla  They may want that efs for something else (like jupyter) and a different efs for this, for example, they want a standard efs for jupyter, but want faster iops for this one since big models benefit from that. so they provision the models efs after having deployed NIC already. or they just want a separate system to make billing easier. Since it is an optional field with a default value, i think most teams can ignore it, but if they need it, it is there. Generally i think exposing config fields, but making them optional with sane defaults fits our stated philosophy of "A platform with opinionated defaults that can be easily customized to your environment"~~

dcmcand: @marcelovilla disregard what i said above. I am working on the llmd pack so had that on the brain.

This is much simpler. I want this to be configurable because orgs may have conventions on how they name k8s storage classes. I think exposing this while giving a reasonable default makes sense here

marcelovilla (approved):
Thanks for the PR @dcmcand and for addressing my comments!

PR #245, opened 2026-04-11
feat: redesign CLI output with renderer abstraction and --format flag

### LINKED ISSUE #244: proposal: redesign CLI output with phase-oriented display and --output flag
## Summary

NIC's current CLI output is a mix of raw JSON structured logs (`slog` to stderr), interleaved third-party tool output (hetzner-k3s, OpenTofu), and ASCII box-art banners. The result is hard to follow during a deploy — operators see a wall of `{"level":"INFO","msg":"Status","message":"..."}` lines with no visual hierarchy, progress indication, or color.

**Proposal:** Replace the default output with a phase-oriented, tree-structured display while preserving full backward compatibility via `--output` flag and automatic TTY detection. Full design doc: [`docs/design-doc/nic-logging-redesign.md`](https://github.com/nebari-dev/nebari-infrastructure-core/blob/main/docs/design-doc/nic-logging-redesign.md).

## Current output

```
{"level":"INFO","msg":"Starting deployment","config_file":"config.yaml"}
{"level":"INFO","msg":"Configuration parsed successfully","provider":"hetzner","project_name":"nic-nebari"}
{"level":"INFO","msg":"Status","message":"Creating Hetzner k3s cluster","resource":"provider","action":"deploy"}
[Instance nic-nebari-master1] Instance status: running
[Instance nic-nebari-master1] Installing k3s...
{"level":"INFO","msg":"Infrastructure deployment completed","provider":"hetzner"}
{"level":"INFO","msg":"Installing Argo CD on cluster"}
{"level":"INFO","msg":"Success","message":"Argo CD is ready","resource":"argocd","action":"ready"}
```

## Proposed output (TTY / pretty mode)

```
nebari v0.2.0

  Deploying nic-nebari (hetzner)

  ✓ Configuration validated                                           0.2s

  ● Infrastructure                                                   3m 42s
    ├─ ✓ SSH keys verified                                            0.8s
    ├─ ✓ k3s version resolved (v1.32.13+k3s1)                        1.2s
    ├─ ✓ Cluster created                                             3m 38s
    │     master1 (10.0.0.2) ✓  worker1 (10.0.0.3) ✓
    └─ ✓ Kubeconfig saved                                             0.4s

  ● GitOps                                                            4.1s
    ├─ ✓ Repository cloned                                            2.8s
    ├─ ✓ Manifests generated                                          0.9s
    └─ ✓ Changes pushed                                               0.4s

  ● ArgoCD                                                           28.3s
    ├─ ✓ Helm chart installed (v9.4.1)                               12.1s
    ├─ ✓ Project created (nebari)                                     0.3s
    └─ ✓ Root app-of-apps applied                                    15.9s

  ● Foundational Services                                            1m 12s
    ├─ ✓ cert-manager                                  Healthy        18.4s
    ├─ ✓ envoy-gateway                                 Healthy        22.1s
    ├─ ✓ keycloak                                      Healthy        31.2s
    └─ ✓ nebari-operator                               Healthy         0.5s

  ● DNS                                                               2.4s
    ├─ ✓ nebari.example.com → 1.2.3.4                                 1.2s
    └─ ✓ *.nebari.example.com → 1.2.3.4                               1.2s

  ✓ Deployment complete                                              5m 29s

  ┌─────────────────────────────────────────────────────────────┐
  │  ArgoCD    https://argocd.nebari.example.com                │
  │  Keycloak  https://keycloak.nebari.example.com              │
  │                                                             │
  │  Kubeconfig: ~/.cache/nic/hetzner-k3s/nic-nebari/kubeconfig │
  └─────────────────────────────────────────────────────────────┘
```

## Proposed output (error case)

```
  ● ArgoCD                                                           12.1s
    ├─ ✓ Helm chart installed (v9.4.1)                               12.1s
    └─ ✗ Project creation failed                                      0.3s

      Error: failed to create ArgoCD project "nebari":
      admission webhook "validate.argocd" denied the request:
      project "nebari" already exists

      Hint: Run `nic destroy` first, or use `--force` to overwrite.

  ✗ Deployment failed                                                3m 54s
```

## Output mode flag (`--output`)

To ensure backward compatibility and support terminals that don't render Unicode/ANSI well, a global `--output` flag controls the rendering mode:

```
nic deploy -f config.yaml --output=pretty   # tree-structured, colored (default on TTY)
nic deploy -f config.yaml --output=json     # current JSON structured logs (default on pipe/CI)
nic deploy -f config.yaml --output=plain    # human-readable but no color/unicode (for dumb terminals)
```

**Auto-detection (default):**
- `stdout` is a TTY → `pretty`
- `stdout` is a pipe or `NO_COLOR` env is set → `json`

**Override:** `--output` flag always wins over auto-detection. This ensures:
- Users on terminals that don't handle Unicode can force `--output=plain`
- CI pipelines that want pretty logs (e.g., for GitHub Actions summaries) can force `--output=pretty`
- Scripts that parse JSON can force `--output=json` regardless of TTY state

The `plain` mode is identical to `pretty` but uses ASCII-only characters (`+`/`x`/`*` instead of `✓`/`✗`/`●`, `-` instead of `─`, `|` instead of `│`) and no ANSI color codes.

### Architecture

```
cmd/nic/
├── main.go              # --output flag + isatty check → select renderer
├── status_handler.go    # current: slog bridge → new: renderer dispatch
└── renderer/
    ├── renderer.go      # Renderer interface
    ├── pretty.go        # TTY renderer (pterm-based, color + unicode)
    ├── plain.go         # ASCII-only, no color (extends pretty with different symbols)
    └── json.go          # Pipe/CI renderer (current slog JSON behavior)
```

### Renderer interface

```go
type Renderer interface {
    StartPhase(name string)
    EndPhase(status PhaseStatus, duration time.Duration)
    StartStep(name string)
    EndStep(status StepStatus, duration time.Duration, detail string)
    Detail(line string)         // verbose sub-output (hetzner-k3s, OpenTofu lines)
    Summary(items []SummaryItem)
    Error(err error, hint string)
    Confirm(message string, expected string) bool
}
```

### Dependencies

- [`pterm`](https://github.com/pterm/pterm) — Go terminal formatting (MIT, no CGo, 5k+ stars)
- [`mattn/go-isatty`](https://github.com/mattn/go-isatty) — TTY detection

### Migration path

Each step is independently shippable — the `json` renderer preserves current behavior throughout:

1. Add `Renderer` interface + `json.go` (wraps current slog) + `pretty.go` + `plain.go`
2. Wire `--output` flag + isatty check in `main.go`
3. Migrate `status_handler.go` to dispatch to renderer
4. Migrate `print*Instructions` → `renderer.Summary()`
5. Redirect third-party stdout to `renderer.Detail()`

viniciusdc:
## Summary

Closes #244

Replaces NIC's raw JSON slog output with a renderer abstraction that supports three output modes: **pretty** (colored Unicode tree for TTYs), **json** (backward-compatible structured logs for pipes/CI), and **plain** (ASCII-only, no color for dumb terminals). Auto-detects TTY and respects `NO_COLOR` env var.

### `cmd/nic/renderer/` (new package)
- `renderer.go` — `Renderer` interface with phase/step lifecycle, summary, error, confirm, version methods
- `json.go` — JSON renderer wrapping slog (preserves current behavior exactly)
- `pretty.go` — Colored Unicode tree output with right-aligned timing, `NewPretty()` and `NewPlain()` constructors
- `line_writer.go` — goroutine-safe `io.Writer` adapter that splits bytes on newlines and routes to `Detail()`
- `context.go` — Context helpers for threading renderer through cobra commands
- `renderer_test.go` — Unit tests for all three renderers, line_writer, and visibleLen

### `cmd/nic/main.go`
- Added `--format` flag (auto/pretty/json/plain) and `--verbose` flag
- `PersistentPreRunE` selects renderer based on flag + isatty detection + `NO_COLOR`
- Uses `--format` instead of `--output` to avoid collision with kubeconfig's `-o` flag

### `cmd/nic/status_handler.go`
- Replaced `statusLogHandler()` (slog bridge) with `statusRendererHandler(r)` that routes status bus events to the renderer

### `cmd/nic/deploy.go`
- Replaced ~86 `fmt.Print*` calls (`printArgoCDInstructions`, `printKeycloakInstructions`, `printDNSGuidance`) with `renderer.Summary()`
- Replaced ~35 `slog.*` calls with renderer phase/step calls
- Added phase grouping: Configuration, Infrastructure, GitOps, ArgoCD, Foundational Services, DNS

### `cmd/nic/destroy.go`
- Replaced `confirmDestruction` with `renderer.Confirm()`
- Replaced slog calls with renderer calls, added phase grouping

### `cmd/nic/validate.go` and `cmd/nic/version.go`
- Replaced `fmt.Printf` calls with renderer methods

### `pkg/provider/provider.go`
- Added `Output io.Writer` field to `DeployOptions` and `DestroyOptions` (backward compatible — nil defaults to os.Stdout)

### `pkg/tofu/tofu.go`
- Added `SetupOption` / `WithOutput()` functional options to `Setup()` for redirecting OpenTofu stdout/stderr (backward compatible — variadic, existing callers unchanged)

## Before / After

```bash
# Before (JSON noise on every deploy)
{"level":"INFO","msg":"Starting deployment","config_file":"config.yaml"}
{"level":"INFO","msg":"Status","message":"Creating cluster","resource":"provider"}
{"level":"INFO","msg":"Success","message":"Argo CD is ready","resource":"argocd"}

# After (pretty mode — default on TTY)
  Deploying nic-nebari (hetzner)

  ✓ Configuration validated                                     0.2s

  ● Infrastructure
    ├─ ✓ Cluster created                                       3m 38s

  ● ArgoCD
    ├─ ✓ Helm chart installed                                  12.1s

  ✓ Deployment complete                                        5m 29s

# After (json mode — default on pipe, or --format json)
# Identical structured JSON output as before
```


viniciusdc:
## Live Output Examples

Tested locally against the `examples/local-config.yaml` with the local provider.

### `nic version --format pretty`
nic-version-styled

### `nic version --format plain`
```
Nebari Infrastructure Core (NIC)
Version:    1.0.0
Commit:     dev
OpenTofu:   1.11.3
Providers:  aws, gcp, azure, local, hetzner
DNS:        cloudflare
```

### `nic version --format json`
```json
{"time":"...","level":"INFO","msg":"Version","version":"1.0.0","commit":"dev","opentofu_version":"1.11.3","cluster_providers":["aws","gcp","azure","local","hetzner"],"dns_providers":["cloudflare"]}
```

### `nic validate -f examples/local-config.yaml --format pretty`
nic-validate-styled

### `nic validate -f examples/local-config.yaml --format plain`
nic-validate-plain-styled

### `nic deploy -f local-test.yaml --format pretty` (local provider, no kind cluster)

Shows phase grouping, tree structure, warnings, and error formatting:
nic-deploy-styled

### Key behaviors

- `--format auto` (default): detects TTY → pretty, pipe → json. Respects `NO_COLOR` env → plain
- `--format` uses a different flag name than kubeconfig's `--output`/`-o` to avoid cobra collision
- `--verbose` / `-v` shows third-party tool detail lines (hetzner-k3s, OpenTofu output) indented under the current phase
- JSON mode preserves the exact structured slog output format for backward compatibility

pkg/provider/provider.go
```diff
@@ -11,13 +12,19 @@ import (
 type DeployOptions struct {
 	DryRun  bool
 	Timeout time.Duration
+	// Output is an optional writer for third-party tool output (OpenTofu, hetzner-k3s).
+	// If nil, defaults to os.Stdout.
+	Output io.Writer
```
dcmcand, line 17: **Blocker - dead surface on the provider interface.** This field is added but never set by any caller (`cmd/nic/deploy.go:131` constructs `DeployOptions{DryRun, Timeout}` without `Output`) and no provider reads it. It leaks presentation plumbing into the domain contract with no consumer. Either wire `renderer.DetailWriter()` through the CLI and into each provider's tofu invocation in this PR, or drop the field and add it when the consumer lands.

pkg/provider/provider.go
```diff
@@ -11,13 +12,19 @@ import (
...
 }
 
 // DestroyOptions holds runtime flags for infrastructure destruction.
 type DestroyOptions struct {
 	DryRun  bool
 	Force   bool
 	Timeout time.Duration
+	// Output is an optional writer for third-party tool output.
+	// If nil, defaults to os.Stdout.
+	Output io.Writer
```
dcmcand, line 27: **Blocker - dead surface (same as `DeployOptions.Output`).** Not set by the `destroy.go:146` call site, not read by any provider. Remove or wire it.

pkg/tofu/tofu.go
```diff
@@ -223,7 +224,23 @@ func extractTemplates(appFs afero.Fs, templates fs.FS) (string, error) {
...
+// SetupOption configures optional behavior for Setup.
+type SetupOption func(*setupConfig)
+
+type setupConfig struct {
+	stdout io.Writer
+	stderr io.Writer
+}
+
+// WithOutput redirects OpenTofu stdout and stderr to the given writers.
+func WithOutput(stdout, stderr io.Writer) SetupOption {
```
dcmcand, line 236: **Blocker - dead code.** Grep for `tofu.WithOutput` returns zero call sites; the three `provider/aws` `tofu.Setup` invocations all use the no-option form. This functional-option seam is added to `pkg/tofu` with no consumer. Pair it with the CLI wiring that routes `renderer.DetailWriter()` into `tofu.Setup(...)`, or remove until the consumer exists.

cmd/nic/renderer/pretty.go
```diff
@@ -0,0 +1,403 @@
...
+	verbose bool
+	sym     symbols
+	color   bool
+
+	// state
+	phase      string
+	inDetail   bool         // currently inside a detail block (between dotted lines)
+	pending    *pendingStep // step waiting to know if it's the last in its phase
+	afterPhase bool         // true after EndPhase, used to print separator before out-of-phase steps
+}
```
dcmcand, line 82: **Blocker - not goroutine-safe.** The `Pretty` struct carries mutable state (`phase`, `inDetail`, `pending`, `afterPhase`) with no mutex, but in `deploy.go`/`destroy.go` two goroutines hit it: the main goroutine calls `r.EndStep/EndPhase/Error/Warn/Info` directly, and the status-handler goroutine (`pkg/status/status.go:217-222`) dispatches provider `status.Send*` calls into the same instance. Providers emit ~109 of these across `pkg/`; queued updates being drained after `provider.Deploy` returns race with the main goroutine's subsequent renderer calls.

`go test -race` doesn't catch this because no test exercises the real renderer under concurrent producers. Fix: add a `sync.Mutex` guarding every method entry, or serialize all calls through the statusRendererHandler boundary.

cmd/nic/renderer/tui.go
```diff
@@ -0,0 +1,826 @@
...
+	return t
+}
+
+func (t *TUI) send(msg tea.Msg) {
+	t.mu.Lock()
+	done := t.done
+	t.mu.Unlock()
+	if !done {
+		t.program.Send(msg)
+	}
```
dcmcand, line 684: **Blocker - TOCTOU race.** `t.done` is read under the mutex, then `t.program.Send(msg)` is called outside the lock. Combined with `restart()` below (which mutates `t.program` outside the lock), a status-handler goroutine calling `send()` during a `Confirm()`-triggered restart can read a racy pointer. Suggested: guard reads/writes of `t.program` under the mutex, or use `atomic.Pointer[tea.Program]`.

cmd/nic/renderer/tui.go
```diff
@@ -0,0 +1,826 @@
...
+	t.mu.Lock()
+	t.done = false
+	t.mu.Unlock()
+
+	model := newTuiModel()
+	p := tea.NewProgram(model,
+		tea.WithAltScreen(),
+		tea.WithMouseCellMotion(),
+	)
+	t.program = p
```
dcmcand, line 812: **Blocker - unsynchronized write to `t.program`.** Only `t.done = false` is guarded; `t.program = p` and `t.ready = make(...)` happen outside the mutex. Paired with the read in `send()` above, this is a data race.

cmd/nic/main.go
```diff
@@ -25,23 +29,73 @@ var (
...
+				return err
+			}
+			cmd.SetContext(renderer.WithRenderer(cmd.Context(), r))
+
+			// For JSON mode, use the JSON renderer's slog logger as default.
+			// For pretty/plain mode, discard slog output to avoid JSON noise.
+			if jr, ok := r.(*renderer.JSON); ok {
+				slog.SetDefault(jr.Logger())
+			} else {
+				slog.SetDefault(slog.New(slog.NewJSONHandler(io.Discard, nil)))
```
dcmcand, line 54: **Blocker - silently swallows errors in non-JSON modes.** Sending the default slog sink to `io.Discard` means:
- `main.go:161, 168, 179` (telemetry setup / shutdown / command execution failure)
- `main.go:154` (Ctrl+C signal warning)
- Several returned-error paths in `deploy.go`/`destroy.go` that don't call `r.Error` first (e.g. `deploy.go:115` invalid timeout, `deploy.go:128` provider lookup, `destroy.go:86` invalid timeout)

...all produce exit 1 with no user-visible output in pretty/plain/tui modes. Suggestion: route top-level error output through `fmt.Fprintln(os.Stderr, err)` or the renderer's Error method, not through a silenced slog sink.

cmd/nic/deploy.go
```diff
@@ -86,96 +79,102 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
-	// Parse custom timeout if specified
+	// Parse custom timeout
 	var timeout time.Duration
 	if deployTimeout != "" {
-		var err error
 		timeout, err = time.ParseDuration(deployTimeout)
 		if err != nil {
 			span.RecordError(err)
-			slog.Error("Invalid timeout duration", "error", err, "timeout", deployTimeout)
 			return fmt.Errorf("invalid timeout duration %q: %w", deployTimeout, err)
```
dcmcand, line 115: **Blocker (part of the silenced-errors issue).** This returns a wrapped error without calling `r.Error` first. Combined with `main.go:54` redirecting slog to `io.Discard`, in pretty mode a bad `--timeout=foo` value produces exit 1 with no error output. Same at `destroy.go:86`. Either call `r.Error(err, ...)` before returning, or make the top-level error handler in `main.go` print to stderr unconditionally.

cmd/nic/deploy.go
```diff
@@ -86,96 +79,102 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
-			slog.Warn("You can install Argo CD manually with: helm install argocd argo/argo-cd --namespace argocd --create-namespace")
+			r.Warn(fmt.Sprintf("Failed to install Argo CD: %v", err))
+			r.Warn("You can install Argo CD manually with: helm install argocd argo/argo-cd --namespace argocd --create-namespace")
+			r.EndPhase(renderer.PhaseFailed, time.Since(argoStart))
 		} else {
-			slog.Info("Argo CD installed successfully")
+			r.EndStep(renderer.StepOK, time.Since(argoStart), "Helm chart installed")
 			argoCDInstalled = true
+			if cfg.Domain != "" {
+				summaryItems = append(summaryItems, renderer.SummaryItem{Label: "ArgoCD", Value: fmt.Sprintf("https://argocd.%s", cfg.Domain)})
```
dcmcand, line 172: **Blocker - content regression, not just a rendering change.** The old `printArgoCDInstructions` emitted a port-forward command (`kubectl port-forward svc/argocd-server -n argocd 8080:443`), the admin-password retrieval recipe (`kubectl -n argocd get secret argocd-initial-admin-secret ...`), and login credentials. The replacement here is a single URL row, and only when `cfg.Domain != ""`. Users deploying locally or without a domain now get zero post-deploy guidance about how to reach ArgoCD.

The PR body frames this as a rendering change ("replace ... with `renderer.Summary()` calls"), but it's also a content-removal change. Re-add the multi-line guidance via additional `Summary` items / `Info` lines, or split this out of the rendering PR.

cmd/nic/deploy.go
```diff
@@ -186,52 +185,59 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
-				slog.Warn("You can install foundational services manually with: kubectl apply -f pkg/foundational/")
+				r.Warn(fmt.Sprintf("Failed to install foundational services: %v", err))
+				r.EndPhase(renderer.PhaseFailed, time.Since(foundStart))
 			} else {
-				slog.Info("Foundational services installed successfully")
+				r.EndStep(renderer.StepOK, time.Since(foundStart), "Foundational services installed")
+				r.EndPhase(renderer.PhaseOK, time.Since(foundStart))
 				keycloakInstalled = true
+				if cfg.Domain != "" {
+					summaryItems = append(summaryItems, renderer.SummaryItem{Label: "Keycloak", Value: fmt.Sprintf("https://keycloak.%s", cfg.Domain)})
```
dcmcand, line 207: **Blocker - same content regression as ArgoCD.** Deleted Keycloak guidance covered pod-status check, port-forward, and admin credential retrieval. Current Summary only emits a URL, and only with a domain configured.

cmd/nic/renderer/json.go
```diff
@@ -0,0 +1,119 @@
...
+		"commit", commitHash,
+		"opentofu_version", tofuVer,
+		"cluster_providers", clusterProviders,
+		"dns_providers", dnsProviders,
+	)
+}
+
+func (j *JSON) Confirm(message string, details map[string]string, expected string) (bool, error) {
+	// In JSON mode we still need to interact with the user via stdin/stdout
+	fmt.Println(message)
```
dcmcand, line 98: `Confirm` writes plain `fmt.Println` output in JSON mode, which breaks pure-JSON-stream consumers. Worth at least documenting in a comment; a stricter fix would emit a structured JSON prompt event and expect tooling to handle it.

cmd/nic/deploy.go
```diff
@@ -186,52 +185,59 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
+				}
+			}
+
+			if !argoCDInstalled {
```
dcmcand, line 211: Stylistic - `argoCDInstalled` is set to `true` at line 170 on this branch with no reset path, so `!argoCDInstalled` here is unreachable. The whole `if/else` can collapse to a single `r.EndPhase(PhaseOK, ...)`.

cmd/nic/deploy.go
```diff
@@ -186,52 +185,59 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
-	// Print instructions after status handler is cleaned up
-	if argoCDInstalled {
-		printArgoCDInstructions(cfg)
-	}
-	if keycloakInstalled {
-		printKeycloakInstructions(cfg)
-	}
+	// --- Final summary ---
+	_ = lbEndpoint
+	_ = keycloakInstalled
```
dcmcand, line 235: Stylistic - `_ = lbEndpoint` and `_ = keycloakInstalled` are orphan locals from the deleted `print*Instructions` call sites. Along with the unused `keycloakInstalled` bookkeeping itself, they can be removed.

cmd/nic/deploy.go
```diff
@@ -86,96 +79,102 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
+			r.EndStep(renderer.StepOK, time.Since(argoStart), "Helm chart installed")
 			argoCDInstalled = true
+			if cfg.Domain != "" {
+				summaryItems = append(summaryItems, renderer.SummaryItem{Label: "ArgoCD", Value: fmt.Sprintf("https://argocd.%s", cfg.Domain)})
+			}
 
-			// Install foundational services via Argo CD
-			slog.Info("Installing foundational services")
+			// --- Foundational Services phase ---
+			r.StartPhase("Foundational Services")
```
dcmcand, line 176: Stylistic - `StartPhase("Foundational Services")` opens a new phase while `"ArgoCD"` is still open (no intervening `EndPhase`). The Pretty renderer doesn't track nesting, so the outer `EndPhase("ArgoCD")` at 212-214 is effectively a no-op (`p.phase` is already `""`). Either end the ArgoCD phase cleanly before starting Foundational Services, or collapse the two into a single phase.

cmd/nic/renderer/renderer.go
```diff
@@ -0,0 +1,71 @@
...
+// Three implementations exist: JSON (machine-readable), Pretty (colored Unicode
+// for TTYs), and Plain (ASCII-only, no color for dumb terminals).
+type Renderer interface {
+	// StartPhase begins a named deployment phase (e.g., "Infrastructure", "ArgoCD").
+	StartPhase(name string)
+	// EndPhase closes the current phase with a status and elapsed duration.
+	EndPhase(status PhaseStatus, elapsed time.Duration)
+
+	// StartStep begins a named step within the current phase.
+	StartStep(name string)
```
dcmcand, line 42: Question - `StartStep` is a no-op in every implementation (Pretty: `pretty.go:213-215`, TUI: `tui.go:695-697`, JSON: emits but isn't paired with EndStep). And in `statusRendererHandler` (`status_handler.go:20`), `LevelProgress` is mapped to `StartStep`, which means progress updates during long operations are silently dropped in pretty/tui mode. Either drop `StartStep` from the interface and remap `LevelProgress` to `Info`/`Detail`, or actually render it.

cmd/nic/version.go
```diff
@@ -24,22 +20,11 @@ var versionCmd = &cobra.Command{
 }
 
 func runVersion(cmd *cobra.Command, args []string) error {
```
dcmcand, line 22: OTEL instrumentation regression - `runVersion` previously created a `cmd.version` span. CLAUDE.md requires all functions to be OTEL-instrumented. Same pattern holds at `destroy.go:106-114` where the deleted `confirmDestruction` had a `cmd.confirmDestruction` span. Worth adding spans back for the modified paths.

cmd/nic/renderer/line_writer.go
```diff
@@ -0,0 +1,44 @@
...
+			break
+		}
+		line := string(w.buf[:idx])
+		w.buf = w.buf[idx+1:]
+		if line != "" {
+			w.r.Detail(line)
+		}
+	}
+	return len(p), nil
+}
```
dcmcand, line 33: Nice - goroutine-safe line-buffering adapter is exactly what's needed to route third-party stdout/stderr through `Detail()` without interleaving partial lines. Clean.

cmd/nic/main.go
```diff
@@ -25,23 +29,73 @@ var (
...
+	case "pretty":
+		return renderer.NewPretty(os.Stdout, verbose), nil
+	case "plain":
+		return renderer.NewPlain(os.Stdout, verbose), nil
+	case "auto", "":
+		if isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()) {
+			if os.Getenv("NO_COLOR") != "" {
+				return renderer.NewPlain(os.Stdout, verbose), nil
+			}
+			return renderer.NewPretty(os.Stdout, verbose), nil
```
dcmcand, line 86: Clean auto-detection: TTY check first, `NO_COLOR` opt-out to ASCII, pipe-to-JSON default. This is the idiomatic pattern.

cmd/nic/renderer/pretty.go
```diff
@@ -0,0 +1,403 @@
...
+			continue
+		}
+		if r == '\033' {
+			inEsc = true
+			continue
+		}
+		n++
+	}
+	return n
+}
```
dcmcand, line 403: Nicely done - correctly strips ANSI escape sequences so `padRight` gets the actual visible width. Easy to get wrong.

cmd/nic/renderer/renderer.go
```diff
@@ -0,0 +1,71 @@
...
+	// Version prints version information.
+	Version(version, commit, tofuVersion string, clusterProviders, dnsProviders []string)
+
+	// Confirm prompts the user to type the expected string to confirm a destructive action.
+	Confirm(message string, details map[string]string, expected string) (bool, error)
+
+	// DetailWriter returns an io.Writer that splits incoming bytes on newlines
+	// and routes each line to Detail(). Used to capture third-party tool stdout/stderr.
+	DetailWriter() io.Writer
+}
```
dcmcand, line 71: Clean three-implementation interface - JSON for backward compat with log aggregators, Pretty for TTYs, Plain for dumb terminals, TUI as an experimental bonus. Good abstraction.

cmd/nic/renderer/renderer_test.go
```diff
@@ -0,0 +1,635 @@
...
+		{"keycloak url", "https://keycloak.example.com"},
+		{"kubeconfig path", "kubeconfig"},
+	}
+
+	for _, c := range checks {
+		if !strings.Contains(output, c.pattern) {
+			t.Errorf("[%s] expected output to contain %q", c.name, c.pattern)
+		}
+	}
+}
```
dcmcand, line 451: Thorough coverage of the Pretty tree logic: connector selection (├─/└─), detail containment, out-of-phase steps, separator-before-final, full deployment flow. The `TestPretty_FullDeploymentFlow` in particular is a great regression canary.

dcmcand (changes requested):
Thanks for taking this on - the renderer abstraction is a clean model and the Pretty output is a real improvement over the JSON wall-of-text. A few things I think need to be addressed before merge:

**Design-gating**
- `DeployOptions.Output` / `DestroyOptions.Output` and `tofu.WithOutput` are added to `pkg/` but never wired or read. Either remove them or land them with a consumer in the same PR - carving presentation seams into domain types without a caller is dead surface that encourages future misuse.

**Code-level blockers**
- The Pretty renderer has mutable state but no synchronization, and it's being called concurrently from the main goroutine and the status-handler goroutine (`status.StartHandler` spawns a consumer at `pkg/status/status.go:217`). `go test -race` doesn't catch it because no test exercises the real renderer under concurrent input - but in production, any status event draining after `provider.Deploy` returns races with the next direct renderer call.
- The TUI renderer has a similar TOCTOU on `t.program` in `send()` vs `restart()`.
- Redirecting slog to `io.Discard` when the renderer isn't JSON silently swallows the top-level error messages in `main.go` and several returned-error paths in `deploy.go`/`destroy.go`. In pretty mode, an invalid `--timeout` or a telemetry setup failure currently exits 1 with no output.
- Deletion of `printArgoCDInstructions` / `printKeycloakInstructions` / `printDNSGuidance` removes real post-deploy guidance (port-forward commands, password-retrieval recipes, DNS A/CNAME tables). The replacement `Summary()` only emits URL rows, and only when `cfg.Domain != ""` - users deploying locally or without a domain now get zero guidance. This is a content regression, not a rendering change.

Happy to dig into any of these further. See inline comments for specifics.

dcmcand:
A few higher-level observations that didn't fit as inline comments:

**Scope and sequencing.** There are really three PRs interleaved here: (1) the renderer abstraction + Pretty/Plain/JSON modes, (2) the experimental TUI, (3) the `pkg/provider.Output` and `pkg/tofu.WithOutput` seams. The first is the core of #244 and is close to ready. The second is a bigger surface than the design doc called for (826 lines, full Bubble Tea app) and would benefit from shipping behind the hidden-ish `--format tui` flag on its own with its own tests. The third is dead until a later PR wires it up. Splitting those three would make this easier to review and safer to land incrementally - the concurrency concerns on Pretty and TUI don't block the abstraction itself.

**Test coverage of the integration path.** The unit tests for Pretty are excellent for visual/layout behavior, but the concurrency bug I flagged lives at the integration boundary between `pkg/status` (goroutine-based) and the renderer. Worth adding a test that exercises `statusRendererHandler` + a real Pretty/TUI instance with concurrent producers under `-race` - that's the scenario that actually fails in production.

**Design-doc alignment.** The design doc in issue #244 calls out `pterm` as the dependency; the implementation went with `lipgloss` for TUI + hand-rolled ANSI for Pretty. That's fine, but worth updating the doc to match, and noting that plain is a flag on Pretty rather than a separate file.

**Backward-compat assumption.** The `--format=auto` default routing to JSON on pipe is the right call for CI pipelines that parse log output today. Worth a note in the README/docs that anyone who wants pretty output in CI (GitHub Actions summary block, for instance) needs `--format=pretty` explicitly.

Thanks again - the direction here is right, I just want us to split the clearly-scoped rendering work out so the TUI and the unused seams don't gate it.

PR #256, opened 2026-04-15
docs: ADR-0004 out-of-tree provider plugin architecture (Proposed)

dcmcand:
## Summary

Adds ADR-0004 capturing a proposed architecture for out-of-tree provider plugins.

- Multiple plugin kinds (cluster, DNS, cert, git, software), not a single provider interface
- Plugins ship as separate binaries speaking gRPC via `hashicorp/go-plugin`
- Motivated by (a) non-tofu providers like Hetzner, (b) legitimate private integrations like OpenTeams' ASCOT DNS system that cannot live upstream, (c) binary size as provider count grows
- Status: **Proposed**. Open questions (kind boundaries, stages overlap, signing, migration path) are called out explicitly rather than pre-decided.

## Test plan

- [ ] ADR renders correctly on GitHub
- [ ] Index entry in `docs/adr/README.md` links to the new file
- [ ] Reviewers sanity-check the open questions section for anything missing

### DIFF
--- a/docs/adr/0004-out-of-tree-provider-plugins.md
+++ b/docs/adr/0004-out-of-tree-provider-plugins.md
@@ -1,0 +1,160 @@
+# ADR-0004: Out-of-Tree Provider Plugin Architecture
+
+## Status
+
+Proposed
+
+## Date
+
+2026-04-15
+
+## Context
+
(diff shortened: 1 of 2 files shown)


viniciusdc (approved):
We all agreed upon this trough our most recent meeting

PR #259, opened 2026-04-16
feat(aws): install AWS Load Balancer Controller after tf.Apply

### LINKED ISSUE #258: Install AWS Load Balancer Controller via Go Helm SDK after tf.Apply
## Problem

`Service type=LoadBalancer` on EKS clusters provisioned by nic currently falls back to the in-tree Kubernetes AWS cloud-provider controller, which is effectively deprecated and breaks on clusters where both the module-created node SG and the EKS-managed primary cluster SG carry `kubernetes.io/cluster/<name>=owned` (observed on `llmd-test`, 2026-04-15). AWS's recommended path for all new clusters is the AWS Load Balancer Controller (LBC), which uses IP-target mode and sidesteps node-SG auto-discovery entirely.

## Proposed fix

Install the `aws-load-balancer-controller` Helm chart after `tf.Apply`, mirroring the existing pattern for Longhorn in `pkg/provider/aws/longhorn.go`.

### Blocked by

- nebari-dev/terraform-aws-eks-cluster#25 - the TF module provisions the IAM role and EKS Pod Identity association for LBC. This issue can land once that PR merges and nic bumps its pinned module version.

### Implementation

New file `pkg/provider/aws/aws_load_balancer_controller.go` mirroring `longhorn.go`:

```go
const (
    lbcRepoName     = "eks"
    lbcRepoURL      = "https://aws.github.io/eks-charts"
    lbcChartName    = "eks/aws-load-balancer-controller"
    lbcChartVersion = "1.17.1" // or pinned to a known-good v1.x release
    lbcNamespace    = "kube-system"
    lbcReleaseName  = "aws-load-balancer-controller"
    lbcTimeout      = 5 * time.Minute
)

func installAWSLoadBalancerController(ctx context.Context, kubeconfigBytes []byte, cfg *Config, tfOutputs map[string]tfexec.OutputMeta) error {
    // mirror longhorn.go: AddRepo, NewActionConfig, check for existing release,
    // install or upgrade with values: clusterName, region, vpcId
    // wait for Deployment/aws-load-balancer-controller to become Ready
}
```

Wire from `provider.go` after `tf.Apply` and before any consumer of `Service type=LoadBalancer`:

```go
// after tf.Apply, and before EFS StorageClass creation
if awsCfg.LoadBalancerControllerEnabled() {
    kubeconfigBytes, err := p.GetKubeconfig(ctx, projectName, clusterConfig)
    if err != nil {
        return fmt.Errorf("failed to get kubeconfig for LBC install: %w", err)
    }
    if err := installAWSLoadBalancerController(ctx, kubeconfigBytes, awsCfg, outputs); err != nil {
        return fmt.Errorf("failed to install AWS Load Balancer Controller: %w", err)
    }
}
```

Helm values required:
- `clusterName` - from tf output `cluster_name`
- `region` - from `awsCfg.Region`
- `vpcId` - from tf output `vpc_id`
- `serviceAccount.create: true`, `serviceAccount.name: aws-load-balancer-controller` (matches the Pod Identity association created by the TF module)

No SA annotations needed - Pod Identity matches by (cluster, namespace, SA name) tuple, not by SA annotation.

### Config surface

Add to `Config`:

```go
type AWSLoadBalancerControllerConfig struct {
    Enabled      bool   `yaml:\"enabled\"`
    ChartVersion string `yaml:\"chart_version,omitempty\"` // optional override
}
```

Default enabled. Document in `examples/` that consumers wanting a `Service type=LoadBalancer` should annotate with `service.beta.kubernetes.io/aws-load-balancer-type: external` and `...nlb-target-type: ip`.

### Tests

- Unit test for `awsLoadBalancerControllerHelmValues(cfg, outputs)` - same style as `longhornHelmValues`.
- Integration test: deploy an echo Service with the external/ip-target annotations, assert the Service status gets a populated `loadBalancer.ingress[0].hostname`. Also assert `aws-load-balancer-controller` Deployment in `kube-system` is Ready before checking Service status, for better failure diagnostics.
- Teardown: explicitly `kubectl delete svc --wait` before `DeleteNamespace` to avoid NLB ENIs blocking VPC destroy.

### Caveats

- LBC chart v3.x (latest) is a major controller version bump requiring K8s >= 1.28 and TargetGroupBinding CRD v1. v1.17.1 is the last v1.x chart line and is the safer default unless there's a specific reason to take v3.
- LBC depends on the Pod Identity agent addon (`eks-pod-identity-agent`) which the TF module already installs `before_compute = true`.
- LBC install should block on node readiness; nic's `tf.Apply` already waits for node groups to be `ACTIVE` before returning, so this should be safe in practice.

### References

- AWS LBC install doc: https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html
- Helm chart values: https://github.com/aws/eks-charts/blob/master/stable/aws-load-balancer-controller/values.yaml
- Existing Longhorn pattern in nic: `pkg/provider/aws/longhorn.go`

dcmcand:
## Summary

- Installs the `aws-load-balancer-controller` Helm chart (v3.2.1) after `tf.Apply` on EKS deploys so `Service type=LoadBalancer` uses IP-target mode via LBC rather than the deprecated in-tree AWS cloud-provider controller.
- Auth via EKS Pod Identity: the chart creates the `aws-load-balancer-controller` service account in `kube-system`; the Pod Identity association provisioned by the terraform-aws-eks-cluster module matches by (cluster, namespace, SA name) with no IRSA annotation needed.
- Config surface: `aws_load_balancer_controller.{enabled, chart_version}` - default enabled, default chart `3.2.1`.

Closes #258 (once merged).

## Key decisions

- **Chart v3.2.1, not v1.17.1 as originally proposed in #258.** v3 requires the same Kubernetes version as v2 (1.22+), and the v2/chart-v1 line is effectively EOL: last v2 release was v2.17.1 on 2026-01-09, and all upstream work since then is on v3. Pinning to v1.17.1 would opt into an unmaintained line. Chart version is overridable via config.
- **Installer takes `clusterName, vpcID` params instead of `map[string]tfexec.OutputMeta`** so the function does not depend on the tfexec package. The caller extracts from outputs (same pattern as EFS).
- **No destroy-time uninstall hook.** The existing `cleanupKubernetesLoadBalancers` in `Destroy` already cleans up AWS-side LBs tagged with the cluster name before `tf.Destroy`.

## Blockers / follow-ups

- Module source is pinned to the branch `feature/aws-lb-controller` of `nebari-dev/terraform-aws-eks-cluster` so this PR can be tested end-to-end. **Before merging to main:** pin to a SHA (or a tagged release) once nebari-dev/terraform-aws-eks-cluster#25 merges.
- Integration test (deploy echo Service, assert `loadBalancer.ingress[0].hostname` populates) deferred - will validate manually on a real EKS cluster first, and add proper e2e coverage in a follow-up.

## Test plan

- [x] `go test ./...` (all pass)
- [x] `golangci-lint run` (0 issues)
- [x] `tofu validate` (module fetches successfully from feature branch)
- [ ] Manual deploy to test cluster, verify `aws-load-balancer-controller` Deployment is Ready in `kube-system`
- [ ] Manually apply a test Service with `aws-load-balancer-type: external` + `nlb-target-type: ip` annotations; confirm `.status.loadBalancer.ingress[0].hostname` populates
- [ ] `nic destroy` completes cleanly (NLB ENIs do not block VPC teardown)

### DIFF
--- a/pkg/argocd/templates/apps/cert-manager.yaml
+++ b/pkg/argocd/templates/apps/cert-manager.yaml
@@ -23,6 +23,13 @@ spec:
         installCRDs: true
         config:
           enableGatewayAPI: true
+        # Serialize ACME HTTP-01 challenges to avoid a race when two Certificates
+        # share a SAN (e.g. nebari-gateway-cert and nebari-landing-nebari-system-cert
+        # both list <baseDomain>). Concurrent challenges for the same hostname
+        # collide on Let's Encrypt's authorization record and one fails with
+        # "No such authorization", leaving the cert stuck on failedIssuanceAttempts.
+        # See nebari-dev/nebari-infrastructure-core#267.
+        maxConcurrentChallenges: 1
         prometheus:
           enabled: true
         resources:
--- a/pkg/provider/aws/aws_load_balancer_controller.go
+++ b/pkg/provider/aws/aws_load_balancer_controller.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -11,6 +12,7 @@ import (
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chart/loader"
 	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/storage/driver"
 
 	"github.com/nebari-dev/nebari-infrastructure-core/pkg/helm"
 	"github.com/nebari-dev/nebari-infrastructure-core/pkg/status"
@@ -104,11 +106,17 @@ func installAWSLoadBalancerController(ctx context.Context, kubeconfigBytes []byt
 
 	histClient := action.NewHistory(actionConfig)
 	histClient.Max = 1
-	if _, err := histClient.Run(lbcReleaseName); err == nil {
+	switch _, err := histClient.Run(lbcReleaseName); {
+	case err == nil:
 		status.Send(ctx, status.NewUpdate(status.LevelInfo, "AWS Load Balancer Controller already installed, upgrading").
 			WithResource("aws-load-balancer-controller").
 			WithAction("upgrading"))
 		return upgradeAWSLoadBalancerController(ctx, actionConfig, cfg, clusterName, vpcID)
+	case errors.Is(err, driver.ErrReleaseNotFound):
+		// No existing release; fall through to fresh install.
+	default:
+		span.RecordError(err)
+		return fmt.Errorf("failed to query Helm release history for %s: %w", lbcReleaseName, err)
 	}
 
 	helmValues := awsLoadBalancerControllerHelmValues(cfg, clusterName, vpcID)
@@ -154,7 +162,7 @@ func installAWSLoadBalancerController(ctx context.Context, kubeconfigBytes []byt
 // upgradeAWSLoadBalancerController upgrades an existing release.
 func upgradeAWSLoadBalancerController(ctx context.Context, actionConfig *action.Configuration, cfg *Config, clusterName, vpcID string) error {
 	tracer := otel.Tracer("nebari-infrastructure-core")
-	_, span := tracer.Start(ctx, "aws.upgradeAWSLoadBalancerController")
+	ctx, span := tracer.Start(ctx, "aws.upgradeAWSLoadBalancerController")
 	defer span.End()
 
 	chartVersion := cfg.LoadBalancerControllerChartVersion()
--- a/pkg/provider/aws/aws_load_balancer_controller.go
+++ b/pkg/provider/aws/aws_load_balancer_controller.go
@@ -36,7 +36,7 @@ const (
 func awsLoadBalancerControllerHelmValues(cfg *Config, clusterName, vpcID string) map[string]any {
 	return map[string]any{
 		"clusterName": clusterName,
-		"region":      cfg.Region,
+		attrKeyRegion: cfg.Region,
 		"vpcId":       vpcID,
 		"serviceAccount": map[string]any{
 			"create": true,
--- a/pkg/provider/aws/aws_load_balancer_controller.go
+++ b/pkg/provider/aws/aws_load_balancer_controller.go
@@ -1,0 +1,192 @@
+package aws
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/cli"
+
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/helm"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/status"
+)
+
+const (
+	lbcRepoName       = "eks"
+	lbcRepoURL        = "https://aws.github.io/eks-charts"
+	lbcChartName      = "eks/aws-load-balancer-controller"
+	lbcNamespace      = "kube-system"
+	lbcReleaseName    = "aws-load-balancer-controller"
+	lbcServiceAccount = "aws-load-balancer-controller"
+	lbcInstallTimeout = 5 * time.Minute
+)
+
+// awsLoadBalancerControllerHelmValues builds the Helm values for the
+// aws-load-balancer-controller chart. The service account is created by the
+// chart; authentication is via the EKS Pod Identity association provisioned
+// by the terraform-aws-eks-cluster module, which matches by (cluster,
+// namespace, service account name) and requires no IRSA annotation.
+func awsLoadBalancerControllerHelmValues(cfg *Config, clusterName, vpcID string) map[string]any {
+	return map[string]any{
+		"clusterName": clusterName,
+		"region":      cfg.Region,
+		"vpcId":       vpcID,
+		"serviceAccount": map[string]any{
+			"create": true,
+			"name":   lbcServiceAccount,
+		},
+	}
+}
+
+// loadLBCChart locates and loads the AWS Load Balancer Controller Helm chart.
+func loadLBCChart(chartPathOptions action.ChartPathOptions) (*chart.Chart, error) {
+	chartPath, err := chartPathOptions.LocateChart(lbcChartName, cli.New())
+	if err != nil {
+		return nil, fmt.Errorf("failed to locate AWS Load Balancer Controller chart: %w", err)
+	}
+
+	loadedChart, err := loader.Load(chartPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load AWS Load Balancer Controller chart: %w", err)
+	}
+
+	return loadedChart, nil
+}
+
+// installAWSLoadBalancerController installs or upgrades the AWS Load Balancer
+// Controller on the cluster via Helm. Safe to call on every deploy; it will
+// upgrade an existing release rather than fail.
+func installAWSLoadBalancerController(ctx context.Context, kubeconfigBytes []byte, cfg *Config, clusterName, vpcID string) error {
+	tracer := otel.Tracer("nebari-infrastructure-core")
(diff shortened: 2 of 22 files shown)


docs/adr/0004-out-of-tree-provider-plugins.md
marcelovilla, line 1: I don't think this belongs here

docs/adr/README.md
```diff
@@ -12,6 +12,7 @@ An ADR is a document that captures an important architectural decision made alon
 |----|-------|--------|------|
 | [ADR-0001](0001-git-provider-for-gitops-bootstrap.md) | Git Provider for GitOps Bootstrap | Proposed | 2025-01-21 |
 | [ADR-0002](0002-longhorn-distributed-block-storage-for-aws.md) | Longhorn Distributed Block Storage for AWS | Proposed | 2026-02-13 |
+| [ADR-0004](0004-out-of-tree-provider-plugins.md) | Out-of-Tree Provider Plugin Architecture | Proposed | 2026-04-15 |
```
marcelovilla, line 15: I don't think this belongs here

pkg/provider/aws/templates/main.tf
```diff
@@ -1,6 +1,5 @@
 module "eks_cluster" {
-  source  = "nebari-dev/eks-cluster/aws"
-  version = "0.2.0"
+  source = "git::https://github.com/nebari-dev/terraform-aws-eks-cluster.git?ref=feature/fix-node-sg-tag-conflict"
```
marcelovilla: Needs to be updated to 0.4.0. See https://github.com/nebari-dev/terraform-aws-eks-cluster/releases/tag/v0.4.0

pkg/provider/aws/templates/main.tf
```diff
@@ -1,6 +1,5 @@
 module "eks_cluster" {
-  source  = "nebari-dev/eks-cluster/aws"
-  version = "0.2.0"
+  source = "git::https://github.com/nebari-dev/terraform-aws-eks-cluster.git?ref=v0.4.0"
```
marcelovilla: I'd rather keep the `source` + `version` format that pulls from the OpenTofu registry.

pkg/argocd/templates/apps/cert-manager.yaml
```diff
@@ -23,6 +23,13 @@ spec:
         installCRDs: true
         config:
           enableGatewayAPI: true
+        # Serialize ACME HTTP-01 challenges to avoid a race when two Certificates
+        # share a SAN (e.g. nebari-gateway-cert and nebari-landing-nebari-system-cert
+        # both list <baseDomain>). Concurrent challenges for the same hostname
+        # collide on Let's Encrypt's authorization record and one fails with
+        # "No such authorization", leaving the cert stuck on failedIssuanceAttempts.
+        # See nebari-dev/nebari-infrastructure-core#267.
+        maxConcurrentChallenges: 1
```
marcelovilla, line 32: This feels out of scope for this PR

pkg/provider/aws/provider.go
```diff
@@ -660,5 +719,10 @@ func (p *Provider) InfraSettings(clusterConfig *config.ClusterConfig) provider.I
 		StorageClass:    sc,
 		NeedsMetalLB:    false,
 		EFSStorageClass: efsSC,
+		LoadBalancerAnnotations: map[string]string{
+			"service.beta.kubernetes.io/aws-load-balancer-type":            "external",
+			"service.beta.kubernetes.io/aws-load-balancer-nlb-target-type": "ip",
+			"service.beta.kubernetes.io/aws-load-balancer-scheme":          "internet-facing",
```
marcelovilla, line 725: I'm wondering if these values should be hardcoded here. The [docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/service/annotations/) suggest the second and third annotations are specific to the LB controller and I'm not sure if we want to have `external` for the `aws-load-balancer-type` annotation when the LB controller is disabled.

pkg/provider/aws/config.go
```diff
@@ -1,25 +1,71 @@
...
+	Tags                      map[string]string                `yaml:"tags,omitempty"`
+	EFS                       *EFSConfig                       `yaml:"efs,omitempty"`
+	Longhorn                  *LonghornConfig                  `yaml:"longhorn,omitempty"`
+	AWSLoadBalancerController *AWSLoadBalancerControllerConfig `yaml:"aws_load_balancer_controller,omitempty"`
+}
+
+type AWSLoadBalancerControllerConfig struct {
+	Enabled        *bool          `yaml:"enabled,omitempty"`
+	ChartVersion   string         `yaml:"chart_version,omitempty"`
+	DestroyTimeout *time.Duration `yaml:"destroy_timeout,omitempty"`
```
marcelovilla, line 31: Should we add this config option to the example? Until we have better docs, it wouldn't be easy for people to find this is something they can configure.

pkg/provider/aws/templates/main.tf
marcelovilla, line 1: We're not passing `enable_aws_load_balancer_controller_pod_identity` as a variable to the upstream TF module, which means that users who explicitly disabled the LB controller, will still have an unnecessary IAM role + pod identity be created.

pkg/provider/aws/aws_load_balancer_controller.go
```diff
@@ -0,0 +1,200 @@
...
+	defer span.End()
+
+	chartVersion := cfg.LoadBalancerControllerChartVersion()
+	helmValues := awsLoadBalancerControllerHelmValues(cfg, clusterName, vpcID)
+
+	client := action.NewUpgrade(actionConfig)
+	client.Namespace = lbcNamespace
+	client.Wait = true
+	client.Timeout = lbcInstallTimeout
+	client.Version = chartVersion
```
marcelovilla, line 175: Small nit: this should be an idempotent action so there's not a problem if we run it on every `nic deploy`. However, it seems that when `helm upgrade` runs, it [creates a new revision](https://helm.sh/docs/intro/using_helm/#helm-upgrade-and-helm-rollback-upgrading-a-release-and-recovering-on-failure) everytime. In [the CLI, this is controlled via the `--history-max` option](https://helm.sh/docs/helm/helm_upgrade/#options):

> limit the maximum number of revisions saved per release. Use 0 for no limit (default 10)

However, looking at the [client constructor](https://github.com/helm/helm/blob/0f09636c796cfbd676c7362e4ada81a0cdb04d76/pkg/action/upgrade.go#L145-L156), it does not specify any default value and it [doesn't seem to be set anywhere](https://github.com/helm/helm/blob/0f09636c796cfbd676c7362e4ada81a0cdb04d76/pkg/action/upgrade.go#L105), which leads me to believe it's 0 by default.

Maybe we should set a default here so we keep things tidy instead of users ending up with hundreds of revisions, one for every time they ran `nic deploy`? For example, to mimic the CLI behavior:

```go
client.MaxHistory = 10
```

marcelovilla (commented):
@dcmcand thanks for this PR and all the effort! It's definitely a good chunk of work and a step in the right direction.

I left some comments before approving this.

marcelovilla:
I just tested deploying this twice, once enabling the LB controller and once disabling it via:

```yaml
cluster:
  aws:
    aws_load_balancer_controller:
      enabled: false
  ...
```

First one deployed fine and was fully functional. Destroyed also succeeded. In the second case, deploy failed with:

```
{"time":"2026-05-11T15:53:02.774319+02:00","level":"INFO","msg":"Waiting for load balancer endpoint..."}
{"time":"2026-05-11T15:58:02.776801+02:00","level":"WARN","msg":"Could not retrieve load balancer endpoint","error":"timed out waiting for load balancer endpoint after 5m0s"}
```

dcmcand:
Follow-up issues opened from the review comments on this PR:

- #290 - bug: `nic deploy` hangs waiting for LB endpoint when `aws_load_balancer_controller.enabled: false`
- #291 - maintenance: cap Helm `MaxHistory` on the LBC upgrade client
- #292 - bug: make LB Service annotations conditional on LBC enabled state
- #293 - test: e2e coverage for AWS Load Balancer Controller install
- #294 - docs: add `aws_load_balancer_controller` block to the AWS example config
- #295 - enhancement: plumb `enable_aws_load_balancer_controller_pod_identity` through to the EKS module

The registry `source` + `version` change is in 6a33da5.

marcelovilla (approved):
Thanks for opening the follow-up issues @dcmcand! Approving

PR #260, opened 2026-04-16
test: demo priority workflow failure without label

tylerpotts:
## Summary
- Trivial Makefile comment to demonstrate that the priority-sync workflow fails when no `priority: *` label is present on a PR.

## Test plan
- [ ] Observe the workflow run fails due to `grep` exit code 1 under `set -eo pipefail`
- [ ] Apply the fix, then verify it passes with no label


tylerpotts:
no longer needed for testing

Issue #72, opened 2026-02-09
Improve provider config validation logic


marcelovilla:
## Description

Our provider config validation is currently hand-rolled if-checks in the `Validate()` method of each Provider. We should evaluate structured validation approaches, such as https://github.com/go-playground/validator. It supports struct tags but also seems to support struct level validators for cross-field logic, which would be certainly beneficial in our case.

## Use Case

Several fields lack validation at the moment and for a better user experience, validation should happen when parsing the config instead of letting OpenTofu or other tools fail because of wrong configuration options were passed.

PR #266, opened 2026-04-24
Extract CLI orchestration into its own package

### LINKED ISSUE #264: Refactor: extract deploy/destroy orchestration into a reusable Go package
## Description

Extract the deploy and destroy orchestration logic from `cmd/nic/deploy.go` and `cmd/nic/destroy.go` into a reusable Go package (e.g. `pkg/orchestrator`) so the CLI becomes a thin wrapper around it and external Go projects can invoke the same flows programmatically without shelling out to the `nic` binary or duplicating its logic.

The concrete API proposal (package layout, request/result types, option handling) is left to a follow-up ADR.

### What moves into the orchestrator

- Config validation (as a safety net — idempotent, cheap).
- Provider lookup and default registry construction.
- Status handler wiring (callers inject the sink).
- `provider.Deploy` invocation.
- GitOps bootstrap (`bootstrapGitOps`).
- Argo CD + foundational services installation.
- Load balancer endpoint lookup and DNS provisioning.
- Password generation (with an injectable `io.Reader` for testability).
- DNS records cleanup (`destroyDNS`).
- `provider.Destroy` invocation, including `--force` handling (continue past provider errors, surface via `PartialFailure`).

### What stays in `cmd/nic`

- Flag parsing, config file auto-discovery, and `config.ParseConfig` call.
- CLI-specific status sink (`statusLogHandler()`).
- Pretty-printed instructions (`printArgoCDInstructions`, `printKeycloakInstructions`, `printDNSGuidance`) — operate on the returned `DeployResult`.
- Interactive destroy confirmation prompt (`confirmDestruction`) — stdin/stdout, not library concern.
- Signal handling / cancellable context.

## Use Case

Other Go projects wanting to use `nebari-infrastructure-core` to deploy a cluster and the foundational software have two options today:

1. Embed the `nic` binary and shell out to it. This approach means they would need to embed the binary, they would lose typed error handling and they still need to pass a YAML config file on disk.
2. Use the different go packages exposed by `nebari-infrastructre-core` and duplicate the orchestration logic from `deploy.go` and `destroy.go`. This approach would drift over time and would increase the maintenance burden for every provider change or new foundational service.

Refactoring so the CLI is a thin wrapper over a reusable package removes this duplication and makes `nebari-infrastructure-core` genuinely consumable as a library. It also improves testability of the deploy/destroy flows themselves, since the orchestrator becomes unit-testable without spinning up the CLI.

marcelovilla:
## Closes

Closes #264

## Description

This PR extracts the deploy/destroy/validate/kubeconfig orchestration logic from `cmd/nic` into a new public `pkg/nic` package. The CLI is now a thin wrapper calling `pkg/nic` and library consumers can interact with nic via constructing a client `nic.NewClient(...)` and then calling the different functions.

This PR tries to be as faithful as possible to what we currently have on the `main` branch with one small caveat: the scrubbed config committed during GitOps bootstrap is now serialised from the parsed `*NebariConfig` via a compile-checked struct scrub, rather than read from a file path. This way, the feature works uniformly for library consumers and CLI users.

## How to Test

To test this PR, one should try interacting with nic in two different ways: using the CLI and consuming the library directly.

#### Using the CLI

Run:
```bash
./nic deploy -f examples/aws-config.yaml --dry-run
```

#### Using the library

On a different folder, create `go.mod` with the following contents:

```go
module example.com/nic-consumer

go 1.25

require github.com/nebari-dev/nebari-infrastructure-core v0.0.0-20260522090721-e467add27b90
```

and `main.go` with the following contents:
```go
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/nebari-dev/nebari-infrastructure-core/pkg/config"
	"github.com/nebari-dev/nebari-infrastructure-core/pkg/nic"
)

func main() {
	ctx := context.Background()

	cfg, err := config.ParseConfig(ctx, "../examples/aws-config.yaml")
	if err != nil {
		fmt.Fprintln(os.Stderr, "parse config:", err)
		os.Exit(1)
	}

	client := nic.NewClient()

	_, err = client.Deploy(ctx, cfg, nic.DeployOptions{
		DryRun:    true,
		RegenApps: false,
	})
	if err != nil {
		fmt.Fprintln(os.Stderr, "deploy:", err)
		os.Exit(1)
	}
}

```

Then run:
```bash
go mod tidy
go run .
```


pkg/nic/client.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// NewClient returns a new NIC client. Pass no arguments for default
+// configuration, or a single *ClientConfig to customise.
+//
+//	client := nic.NewClient()
+//	client := nic.NewClient(&nic.ClientConfig{Logger: myLogger})
+//
+// Only the first ClientConfig is read; additional arguments are ignored.
+func NewClient(cfg ...*ClientConfig) *Client {
```
dcmcand: **Kudos:** variadic optional `ClientConfig` is a nice API shape - zero-arg works for the common case, one-arg customizes, and future fields don't break callers. Worth replicating elsewhere.

pkg/nic/destroy.go
```diff
@@ -0,0 +1,166 @@
...
+
+	prov, err := reg.ClusterProviders.Get(ctx, cfg.Cluster.ProviderName())
+	if err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("get cluster provider: %w", err)
+	}
+
+	c.logger.Info("Provider selected", "provider", prov.Name())
+
+	if opts.Confirm != nil && !opts.DryRun {
```
dcmcand, line 104: **Kudos:** ordering here is right - resolve the provider, build the summary, run `Confirm`, *then* start the status handler. Keeps the y/n prompt from interleaving with progress lines. Worth keeping.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,485 @@
...
+	c.logger.Info("Wrote NIC config to repository (auth fields scrubbed)", "path", configDest)
+	return nil
+}
+
+// scrubbedConfig returns a copy of cfg with sensitive fields zeroed (or
+// nilled out where the schema supports omitempty). Operating on the typed
+// struct means renaming a sensitive field on the source type fails to
+// compile here, instead of silently leaking once a deny-list of string
+// keys drifts out of sync.
+func scrubbedConfig(cfg *config.NebariConfig) *config.NebariConfig {
```
dcmcand: **Kudos:** operating on the typed struct rather than a string deny-list is exactly the right call here, and the comment explains *why* (rename-safety) instead of *what*. The new round-trip YAML test (`marshalled output excludes sensitive strings`) is a great regression guard.

pkg/nic/registry.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return &Providers{
+		Cluster: reg.ClusterProviders.List(ctx),
+		DNS:     reg.DNSProviders.List(ctx),
+	}, nil
+}
+
+// defaultRegistry builds a Registry with all in-tree cluster and DNS
+// providers registered.
+func defaultRegistry(ctx context.Context) (*registry.Registry, error) {
```
dcmcand, line 44: **Question:** `defaultRegistry` is rebuilt and re-registers all six cluster providers and the Cloudflare DNS provider on every `Deploy`/`Destroy`/`Validate`/`Kubeconfig` call. With no caller-facing API to inject custom providers between calls, why not build the registry once in `NewClient` and store it on `Client`?

If the intent is to leave room for per-call provider injection (e.g. a future `DeployOptions.Registry`), a comment saying so would help. As written it reads as accidental rather than deliberate.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,485 @@
...
+	// Bootstrap GitOps (auto-create local directory for providers that support it)
+	gitConfig, err := c.getOrCreateGitConfig(cfg, infraSettings.SupportsLocalGitOps)
+	if err != nil {
+		span.RecordError(err)
+		c.logger.Error("GitOps configuration failed", "error", err)
+		return nil, err
+	}
+	if gitConfig != nil {
+		// Set on cfg so downstream code (Install, InstallFoundationalServices) can use cfg.GitRepository
+		cfg.GitRepository = gitConfig
```
dcmcand: **Question:** `Deploy` mutates the caller's `cfg` here (`cfg.GitRepository = gitConfig`). For a library API this is surprising - calling `client.Deploy(ctx, cfg, ...)` twice with the same `cfg` pointer will see different state on the second call (the auto-created local gitops config sticks).

Was this intentional, or should `getOrCreateGitConfig` return a local variable that gets threaded into `bootstrapGitOps`, `argocd.Install`, and `argocd.InstallFoundationalServices` explicitly? A library consumer would expect input args to stay input.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,485 @@
...
+}
+
+// generateSecurePassword generates a cryptographically secure random password.
+// It accepts an io.Reader to allow for deterministic testing with known bytes.
+func generateSecurePassword(r io.Reader) string {
+	// Generate 32 bytes of random data
+	b := make([]byte, 32)
+	if _, err := r.Read(b); err != nil {
+		// Fallback to timestamp-based generation (not ideal but better than nothing)
+		return fmt.Sprintf("nebari-%d", time.Now().UnixNano())
```
dcmcand: **Question:** if `r.Read` fails the fallback returns `nebari-<UnixNano>` - a predictable timestamp-based string - which then becomes a Keycloak admin password, Postgres password, etc.

`crypto/rand.Read` essentially never fails on a healthy Linux kernel (post-getrandom init), so practical risk is low. But if it does fail, you'd want to abort the deploy rather than silently install a weak admin credential into the cluster. Preexisting code, just relocated - but the move is a natural moment to harden it. Suggest returning an error and propagating up.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,485 @@
...
+		if ctx.Err() == context.Canceled {
+			c.logger.Warn("Deployment interrupted by user")
+		}
+	}()
+
+	// Validate configuration with registered providers
+	if err := cfg.Validate(validateOptions(ctx, reg)); err != nil {
+		span.RecordError(err)
+		c.logger.Error("Configuration validation failed", "error", err)
+		return nil, err
```
dcmcand: **Stylistic:** validation error is returned raw here, while `Destroy` (`destroy.go:90`), `Validate` (`validate.go:28`), and `Kubeconfig` (`kubeconfig.go:28`) all wrap with `fmt.Errorf("configuration validation failed: %w", err)`. Pick one.

pkg/nic/registry.go
```diff
@@ -0,0 +1,78 @@
...
+// without breaking existing callers.
+type Providers struct {
+	Cluster []string
+	DNS     []string
+}
+
+// ProviderNames returns the providers bundled with this build. Intended for
+// diagnostic output (e.g. a `version` command); operational work should go
+// through the Client's methods.
+func (c *Client) ProviderNames(ctx context.Context) (*Providers, error) {
```
dcmcand: **Stylistic:** `ProviderNames` takes `ctx` but doesn't span-instrument or use it. Per CLAUDE.md's instrumentation rule, all `pkg/` functions should start a span, even diagnostic ones. Worth a `tracer.Start(ctx, "nic.ProviderNames")` for symmetry with `Validate`.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,485 @@
...
+		c.logger.Info("Initializing GitOps repository", "url", gitConfig.URL)
+	}
+
+	// Create git client
+	gitClient, err := git.NewClient(gitConfig)
+	if err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("failed to create git client: %w", err)
+	}
+	defer func() { _ = gitClient.Cleanup() }()
```
dcmcand: **Stylistic:** `_ = gitClient.Cleanup()` swallows the error silently. Acceptable for deferred cleanup, but a debug-level log would help when a temp dir gets stuck. Minor.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,485 @@
...
+		result.LBEndpoint = c.lookupEndpointAndProvisionDNS(ctx, cfg, clusterProvider, reg)
+	}
+
+	return result, nil
+}
+
+// defaultGitConfig returns a default local git configuration for development workflows.
+// This is a pure function with no side effects — directory creation happens separately.
+func defaultGitConfig(projectName string) *git.Config {
+	localPath := filepath.Join(os.TempDir(), fmt.Sprintf("nebari-gitops-%s", projectName))
```
dcmcand, line 242: **Stylistic:** `defaultGitConfig` produces `/tmp/nebari-gitops-<projectName>` with `0750` - a predictable path in a world-writable temp dir. On a shared multi-user system this is prediction/TOCTOU-prone. Preexisting; flagging for future hardening rather than blocking this PR.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,485 @@
...
+	// when the GitOps repository is already bootstrapped.
+	RegenApps bool
+}
+
+// DeployResult contains useful information from the deploy process that
+// can be used by callers after Deploy completes.
+type DeployResult struct {
+	// ArgoCDInstalled is true when the Argo CD Helm install completed.
+	// False when skipped (dry-run) or when installation failed.
+	ArgoCDInstalled bool
```
dcmcand, line 47: **Stylistic:** `DeployResult` collapses partial failure into booleans. If ArgoCD install fails, foundational-services is skipped and the caller sees `{ArgoCDInstalled: false, KeycloakInstalled: false}` with no indication of *why*. A `Errors []error` (or a typed `PartialFailure`) would be more useful for programmatic callers. Not blocking - the warning logs cover the CLI case fine.

dcmcand (commented):
Refactor review - see summary comment for the overview. Inline comments below are labelled **Kudos**, **Question**, or **Stylistic**. No blockers; build and full test suite pass.

dcmcand:
## Review summary

This is a clean refactor that extracts CLI orchestration into a new `pkg/nic` package with a `Client` type exposing `Deploy`/`Destroy`/`Validate`/`Kubeconfig`. `cmd/nic` becomes a thin Cobra wrapper that handles flag parsing, stdin prompts, and decorative output. Full test suite passes locally.

### What's working well

- **The library/CLI seam is the right shape.** `DestroyOptions.Confirm` lets callers plug in their own confirmation flow instead of hard-coding stdin prompts, and `DeployResult` returns structured info rather than printing it. This is the kind of API a library consumer can actually use.
- **Architectural boundaries hold up.** `pkg/nic` depends on `pkg/provider`, `pkg/registry`, etc., never on `cmd/nic`. No provider-name switches leaked into the new library layer - `MetalLB`, `KeycloakBasePath`, and `SupportsLocalGitOps` all flow through `provider.InfraSettings` as the project's abstraction rules prescribe.
- **`scrubbedConfig` operating on the typed struct** (with a clear rename-safety rationale) and the new round-trip YAML regression test are a small but real security improvement.
- OTEL spans and slog discipline (logger only in `cmd`, spans in `pkg`) carry through correctly.

### Themes from the inline comments

- **Registry rebuilt per call** - smells unintentional given there's no API for injecting providers. Either cache it on `Client` or add a comment explaining the design intent.
- **`Deploy` mutates the input `cfg`** - surprising for a library API. Worth thinking about whether `getOrCreateGitConfig`'s output should be threaded explicitly rather than written back to the caller's struct.
- **Weak fallback in `generateSecurePassword`** - preexisting, but the move is a natural moment to swap silent degradation for an error return.
- A handful of small consistency nits (error-wrapping pattern across the four methods, `ProviderNames` not span-instrumented, partial-failure surface in `DeployResult`).

### Out of scope but worth tracking

- `defaultGitConfig` writing into a predictable `/tmp/nebari-gitops-<projectName>` path is a TOCTOU concern on multi-user hosts - worth a follow-up issue rather than fixing in this PR.
- A smoke test that constructs a `Client` and calls `ProviderNames` would catch registry-construction regressions cheaply now that `defaultRegistry` is the only path that wires providers up.

Also: the PR description template is still empty - filling in the Description and How-to-Test sections would help future readers.

pkg/nic/registry.go
marcelovilla: Good catch, I wrote this before changing the API to have a client. I did not consider having per-call provider injection and I don't think we need it at this moment so proceeded with your suggestion of having one registry that's built with the client. Addressed in b3953304a2c36ceb38a61454cd76f33fe047764d

pkg/nic/deploy.go
marcelovilla: This was definitely not intentional and I fully agree with you: the function should not mutate the inputs.

I grabbed this from `cmd/nic/deploy.go` and I guess at that point it did not really matter as we were parsing the config from a file:
https://github.com/nebari-dev/nebari-infrastructure-core/blob/0de398ba690b51f175b4b2026194d7cbc71d86fa/cmd/nic/deploy.go#L208-L209

I addressed this in 95515e6e45cbc2003da304816479f75ff49d6440 by adding an explicit `gitConfig` argument in the `argocd` functions to avoid mutating the config. Let me know what you think about this approach.

pkg/nic/deploy.go
marcelovilla: While this is definitely out of the scope of this PR, I agree with you. In order to avoid creating an issue and letting it fall into oblivion, I've addressed this in 76d1b845ceaf4ed5f43f2b756554c3d66dc571ad

pkg/nic/deploy.go
marcelovilla: Addressed in 7a1120283eb99c30328527b0ddaa2432557ac32c

pkg/nic/registry.go
marcelovilla: Addressed in d37f45187ca82f4004088eb84ea0b3e9eda9c809

pkg/nic/deploy.go
marcelovilla: I agree but I'm leaving this out of this PR as it's out of scope and probably needs a little bit of more thinking.

pkg/nic/deploy.go
marcelovilla: Addressed in edb8e4feaa9e02245e4146b91a001d4f2812ebfa

pkg/nic/deploy.go
marcelovilla: Opened https://github.com/nebari-dev/nebari-infrastructure-core/issues/302 as a follow up

marcelovilla:
@dcmcand thanks for your review. I've addressed all of your comments.

Regarding this:

> OTEL spans and slog discipline (logger only in cmd, spans in pkg) carry through correctly.

That's not entirely correct—we're still using the logger in `pkg/nic` but I'm not sure if we would like to rather use the `status` channel at that level.

dcmcand:
You're right to push on this, and after thinking it through I owe you a corrected answer. I said in my summary that slog discipline carries through correctly. It doesn't, and the logger field on `Client` is the symptom.

Background first, then the proposal.

The framing I'd reach for is Bill Kennedy's package-oriented design write-up: https://www.ardanlabs.com/blog/2017/02/package-oriented-design.html. Kennedy roughly splits packages into Kit and Application. Kit packages are foundational and reusable. They never hold or call a logger. They surface progress as structured events and return wrapped errors. Application packages encode this product's orchestration policy and can legitimately own a logger.

By that model, `pkg/nic` is an Application package, so a `*slog.Logger` on `ClientConfig` would be defensible in principle.

But once I work it through, I don't think we should take the Kennedy permission slip. We already have a working progress seam: `pkg/status` plus `cmd/nic/status_handler.go`. Adding a logger field on top means `pkg/nic` ends up with both mechanisms in play, and we have to paper over the seams:

- Default the logger to a discard handler so the library isn't surprising (otherwise it silently captures `slog.Default()`).
- Special-case the "consumer already attached a status handler" path so the bridge doesn't double-emit.
- Maintain two ways to consume progress.

That's a lot of accidental complexity for the ergonomic win of `ClientConfig{Logger: l}` over one extra `status.StartHandler` call.

The cleaner version: drop the logger field, narrate everything in `pkg/nic` with `status.*` emissions the way `pkg/argocd` already does, and leave `cmd/nic/status_handler.go` as the single bridge. Library consumers wire `status.StartHandler(ctx, theirHandler)` before calling `Deploy`. Same API every other consumer of `pkg/...` uses. The moment a consumer wants anything richer than text records (a TUI, metrics, structured routing), the status channel is what they actually want and the logger would have been in the way.

Concretely:

- Remove `Logger` from `ClientConfig` and `Client`.
- Convert the ~47 direct `c.logger.<level>` calls in `deploy.go`, `destroy.go`, and `kubeconfig.go` to `status.*` emissions.
- `cmd/nic` keeps building slog and attaching `status_handler.go` to the ctx before calling client methods, the way it already does for the rest of the orchestration.

Happy to have you open this as a follow-up issue rather than blocking the merge, since it's mechanical but touches a lot of lines. Let me know how you'd like to handle it.

marcelovilla:
@dcmcand thanks for your detailed answer. I agree your suggestion is cleaner and I decided to implement it here while we're at it rather than opening a follow-up issue. See e467add27b907395575897fcbdb5378b017a5e20

Concretely, I removed the logger from the client, converted all logger calls in `pkg/nic` to status emissions, and exposed two functions in `pkg/nic`: `SlogHandler` and `StartSlogHandler`. The latter is meant for those users that want slog output and do not need to interact with the status package directly, such as our CLI. The former is meant for those users that need to wrap the handler and will interact with the status package via `status.StartHandler` after that.

I updated the _How to test_ section in the PR description to reflect these changes if you want to test it again.

cmd/nic/deploy.go
```diff
@@ -117,414 +56,57 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
+		slog.Error("Failed to create NIC client", "error", err)
 		return err
 	}
 
-	slog.Info("Infrastructure deployment completed", "provider", provider.Name())
-
-	// Get provider infrastructure settings for GitOps and foundational services
-	infraSettings := provider.InfraSettings(cfg.Cluster)
+	ctx, cleanup := nic.StartSlogHandler(ctx, slog.Default())
+	defer cleanup()
```
dcmcand, line 85: **Blocker - status cleanup ordering regression.**

The pre-refactor code called `cleanupStatus()` explicitly *before* `printArgoCDInstructions` / `printKeycloakInstructions` / `printDNSGuidance`, with a comment that said it was to drain pending status messages so they wouldn't race into the middle of the printed panels.

Now only `defer cleanup()` is present, which fires after the panels. Status messages queued near the end of `client.Deploy` (e.g. "DNS records provisioned successfully") can interleave with the ASCII panels, especially when stdout and the slog sink land on the same descriptor (`nic deploy ... 2>&1 | tee deploy.log`).

Fix: call `cleanup()` eagerly right after `client.Deploy` returns and before the `printXxx` calls, matching the original sequence. The `defer` can stay as a safety net.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,540 @@
...
+				LandingPage: argocd.LandingPageConfig{
+					RedisPassword: genPwd(),
+				},
+				// Enable MetalLB only for providers that need it
+				MetalLB: argocd.MetalLBConfig{
+					Enabled:     infraSettings.NeedsMetalLB,
+					AddressPool: infraSettings.MetalLBAddressPool,
+				},
+			}
+			if pwErr != nil {
```
dcmcand: **Blocker - batched error check is misleading for the foundational secrets block.**

The `genPwd` closure captures `pwErr` and the check at this line happens *after* the entire `foundationalCfg` struct literal has been evaluated. By the time we check `pwErr`, `AdminPassword`, `DBPassword`, `PostgresAdminPassword`, `PostgresUserPassword`, `RealmAdminPassword`, and `RedisPassword` have already been assigned into the struct - so the comment at L160-162 claiming we "check pwErr at the end of each generation batch before any secret is consumed" doesn't match reality. The secrets were consumed by struct-literal evaluation.

In practice `rand.Reader` won't fail on a healthy Linux host, so this is unlikely to bite. But the gap between the stated security contract and the actual behavior is itself a correctness concern - and the same pattern was applied successfully a few lines up at L175-180 for `argoCDClientSecret`.

Fix: either check `pwErr` after each `genPwd()` call individually (annoying but matches the comment), or just call `generateSecurePassword(rand.Reader)` directly inline and return on error immediately. The single-secret pattern at L175-180 is already correct and could be the template.

pkg/nic/registry.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return &Providers{
+		Cluster: c.registry.ClusterProviders.List(ctx),
+		DNS:     c.registry.DNSProviders.List(ctx),
+	}
+}
+
+// defaultRegistry builds a Registry with all in-tree cluster and DNS
+// providers registered.
+func defaultRegistry(ctx context.Context) (*registry.Registry, error) {
```
dcmcand, line 44: **Question.** `defaultRegistry` takes a `ctx` and passes it to each `Register` call, but `NewClient` calls this with `context.Background()` (client.go:19). If the `Register` implementations don't actually do cancellable work, the ctx is vestigial from the old CLI `init()` shape and could be dropped.

If they do (or might in the future), consider promoting it: `func NewClient(ctx context.Context) (*Client, error)`. That lets library consumers attach a deadline or cancellation to the registration step, which is the only place `Background()` shows up in the new public surface.

cmd/nic/deploy.go
```diff
@@ -52,61 +39,13 @@ func init() {
...
-	return gitCfg, nil
-}
-
 func runDeploy(cmd *cobra.Command, args []string) error {
-	// Get cancellable context from cobra (for signal handling)
 	ctx := cmd.Context()
 
-	// Resolve config file path via auto-discovery if not explicitly provided.
-	resolved, err := resolveConfigFile(deployConfigFile)
+	deployConfigFile, err := resolveConfigFile(deployConfigFile)
```
dcmcand: **Question.** `deployConfigFile, err := resolveConfigFile(deployConfigFile)` introduces a local `deployConfigFile` that shadows the package-level `var deployConfigFile string` declared at L18 and used by the cobra flag at L36. It compiles and behaves correctly because L55, L70, L73 all use the local. But the pre-refactor code used a `resolved` intermediate that made the relationship explicit.

Is the shadow intentional? If so, a comment would help - otherwise a future reader may try to read `deployConfigFile` elsewhere in the file and get the unset package-level value.

pkg/nic/status.go
```diff
@@ -0,0 +1,53 @@
...
+
+// StartSlogHandler wires Client progress to the supplied slog logger. Pass
+// the returned ctx to subsequent Client calls, and defer the returned
+// cleanup to flush in-flight updates. Consumers who want a non-slog sink
+// should call status.StartHandler with their own handler instead.
+func StartSlogHandler(ctx context.Context, logger *slog.Logger) (context.Context, status.CleanupFunc) {
+	return status.StartHandler(ctx, SlogHandler(logger))
+}
+
+func mapSlogLevel(l status.Level) slog.Level {
```
dcmcand, line 40: **Question.** `SlogHandler` is exported but `mapSlogLevel` is not. A library consumer who wants to wrap or compose `SlogHandler` (e.g. to add fields, fan out to multiple sinks) cannot reproduce the status.Level to slog.Level mapping without re-implementing it. Is keeping it unexported intentional? If yes, no action needed; flagging in case the symmetry was an oversight.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,540 @@
...
+	status.Send(ctx, status.NewUpdate(status.LevelSuccess, "Deployment completed successfully").
+		WithMetadata("provider", clusterProvider.Name()))
+
+	result := &DeployResult{}
+
+	// Install Argo CD (skip in dry-run mode)
+	if !opts.DryRun {
+		status.Progress(ctx, "Installing Argo CD on cluster")
+
+		// Generate all deployment secrets up front. genPwd shares a single
```
dcmcand: **Stylistic.** `Deploy` runs ~120 lines across roughly seven phases (validate, provision, resolve GitOps config, bootstrap GitOps, install ArgoCD, install foundational services, look up LB endpoint). It's an orchestration function so the size is defensible, but extracting the ArgoCD + foundational-services block into a private `installSoftwareLayer` helper would also give a natural home for proper per-call password error handling and shrink the parent to a more scannable spine.

dcmcand (changes requested):
Re-review of the orchestration extraction. Overall this is a well-executed refactor: `cmd/nic/deploy.go` shrinks from ~480 LOC to a thin wrapper, no provider-name switches leak in, `InfraSettings` is used correctly, and several pre-existing issues were actively improved during the move (compile-checked `scrubbedConfig`, `generateSecurePassword` failing closed, `*git.Config` decoupling in `pkg/argocd`).

Requesting changes for two correctness issues:

**Blockers**
1. Status cleanup ordering regression in `cmd/nic/deploy.go` (inline).
2. Batched `pwErr` check in `pkg/nic/deploy.go` happens after secrets are already assigned (inline).

**Questions** (inline)
- `pkg/nic/registry.go`: `context.Background()` usage in `NewClient`.
- `cmd/nic/deploy.go`: shadowed `deployConfigFile`.
- `pkg/nic/status.go`: `mapSlogLevel` unexported.

**Architecture notes (non-blocking)**
- Several private helpers in `pkg/nic` perform distinct I/O without their own spans: `getOrCreateGitConfig`, `lookupEndpointAndProvisionDNS`, `writeConfigToRepo`, `destroyDNS`. CLAUDE.md says "All Functions Must Follow" the span pattern; either add spans or amend the doc to exempt private helpers covered by a parent span.
- `pkg/nic/status.go` importing `log/slog` to expose `SlogHandler` doesn't violate the no-logging-in-pkg rule (it's an adapter), but it does dilute the cmd-as-slog-seam pattern. Consider relocating to `pkg/nic/sloghandler` so consumers who don't want slog don't pull it transitively.

**Stylistic**
- PR description sample shows `client := nic.NewClient()` but the real signature returns `(*Client, error)`. New consumers reading the description will hit a compile error.
- `scrubbedConfig` shallow-copies `*cfg`; pointer fields still alias. Safe today, but a one-line comment future-proofs it.
- Interrupt-warning `defer` is present on `Deploy` but no longer on `Destroy` (it was there pre-refactor). Asymmetry worth a comment if intentional.

**Highlights**
- Compile-checked struct scrub beats the old string deny-list - a renamed sensitive field forces an explicit decision instead of silent drift.
- Decoupling `pkg/argocd` from `*config.NebariConfig.GitRepository` in favor of an explicit `*git.Config` parameter makes those functions independently testable and composable.
- `generateSecurePassword` failing closed (vs. timestamp fallback) is a real security improvement, and the test documents the contract.
- `Confirm` callback on `DestroyOptions` is a clean inversion that keeps the TTY prompt in `cmd/nic`.
- All exported `Client` methods are span-instrumented.

cmd/nic/deploy.go
marcelovilla: Good catch, thanks. Addressed in 999840eab1876d01faab08329e716ab826dcd52c by matching the original sequence and making sure the `cleanup` function is idempotent to avoid a panic after trying to close an already closed channel.

pkg/nic/deploy.go
marcelovilla: I addressed this in 5cc409684eb320dd926a97d5b84c84e95eb1f479 by ditching the `genPwd` function and calling `generateSecurePassword(rand.Reader)` directly for each secret. While more verbose, I also feel it's clearer.

pkg/nic/registry.go
marcelovilla: I was not fully convinced the client should take a context but I think you raise good points here. I've updated the `NewClient` function to take a context in 66aee004217f68c930936ed627f25fb1d88dfeb5

cmd/nic/deploy.go
marcelovilla: Good catch, this was not intentional. I've updated the variable in 2c8e5de1b0f507d7d81e58a03ef06a34d582bc70

pkg/nic/deploy.go
marcelovilla: I fully agree and that would make it cleaner. I won't be doing it in this PR though as it is out of the initial scope but I'd like to spend some time looking into it in the near future.

pkg/nic/status.go
marcelovilla: This was intentional as `SlogHandler` is already aware of `mapSlogLevel`: https://github.com/nebari-dev/nebari-infrastructure-core/blob/2c8e5de1b0f507d7d81e58a03ef06a34d582bc70/pkg/nic/status.go#L28

Consumers wrapping the handler already get the mapping.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,566 @@
...
+		return fmt.Errorf("create config directory: %w", err)
+	}
+	if err := os.WriteFile(configDest, configBytes, 0600); err != nil {
+		return fmt.Errorf("write config to repository: %w", err)
+	}
+	status.Send(ctx, status.NewUpdate(status.LevelInfo, "Wrote NIC config to repository (auth fields scrubbed)").
+		WithMetadata("path", configDest))
+	return nil
+}
+
```
dcmcand, line 496: Nice - taking an `io.Reader` here lets the tests pass deterministic bytes and the error-returning shape (no fallback) prevents the prior weak-password-on-failure regression. Test coverage in `deploy_test.go:14-89` is exemplary.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,566 @@
...
+			WithMetadata("url", gitConfig.URL))
+	}
+	return nil
+}
+
+// writeConfigToRepo serialises cfg (with sensitive fields scrubbed and the
+// effective gitConfig substituted in) and writes the result into the git
+// working directory. Sourcing from the parsed config keeps this feature
+// available to library consumers who don't construct cfg from a file.
+func (c *Client) writeConfigToRepo(ctx context.Context, cfg *config.NebariConfig, gitConfig *git.Config, workDir string) error {
```
dcmcand, line 479: Struct-typed scrubbing is the right call - if someone adds a sensitive field upstream, this fails to compile. Much better than a string deny-list.

pkg/nic/status.go
```diff
@@ -0,0 +1,53 @@
...
+	"log/slog"
+
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/status"
+)
+
+// SlogHandler renders status Updates as slog records on the supplied logger.
+// It returns the handler function itself, so callers can wrap it (e.g., to
+// filter, add fields, or fan out to a second sink) before passing the result
+// to status.StartHandler. For plain slog-only integration, use StartSlogHandler.
+func SlogHandler(logger *slog.Logger) status.Handler {
```
dcmcand, line 14: Returning the handler function rather than installing it directly is the right level of abstraction - callers can wrap, filter, or fan out before calling `status.StartHandler`. Composable.

pkg/nic/registry.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/config"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/dnsprovider/cloudflare"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/provider/aws"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/provider/azure"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/provider/existing"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/provider/gcp"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/provider/hetzner"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/provider/local"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/registry"
```
dcmcand, line 17: **Question:** importing every concrete provider here means any library consumer that pulls in `pkg/nic` transitively gets every cloud SDK. The CLI use case wants the one-line constructor, but out-of-tree consumers (the future ADR-0004 motivation) can't opt out today. Worth a maintainer call: keep this and accept the SDK fanout, or add `NewClientWithRegistry(reg)` so consumers can build a minimal registry. No need to decide here; the current shape is fine for the CLI.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,566 @@
...
+					WithMetadata("error", err.Error()))
+				return nil, fmt.Errorf("generate Postgres user password: %w", err)
+			}
+			realmAdminPassword, err := generateSecurePassword(rand.Reader)
+			if err != nil {
+				span.RecordError(err)
+				status.Send(ctx, status.NewUpdate(status.LevelError, "Failed to generate foundational secrets").
+					WithMetadata("error", err.Error()))
+				return nil, fmt.Errorf("generate realm admin password: %w", err)
+			}
```
dcmcand: **Stylistic:** the `genPwd` closure + 6 password fields + 2 error checks span 60+ lines and dominate the orchestrator. Extracting `generateFoundationalSecrets(r io.Reader) (foundationalSecrets, error)` would tighten the flow.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,566 @@
...
+				span.RecordError(err)
+				status.Send(ctx, status.NewUpdate(status.LevelError, "Failed to generate foundational secrets").
+					WithMetadata("error", err.Error()))
+				return nil, fmt.Errorf("generate Redis password: %w", err)
+			}
+
+			foundationalCfg := argocd.FoundationalConfig{
+				Keycloak: argocd.KeycloakConfig{
+					Enabled:               true,
+					AdminUsername:         "admin",
```
dcmcand, line 196: **Stylistic:** `kubectl apply -f pkg/foundational/` references an internal repo path. The string is carried verbatim from the old CLI, but now that this ships in a library, an external consumer sees a path that doesn't exist in their tree. Suggest a generic message ("install foundational services manually") or drop the command.

pkg/nic/deploy.go
```diff
@@ -0,0 +1,566 @@
...
+				LandingPage: argocd.LandingPageConfig{
+					RedisPassword: redisPassword,
+				},
+				// Enable MetalLB only for providers that need it
+				MetalLB: argocd.MetalLBConfig{
+					Enabled:     infraSettings.NeedsMetalLB,
+					AddressPool: infraSettings.MetalLBAddressPool,
+				},
+			}
+
```
dcmcand, line 217: **Stylistic:** `os.TempDir() + "/nebari-gitops-<project>"` is keyed only by project name, which risks collisions across processes for library consumers. Either add a process/random suffix or accept an explicit base dir on `DeployOptions`.

cmd/nic/version.go
```diff
@@ -26,7 +27,7 @@ var versionCmd = &cobra.Command{
 func runVersion(cmd *cobra.Command, args []string) error {
 	ctx := context.Background()
 	tracer := otel.Tracer("nebari-infrastructure-core")
-	_, span := tracer.Start(ctx, "cmd.version")
+	ctx, span := tracer.Start(ctx, "cmd.version")
```
dcmcand, line 30: **Stylistic:** uses `context.Background()` while sibling commands (`deploy`, `destroy`, `validate`, `kubeconfig`) use `cmd.Context()`. Minor inconsistency, but it means a signal interrupt during `version` wouldn't propagate.

dcmcand (dismissed):
This PR cleanly extracts the orchestration layer into `pkg/nic` so library consumers can drive deploys without owning the CLI plumbing. The boundary between `cmd/nic` (flags, prompts, panels) and `pkg/nic` (orchestration) is well respected, and the new code is consistent with the documented invariants: provider access through the interface only, no provider-name switches, every public function span-instrumented, no `slog` calls inside `pkg/` other than the explicit `SlogHandler` adapter, errors wrapped with `%w`.

Findings below are split into two questions that need a maintainer decision, and a set of stylistic suggestions. **None are blockers.** You can address them inline, file follow-ups, or ignore and merge - your call.

dcmcand:
A few cross-cutting items that don't fit inline:

**Question - `Provider.Validate` is dead code on the runtime paths.** `pkg/nic/{deploy,destroy,kubeconfig,validate}.go` only call `cfg.Validate(...)` (config-package); the `Provider.Validate(ctx, ...)` interface method is never invoked anywhere outside tests. This is pre-existing (the old `cmd/nic/deploy.go` did the same), but the interface docstring says "checks that the configuration is valid before any infrastructure operations" - strongly implying it's called. Either delete the interface method or wire it in. Worth a follow-up issue. (Concrete impact: PR #330's new `load_balancer_scheme` validation lives in `aws.Provider.Validate` and therefore currently never runs in real deploys.)

**Stylistic - inconsistent error wrapping in `Deploy`.** `pkg/nic/deploy.go:104-111` and `:117-123` return the raw error from `reg.ClusterProviders.Get` and `clusterProvider.Deploy`. Most siblings wrap with `fmt.Errorf("...: %w", err)`. Either choice is reasonable; the inconsistency makes downstream `errors.Is/As` matching unpredictable.

**Stylistic - `NewClient()` discards context.** Calls `defaultRegistry(context.Background())`, so OTel spans created during provider registration are orphaned from the caller's trace. Take a `ctx` parameter, or document that registration spans are detached.

**Stylistic - orchestration has no tests.** `generateSecurePassword` and `scrubbedConfig` are well-covered; `Deploy/Destroy/Validate/Kubeconfig` themselves aren't. The `Destroy` `Confirm` callback path is reachable with a fake provider and is worth at least a smoke test.

**Doc gap (suggested follow-up PR):** ARCHITECTURE.md doesn't yet describe `pkg/nic` as the library entry point or the new two-layer arrangement (CLI -> `pkg/nic.Client` -> `provider.Provider`).

pkg/nic/registry.go
marcelovilla: I'm fine with keeping it for now, knowing that once we start working on out-of-tree consumers, we'll make it so only the required providers are registered dynamically.

pkg/nic/deploy.go
marcelovilla: Even though this comment is outdated (there's no `genPwd` closure anymore), I went ahead and extracted all the inline password generation calls into its own helper function. See 3259d44083dfd33402643cb6be275ae22b52361e

pkg/nic/deploy.go
marcelovilla: Good catch. I'm dropping the comment as there's no much help in telling users to install the services manually if we don't provide a guide on how to do it. See 7816683af55ab638223adccabd3aa9f30b87f50e

cmd/nic/version.go
marcelovilla: Thanks for catching that inconsistency. Addressed in 3d88fda2130bbc6265343077fe1c45abe9054c2b

pkg/nic/deploy.go
marcelovilla: Opened https://github.com/nebari-dev/nebari-infrastructure-core/issues/337 to follow up on this

marcelovilla:
Thanks for the re-review @dcmcand!

I'll address your cross-cutting items here:
> Question - Provider.Validate is dead code on the runtime paths. pkg/nic/{deploy,destroy,kubeconfig,validate}.go only call cfg.Validate(...) (config-package); the Provider.Validate(ctx, ...) interface method is never invoked anywhere outside tests. This is pre-existing (the old cmd/nic/deploy.go did the same), but the interface docstring says "checks that the configuration is valid before any infrastructure operations" - strongly implying it's called. Either delete the interface method or wire it in. Worth a follow-up issue. (Concrete impact: PR https://github.com/nebari-dev/nebari-infrastructure-core/pull/330's new load_balancer_scheme validation lives in aws.Provider.Validate and therefore currently never runs in real deploys.)

This is something I'm aware of and something that's being tracked [here](https://github.com/nebari-dev/nebari-infrastructure-core/issues/72). I still need to take the time to work on that before we have a stable release.

> Stylistic - inconsistent error wrapping in Deploy. pkg/nic/deploy.go:104-111 and :117-123 return the raw error from reg.ClusterProviders.Get and clusterProvider.Deploy. Most siblings wrap with fmt.Errorf("...: %w", err). Either choice is reasonable; the inconsistency makes downstream errors.Is/As matching unpredictable.

Addressed in 4d46caeb3feb40c5bd5fcfeafb263a56c0dc3322.

> Stylistic - NewClient() discards context. Calls defaultRegistry(context.Background()), so OTel spans created during provider registration are orphaned from the caller's trace. Take a ctx parameter, or document that registration spans are detached.

This is not true. See https://github.com/nebari-dev/nebari-infrastructure-core/blob/3d88fda2130bbc6265343077fe1c45abe9054c2b/pkg/nic/client.go#L20

> Stylistic - orchestration has no tests. generateSecurePassword and scrubbedConfig are well-covered; Deploy/Destroy/Validate/Kubeconfig themselves aren't. The Destroy Confirm callback path is reachable with a fake provider and is worth at least a smoke test.

I opened  https://github.com/nebari-dev/nebari-infrastructure-core/issues/338 to track this.

> Doc gap (suggested follow-up PR): ARCHITECTURE.md doesn't yet describe pkg/nic as the library entry point or the new two-layer arrangement (CLI -> pkg/nic.Client -> provider.Provider).

I opened  https://github.com/nebari-dev/nebari-infrastructure-core/issues/339 to track this

viniciusdc (approved):

PR #280, opened 2026-05-05
fix(lint): resolve goconst violations across packages

dcmcand:
## Summary

CI's golangci-lint job has been failing on main since the goconst rules tightened in golangci-lint v2.12.1. This PR clears all 39 goconst issues so CI goes green.

- **`.golangci.yml`**: bump `goconst.min-occurrences` from the default 3 to 4, and add `ignore-string-values: ^(true|false|Config|name)$`. Together these skip strings that are too incidental or too generic to meaningfully name as constants.
- **Code**: extract named constants for the meaningful repeats: Argo CD chart version and namespace, the `argocd-server` deployment name, the `local-gitops` volume name, the AWS IAM Authenticator kubeconfig exec args (`eks`, `get-token`, `--cluster-name`, `--region`, `client.authentication.k8s.io/v1beta1`), Longhorn webhook security-group keys and node label, the `0.0.0.0/0` default CIDR allowlist, and the local provider's `standard` StorageClass and MetalLB address pool defaults.
- Reuse the existing `aws.ProviderName` for the kubeconfig exec command; add a matching `local.ProviderName`.

The lint failures are pre-existing on main (not introduced by any open PR), so this lands as a standalone fix.

## Test plan

- [x] `golangci-lint run` reports 0 issues locally (v2.12.1)
- [x] `go build ./...` succeeds
- [x] `go test ./...` passes
- [ ] CI passes on this PR

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,6 +23,14 @@ linters:
     - unconvert
     - unused
 
+  settings:
+    goconst:
+      # Default is 3, which flags many genuinely-incidental repeats.
+      # Bump to 4 so only repeated strings that clearly warrant a constant get flagged.
+      min-occurrences: 4
+      # Skip strings that are too generic to meaningfully name as constants.
+      ignore-string-values: "^(true|false|Config|name)$"
+
   exclusions:
     rules:
       # Test files can have repeated strings for readability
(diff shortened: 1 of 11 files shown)


marcelovilla (approved):
LGTM @dcmcand! Thanks for taking care of this

Issue #66, opened 2026-02-06
Pipe OpenTofu output through slog and add pretty-print option for structured logs


dcmcand:
## Description

Currently OpenTofu output goes directly to stdout/stderr, bypassing the structured logging (slog) that the rest of the application uses. This creates inconsistent output where application messages are structured JSON but OpenTofu output is raw text interleaved with it.

Two improvements needed:

1. **Route OpenTofu output through slog** — OpenTofu stdout/stderr should be captured and emitted as structured log entries, consistent with the rest of NIC's output. This makes logs parseable and filterable.

2. **Add a pretty-print option for structured logs** — JSON logs are great for machines but hard to read for humans during interactive use. Add a flag (e.g., `--log-format=pretty` or `--log-format=text`) that formats structured log output in a human-readable way (colored, aligned, timestamps).

## Use Case

- Operators piping NIC output to log aggregators get consistent structured output for all operations, not a mix of JSON and raw text
- Developers running NIC interactively get readable output instead of JSON blobs
- OpenTofu plan/apply output becomes searchable and filterable alongside application logs

PR #281, opened 2026-05-06
Stream OpenTofu output through the status channel

marcelovilla:
## Closes
Partially addresses #66 

## Description

This PR routes OpenTofu's machine-readable (`-json`) output through `pkg/status` by giving the executor's `Init`/`Plan`/`Apply`/`Destroy` wrappers an `io.Writer` that parses each event line and forwards it as a `status.Update` (with the full event preserved as `Metadata["payload"]`). This way, `statusLogHandler` can then render it using `slog`, along with the rest of the application logs.

Furthermore, this PR bumps the `DefaultChannelSize` from 100 to 1000 as OpenTofu can generate more than 100 messages on a fresh deployment and leaving the default means some of the messages would start getting lost.

## How to Test

Build and run:

```sh
./nic deploy -f examples/aws-config.yaml --dry-run
```

You should see OpenTofu's output as structured logs.

# Notes

I opened https://github.com/nebari-dev/nebari-infrastructure-core/issues/282 as a follow-up issue to make sure we have a standard way of specifying the source package in the logs via the status channel.


pkg/status/writer.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"bytes"
+	"context"
+	"io"
+)
+
+// MetadataKeyPayload is the conventional Metadata key on Updates carrying a
+// structured per-source payload (typically a json.RawMessage). The shape of
+// the payload depends on the producer.
+const MetadataKeyPayload = "payload"
```
marcelovilla: I'm not a huge fan of the name `payload` but I think we need a field to add original data from third party events, such as OpenTofu. Let me know if there's other name that makes more sense or if eventually we would like to have each field from the event be a different field as metadata.

dcmcand: Looking into this, something generic but explanatory like `additionalFields` or `extraFields` might be good 🤷

pkg/tofu/log.go
```diff
@@ -0,0 +1,53 @@
...
+// typed fields, and forwards the full event as the payload. Lines that
+// don't parse as JSON fall through as raw text at info level so unexpected
+// output is never lost.
+func jsonLineMapper(line []byte) status.Update {
+	var ev tofuEvent
+	if err := json.Unmarshal(line, &ev); err != nil {
+		return status.NewUpdate(status.LevelInfo, string(line))
+	}
+	// Copy the line so the payload isn't aliasing the Writer's internal
+	// buffer (which gets reused on the next Write).
```
dcmcand, line 48: Nice catch on the defensive copy. If we reused the bytes.Buffer slice, the next Write would alias into the payload we already handed off, which is the kind of bug that only shows up once in a while and ruins a Tuesday. Keeping the rationale next to the copy helps the next person reading this.

pkg/status/writer.go
```diff
@@ -0,0 +1,89 @@
...
+type LineMapper func(line []byte) Update
+
+// Writer is an io.Writer that splits the bytes written to it on newlines and
+// emits one status.Update per non-blank line via the configured LineMapper.
+//
+// Partial lines are buffered across Write calls until a newline arrives; call
+// Flush after the producer finishes writing to drain a final line that wasn't
+// newline-terminated.
+//
+// Writer is not safe for concurrent use. tfexec and os/exec each drive a
```
dcmcand, line 28: The single-writer note plus the "tfexec drives a single io.Copy goroutine per stream" rationale is the kind of invariant I want spelled out. Saves the next person from having to reverse-engineer it from tfexec's source. Thanks for writing it down.

pkg/tofu/log.go
```diff
@@ -0,0 +1,53 @@
...
+		return status.LevelInfo
+	}
+}
+
+// jsonLineMapper is the status.LineMapper used for tofu's `-json` stdout. It
+// parses each event line, extracts @level and @message into the Update's
+// typed fields, and forwards the full event as the payload. Lines that
+// don't parse as JSON fall through as raw text at info level so unexpected
+// output is never lost.
+func jsonLineMapper(line []byte) status.Update {
```
dcmcand, line 42: I like this shape. Pulling out @level and @message into typed fields and shipping the rest of the event through as Metadata["payload"] means new tofu event types and sub-fields flow through without us having to touch the producer. The case-table in log_test.go does a good job of pinning that down.

pkg/status/writer_test.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		if got[0].Message != "connection refused" {
+			t.Errorf("message = %q", got[0].Message)
+		}
+	})
+}
+
+// TestConventionKeys pins the documented Metadata key names so a rename forces
+// a deliberate update across all consumers.
+func TestConventionKeys(t *testing.T) {
```
dcmcand, line 133: Small test, big leverage. Pinning MetadataKeyPayload = "payload" forces anyone who renames the key to update the consumers on purpose, instead of finding out at runtime.

pkg/tofu/tofu.go
```diff
@@ -30,27 +33,67 @@ func (te *TerraformExecutor) Cleanup() error {
...
+// ctx for the duration of op. The op callback receives the stdout writer to
+// pass into a tfexec *JSON method (which handles SetStdout itself); stderr
+// is set on the executor directly since the JSON variants only touch stdout.
+// Both writers buffer partial lines, so we Flush both after op returns to
+// drain any final non-newline-terminated content.
+func (te *TerraformExecutor) streamThroughStatus(ctx context.Context, op func(io.Writer) error) error {
+	stdout := status.NewWriter(ctx, jsonLineMapper)
+	stderr := status.NewWriter(ctx, status.RawMapper(status.LevelError))
+
+	te.SetStderr(stderr)
```
dcmcand, line 46: No action needed here. te.SetStderr(stderr) mutates state on the embedded *tfexec.Terraform and we never restore it, so the executor keeps a *status.Writer with a captured ctx after the op returns. In general that is a hazard, because a later tfexec method that emits to stderr would route into a stale Writer tied to a possibly-canceled ctx. It is fine here for three reasons: the CLI is single-threaded for infra ops, Output is documented as not touching stderr, and the next Init / Plan / Apply / Destroy overwrites the setter anyway. Worth a mental note if Output ever grows a stderr path, or if we add a new wrapper method.

pkg/status/status.go
```diff
@@ -9,7 +9,7 @@ import (
 
 const (
 	// DefaultChannelSize is the default buffer size for the status channel
-	DefaultChannelSize = 100
+	DefaultChannelSize = 1000
```
dcmcand, line 12: Could you add a quick note to the PR description about why DefaultChannelSize jumps from 100 to 1000?

pkg/status/writer.go
```diff
@@ -0,0 +1,89 @@
...
+// single io.Copy goroutine per stream, so single-writer is the only shape we
+// see in practice.
+type Writer struct {
+	ctx    context.Context
```
dcmcand, line 32: No action needed here. Storing ctx on a struct is generally discouraged by Go style and by contextcheck-style lints, because it hides cancellation propagation from the call site. It is fine here because io.Writer.Write(p []byte) (int, error) has no ctx parameter, so this is the only practical shape for the seam. One small suggestion: add a line to the doc comment along the lines of "ctx is captured at construction; do not reuse a Writer across operations with different cancellation scopes." That captures the constraint where someone is most likely to read it.

cmd/nic/status_handler.go
```diff
@@ -1,48 +1,42 @@
...
-			attrs = append(attrs, "action", update.Action)
+			attrs = append(attrs, slog.String("action", update.Action))
 		}
-
-		// Add metadata as individual attributes
 		for key, value := range update.Metadata {
-			attrs = append(attrs, key, value)
+			attrs = append(attrs, slog.Any(key, value))
 		}
+		slog.LogAttrs(context.Background(), mapSlogLevel(update.Level), update.Message, attrs...)
```
dcmcand, line 25: No action needed here. slog.LogAttrs(context.Background(), ...) would normally be a flag for dropping ctx propagation. It is fine here because the handler runs in the goroutine started by StartHandler, which does not carry the caller's ctx anyway. A one-line comment such as "handler runs detached from the caller's ctx" would save the next reader the same trip I just took.

dcmcand (dismissed):
Approving. See the separate summary comment for run-through and notes; inline comments are kudos plus stylistic notes (no action needed) plus one small ask on the channel size bump.

dcmcand:
Nice piece of work. The central design choice, routing tofu's -json event stream through the existing pkg/status channel and translating to slog only inside cmd/nic/status_handler.go, is the right shape. It respects the documented "logging at the application layer only" rule, reuses the seam we already had for app-level progress, and keeps the producer side (the line mapper) ignorant of how output gets rendered. The forward-compatible payload pattern, the defensive copy with rationale, and the convention-key test are all small but high-leverage choices.

I ran `./nic deploy -f examples/aws-config.yaml --dry-run` and the output looks right end to end: app-level slog records still come through unchanged, every tofu line gets `level` and `msg` from `@level` / `@message`, and the full event rides through as `payload` with hook / change / changes / outputs blocks intact. With around 200+ events in a single dry-run, the bump from 100 to 1000 on `DefaultChannelSize` is doing real work, the old default would have started dropping messages once the handler goroutine fell behind.

Tests pass on a fresh worktree (`go test ./pkg/status/... ./pkg/tofu/...`) and `go vet` is clean. Supply-chain screening is clean too: no new deps in go.mod, no init(), no os/exec, no network calls, no unsafe, no embeds.

On issue #66: partially resolved by design, as the PR body says. The `--log-format=pretty` half stays open for a follow-up PR.

Approving. The one ask, a line in the PR description about the channel size bump, is on the comment, not blocking.

pkg/status/status.go
marcelovilla: Just added a note on the PR description

pkg/status/writer.go
marcelovilla: I ended up renaming it do `detail` as I think it conveys better the fact that it contains the original raw JSON message. Happy to change it in the future if you have strong feelings about it.

dcmcand (approved):

PR #301, opened 2026-05-12
docs(design-doc): audit and rewrite design docs against current code

dcmcand:
## Summary

The design docs under `docs/design-doc/` had drifted substantially from the codebase. This PR audits every file against current code and rewrites the heavily-drifted ones from scratch; the rest get surgical fixes.

Closes #300.

## What changed

**Heavily rewritten (substantially wrong before):**
- `architecture/02-system-overview.md`, `04-key-decisions.md`, `05-state-management.md`
- `implementation/06-opentofu-module-architecture.md`, `07-configuration-design.md`, `08-terraform-exec-integration.md`, `10-foundational-software.md`, `11-nebari-operator.md`
- `appendix/16-configuration-reference.md`
- `operations/12-testing-strategy.md`, `13-milestones.md`

**Surgical edits:**
- `nic-summary.md`, `architecture/01-introduction.md`, `architecture/03-goals-and-non-goals.md`, `implementation/09-dns-provider-architecture.md`, `appendix/14-open-questions.md`, `appendix/15-future-enhancements.md`, `appendix/17-appendix.md`, `operations/longhorn-node-maintenance.md`

Net diff: +1415 / -4952 (mostly removing fictional content).

## Highlights of what was wrong

- **Provider story**: docs assumed a uniform OpenTofu-driven multi-cloud implementation. Reality: only AWS uses tofu. Hetzner uses the `hetzner-k3s` binary. `local` is a Kind stub via `make localkind-up`. `existing` is a no-op. GCP and Azure are stubs. Hetzner and `existing` weren't documented at all. [ADR-0004](docs/adr/0004-out-of-tree-provider-plugins.md) is now cross-referenced where relevant.
- **Package layout**: docs referenced a root-level `terraform/modules/{aws,gcp,azure,local,kubernetes,argocd,foundational-apps}/` tree that doesn't exist. AWS templates actually live in `pkg/provider/aws/templates/`. References to `pkg/operator`, `pkg/tofu/executor.go`, `pkg/tofu/workspace.go`, `pkg/tofu/outputs.go`, `pkg/kubernetes/`, `api/v1alpha1/` were removed (none of those exist).
- **CLI commands**: invented `nic plan`, `nic status`, `nic state list/show/rm/mv`, `nic unlock`, `nic init-backend`, `nic health check`, `nic stack ...`, `nic init`, `nic marketplace` removed. Real verbs documented: `deploy`, `destroy`, `validate`, `kubeconfig`, `version`.
- **Foundational software**: the LGTM stack (Loki / Grafana / Tempo / Mimir) was presented as deployed. It isn't; only opentelemetry-collector ships today. The full app set under `pkg/argocd/templates/apps/` is now documented correctly.
- **Config schema**: the configuration reference documented a top-level `provider:` field with sibling `amazon_web_services:` / `google_cloud_platform:` / `azure:` / `hetzner_cloud:` / `local:` keys. The real schema is `cluster.<provider-name>:` with no top-level `provider:` field. Only the Hetzner section had been correct. The reference now matches `pkg/config/config.go` and the per-provider config packages, and adds the missing `certificate:`, `git_repository:`, and `existing` provider sections.
- **CRD name**: `NicApp` / `NebariApplication` corrected to `NebariApp` throughout. The operator is reframed as an out-of-tree project at `github.com/nebari-dev/nebari-operator`; NIC just deploys it.
- **State management (`05-state-management.md`)**: DynamoDB-based locking replaced with the real S3 native lockfile (`use_lockfile = true`). Fake `state_backend:` config block removed. Fake `DetectDrift` code sample removed.
- **Testing strategy / milestones**: invented `nic health check` subsystem removed. CI YAML corrected to match `.github/workflows/ci.yml` (Go 1.25.1, real test command, no fictional integration / scheduled jobs). Mocking libraries corrected (`moto` / `fake-gcs-server` / `azurite` -> LocalStack via `docker-compose.test.yml`). Milestones no longer mark GCP, Azure, Grafana dashboards, multi-cloud CI, or v1.0.0 as shipped; ADR-0004 referenced.

## Test plan

- [x] `go build ./...` clean
- [x] `make lint` clean
- [x] `make test` clean (pre-commit hooks ran on commit)
- [ ] Maintainer review: spot-check claims in each rewritten doc against current code

## Notes

- The full audit reports the rewrites are based on are in the conversation that produced this PR; happy to drop them in if useful.
- Some items in `appendix/14-open-questions.md` and `appendix/15-future-enhancements.md` have shipped (e.g., git repo *consumption* via `git_repository:`, `.env`-based secrets). Those are flagged inline rather than removed wholesale, so the future-work appendix stays useful.


dcmcand:
@khuyentran1401 I updated the docs here, that should hopefully help you out

docs/design-doc/implementation/11-nebari-operator.md
```diff
@@ -1,485 +1,107 @@
...
-      lastTransitionTime: "2025-01-30T12:02:00Z"
-      reason: AllComponentsReady
-      message: "Application is fully configured and accessible"
+  hostname: jupyter.example.com
+  routing:
+    routes:
+      - path: /
+        backend:
+          name: jupyterhub
+          port: 8000
```
khuyentran1401: A few fields in this example didn't match the upstream types when I cross-checked:

- apiVersion is `reconcilers.nebari.dev/v1` (from [`groupversion_info.go`](https://github.com/nebari-dev/nebari-operator/blob/main/api/v1/groupversion_info.go)), not `nebari.dev/v1`.
- `routing.routes[].path` is `pathPrefix` (see [`nebariapp_types.go`](https://github.com/nebari-dev/nebari-operator/blob/main/api/v1/nebariapp_types.go) `RouteMatch`).
- `RouteMatch` has no `backend`; `service` sits at the top of the spec.

The rest of the example matches.

docs/design-doc/architecture/02-system-overview.md
```diff
@@ -6,131 +6,171 @@
...
+- The nebari-operator app references the upstream repository (`github.com/nebari-dev/nebari-operator`) via Kustomize; the operator's source code does not live in this repo.
+
+**`pkg/dns`/`pkg/endpoint`/`pkg/git`/`pkg/helm`/`pkg/kubeconfig`/`pkg/status`/`pkg/telemetry`**
+
+- `pkg/endpoint` waits for the Envoy Gateway `Service` to receive an LB hostname or IP, so the CLI can either provision DNS or print manual instructions.
+- `pkg/git` clones, commits, and pushes the GitOps repo (including `file://` local paths).
+- `pkg/helm` is a thin wrapper around `helm.sh/helm/v3/pkg/action` used by `pkg/argocd`.
+- `pkg/status` is the in-process status channel used to surface user-visible progress from library code without violating the "no `slog` in `pkg/`" rule.
+- `pkg/telemetry` wires up the OpenTelemetry tracer provider, with exporters selected via `OTEL_EXPORTER` (`console` default, `otlp`, `both`, `none`).
```
marcelovilla: The exporter type is `none` by default, not `console`: https://github.com/nebari-dev/nebari-infrastructure-core/blob/fce61f79229b24b96366db12a57305b5fa1546e5/pkg/telemetry/telemetry.go#L26-L29

docs/design-doc/architecture/04-key-decisions.md
```diff
@@ -2,267 +2,172 @@
...
-- Performance monitoring
-- Vendor-neutral (can export to any backend)
-- Unified observability story (NIC uses same stack it deploys)
-- Compliance with industry standards
+**Decision:**
 
-**Implementation:**
+- All new functions in `pkg/` are wrapped in OpenTelemetry trace spans, with the documented exemptions in [`CLAUDE.md`](../../../CLAUDE.md) (e.g., per-line writers in `pkg/status` and byte/line helpers in `pkg/tofu`).
+- Library code never calls `slog`. User-visible progress goes through the status channel; `cmd/nic/status_handler.go` is the only translator into structured logs.
+- Exporters are configurable via `OTEL_EXPORTER` (`console` default, `otlp`, `both`, `none`) and `OTEL_ENDPOINT`.
```
marcelovilla: The exporter type is `none` by default, not `console`: https://github.com/nebari-dev/nebari-infrastructure-core/blob/fce61f79229b24b96366db12a57305b5fa1546e5/pkg/telemetry/telemetry.go#L26-L29

docs/design-doc/appendix/16-configuration-reference.md
```diff
@@ -1,1387 +1,334 @@
...
 
+| Variable | Used by | Purpose |
+|----------|---------|---------|
+| `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, `AWS_REGION` | AWS provider | Standard AWS SDK credentials |
+| `HCLOUD_TOKEN` | Hetzner provider | Hetzner Cloud API token |
+| `CLOUDFLARE_API_TOKEN` | Cloudflare DNS | Zone:Read + DNS:Edit on the configured zone |
+| `GIT_SSH_PRIVATE_KEY` (or whatever you point `git_repository.auth.ssh_key_env` at) | `pkg/git` | SSH private key in PEM form |
+| `GIT_TOKEN` (or whatever you point `git_repository.auth.token_env` at) | `pkg/git` | Personal access token for HTTPS git URLs |
+| `KUBECONFIG` | `existing` provider, `nic kubeconfig` | Kubeconfig path (used when `cluster.existing.kubeconfig` is empty) |
+| `OTEL_EXPORTER` | `pkg/telemetry` | `console` (default), `otlp`, `both`, `none` |
```
marcelovilla: The exporter type is `none` by default, not `console`: https://github.com/nebari-dev/nebari-infrastructure-core/blob/fce61f79229b24b96366db12a57305b5fa1546e5/pkg/telemetry/telemetry.go#L26-L29

docs/design-doc/implementation/10-foundational-software.md
```diff
@@ -1,387 +1,115 @@
...
-      processors: [batch]
-      exporters: [loki]
-    traces:
-      receivers: [otlp]
-      processors: [batch]
-      exporters: [otlp/tempo]
+<repo>/<path>/
+├── root.yaml                        # App-of-apps root
+├── nic-config.yaml                  # Scrubbed copy of nebari-config.yaml
+├── .nic-bootstrapped                # Marker file
```
marcelovilla: This file is `.bootstrapped` now: https://github.com/nebari-dev/nebari-infrastructure-core/blob/fce61f79229b24b96366db12a57305b5fa1546e5/pkg/git/client_impl.go#L22

docs/design-doc/nic-summary.md
```diff
@@ -2,12 +2,12 @@
...
+Nebari Infrastructure Core (NIC) is a next-generation platform that rethinks how organizations deploy and manage data science infrastructure. Unlike Nebari Classic, which delivers a monolithic, opinionated data science stack (JupyterHub, Dask, conda-store, etc.) as a single deployable unit, NIC takes a fundamentally different approach: it provides a **stable, composable foundation** upon which various workloads can be built and deployed independently. NIC is a Go CLI that provisions Kubernetes clusters and bootstraps a foundational software stack via GitOps. Each cluster provider chooses the right backing tool for its environment - OpenTofu for AWS (EKS), the `hetzner-k3s` binary for Hetzner, Kind for local development, and an `existing` adapter for pre-provisioned clusters - while GCP and Azure are stubbed and not yet implemented. The result is a platform that separates infrastructure concerns from application concerns, enabling teams to evolve each layer independently.
```
marcelovilla: Do we want to use the `next-generation` term here? I think we discussed it's not the most descriptive one.

docs/design-doc/nic-summary.md
```diff
@@ -2,12 +2,12 @@
...
+The key advantage of NIC is its **composable architecture through Software Packs**. Where Nebari Classic bundles everything together (meaning you get the full data science stack whether you need it all or not), NIC lets you choose exactly what you need. Software Packs are curated collections of open-source tools packaged as ArgoCD applications with a `NebariApp` Custom Resource that enables automatic registration with the platform. Want just JupyterHub and conda-store? Install the Data Science Pack. Need model serving capabilities? Add the Model Serving Pack (MLflow, KServe, Envoy AI Gateway). This modular approach means faster deployments, smaller attack surfaces, easier upgrades, and the flexibility to mix-and-match capabilities. Additionally, all services automatically integrate with centralized authentication (Keycloak), routing (Envoy Gateway), and TLS certificates (cert-manager) through the Nebari Operator.
```
marcelovilla: > Add the Model Serving Pack (MLflow, KServe, Envoy AI Gateway).

Should we update this to reflect the fact that the model serving pack uses llm-d instead? As a user it'd feel misleading reading this and seeing that there's no pack providing kserve.

docs/design-doc/appendix/16-configuration-reference.md
```diff
@@ -1,1387 +1,334 @@
...
+
+    # Optional: Longhorn distributed storage (default: enabled when nil)
+    # longhorn:
+    #   enabled: true
+    #   replica_count: 2
+    #   dedicated_nodes: false
+    #   node_selector: { workload: storage }
+```
+
+Fields not in `aws.NodeGroup`: `single_subnet`, per-node-group `permissions_boundary`. If you see them in older docs, they are not real.
```
marcelovilla: Do we need notes like these? I feel if our documentation is the source of truth, these are not really needed.

docs/design-doc/appendix/17-appendix.md
```diff
@@ -1,73 +1,89 @@
...
-- ✅ User can access Grafana dashboards immediately
-- ✅ User can troubleshoot via traces/logs/metrics
+**v1.0 success (planned):**
 
-### 14.4 Risks and Mitigations
+- ⏳ GCP and Azure providers functional (or replaced by out-of-tree plugins per ADR-0004)
+- ⏳ LGTM observability backend deployed by NIC
+- ⏳ Documented upgrade paths between releases
+- ⏳ End-to-end test coverage across providers
+- ⏳ AWS cluster deploy under 20 minutes from a fresh account
```
marcelovilla: Is this really a goal? I feel there's not a lot we can do to reduce the deployment time to under 20 minutes

marcelovilla (commented):
Thanks for this PR @dcmcand! I left some comments

docs/design-doc/architecture/04-key-decisions.md
```diff
@@ -2,267 +2,172 @@
...
+   ├── Envoy Gateway + gateway-config + httproutes
+   ├── postgresql + Keycloak
+   ├── metallb + metallb-config (only when InfraSettings.NeedsMetalLB)
+   ├── opentelemetry-collector
+   ├── nebari-operator (Kustomized from nebari-dev/nebari-operator)
+   └── nebari-landingpage
 ```
 
-### 4.6 Decision: Nebari Kubernetes Operator
+A full LGTM stack (Loki / Grafana / Tempo / Mimir) is not deployed today; that is roadmap work.
```
tylerpotts: > A full LGTM stack (Loki / Grafana / Tempo / Mimir) is not deployed today; that is roadmap work.

Perhaps link the LGTM pack here. I think now that the pack exists we don't need to consider it a roadmap item

dcmcand:
@tylerpotts Thanks for the review and for #510. I've merged #510 into this branch (squashed as `ac8a496`), so its config-reference corrections ship with this PR and you don't need to carry it separately. Before merging I spot-checked its claims against `main` and every one held up, including the ones asserting the existing doc was wrong: `cluster.local` really did document a `kube_context` and `storage_class` that don't exist on `local.Config`, and because that struct carries an inline `AdditionalFields map[string]any` both keys parse silently, so `nic validate` passing was never evidence they worked. Good catch.

On the LGTM comment: fixed, but the claim was repeated in 11 places across 7 files rather than just the one line, so it's worth a look at the wording I landed on. Rather than flipping the ⏳ rows to ✅ outright, I kept the distinction that LGTM is not part of the foundational stack that `nic deploy` owns, and said it installs on top as the `lgtm-pack` software pack like any other pack. Rows about NIC-specific Grafana dashboards stay pending, since `lgtm-pack` ships Grafana but nothing provisions NIC operational dashboards. I left your thread open rather than resolving it so you can check that reading.

Two other things landed while I was in here, both worth a skim:

`2865326` reconciles the docs with the six commits that hit `main` after this branch was last pushed. #481's AppProject scoping was undocumented, so there's a new §10.5 in `10-foundational-software.md` covering the three projects, the fact that they're applied via client-go at bootstrap rather than through GitOps, and that `foundational`'s scopes are derived from the embedded templates by `deriveProjectScopes` instead of hardcoded. #443's `longhorn-backup` app was missing from all five app lists. The sync-wave ordering in §10.6 turned out to be wrong rather than merely stale: it had cert-manager first and MetalLB third, where the templates put envoy-gateway and MetalLB in wave 1 and cert-manager in wave 2, so I replaced the prose list with the actual six waves and corrected the same ordering in `02-system-overview.md` and `04-key-decisions.md`. I also documented that Longhorn itself is Helm-installed by `pkg/storage/longhorn.Install` rather than being an ArgoCD Application, which the "everything foundational goes through ArgoCD" framing was obscuring.

`4db6bc7` also fixes something unrelated I tripped over: seven design docs cited `CLAUDE.md`, four of them as links to `../../../CLAUDE.md`. That file is gitignored, so those links 404 for anyone who clones. They now point at `AGENTS.md`, which carries every convention that was being cited and was already the target `02-system-overview.md` used. `docs/adr/0004-out-of-tree-provider-plugins.md` still has three such references but it's pre-existing on `main` and outside this PR's scope, so I left it for a follow-up.

Sorry for the churn on your approval: the three pushes dismissed it, hence the re-review request. Nothing structural changed in what you already looked at.

PR #315, opened 2026-05-21
Use EKS client to get kubeconfig and cache it

### LINKED ISSUE #241: perf: avoid redundant tofu init/module downloads during deploy
## Problem

During a single `nic deploy` on AWS, `tofu init` runs **4-5 times** and re-downloads terraform modules (VPC, EKS, etc.) on every run. Each invocation of `tofu.Setup()` in `pkg/tofu/tofu.go` creates a fresh temp working directory, so `.terraform/modules` is ephemeral. Provider plugins are cached via `TF_PLUGIN_CACHE_DIR`, but modules are not.

## Call chain per deploy

Each row runs a full `tofu.Setup` + `tf.Init`:

1. `provider.Deploy` - actually applies infra (`pkg/provider/aws/provider.go:288,307`)
2. Inside Deploy, if Longhorn enabled -> `p.GetKubeconfig` (`aws/provider.go:337,568`)
3. `argocd.Install` -> `prov.GetKubeconfig` (`pkg/argocd/install.go:38`)
4. `argocd.InstallFoundationalServices` -> `prov.GetKubeconfig` (`pkg/argocd/foundational.go:93`)
5. `lookupEndpointAndProvisionDNS` -> `prov.GetKubeconfig` (`cmd/nic/deploy.go:244`)

Rows 2-5 only need to read terraform outputs, not re-download modules.

## Proposal

Combine two fixes:

**A. Memoize kubeconfig at the CLI layer.** Call `provider.GetKubeconfig` once in `runDeploy` right after `provider.Deploy` returns, then pass the `kubeconfigBytes` down to:
- `argocd.Install`
- `argocd.InstallFoundationalServices`
- `lookupEndpointAndProvisionDNS`

Inside `aws.Deploy`, the Longhorn branch should also reuse the kubeconfig already fetched in the calling scope (or fetch once within Deploy and pass to `installLonghorn`). This eliminates calls 2-5 entirely.

Requires changing `argocd.Install` / `argocd.InstallFoundationalServices` signatures to accept `kubeconfigBytes []byte` (or a `*rest.Config` / `kubernetes.Interface`) instead of a `Provider`.

**B. Skip module fetching for output-only init.** In `GetKubeconfig` (and other output-only paths like `nic kubeconfig`, `nic status`), pass `tfexec.Get(false)` to `tf.Init` so it configures the backend without re-fetching modules. Cheaper per call, and makes standalone subcommands fast.

## Out of scope (possible follow-up)

A persistent per-project workdir (e.g. under `~/.cache/nic/workdirs/<project>`) that survives across runs so cold `nic deploy` reruns also skip module downloads. Worth considering but a larger change - track separately if desired.

## Definition of Done

- [ ] `argocd.Install`, `argocd.InstallFoundationalServices`, and `lookupEndpointAndProvisionDNS` accept kubeconfig bytes (or kube client) instead of `Provider`
- [ ] `runDeploy` calls `provider.GetKubeconfig` once after `provider.Deploy` and passes it to downstream callers
- [ ] `aws.Deploy`'s Longhorn branch no longer calls `p.GetKubeconfig` internally
- [ ] Output-only `GetKubeconfig` paths use `tfexec.Get(false)` on init
- [ ] Unit tests updated for the new `argocd.*` signatures (table-driven)
- [ ] Manual verification: `nic deploy` on AWS runs `tofu init` exactly once and does not re-download modules (capture before/after tofu output or timing)
- [ ] `docs/` updated if deploy-flow docs reference the old multi-init behavior

marcelovilla:
## Closes

Closes #241 

## Description

This PR updates the `GetKubeconfig` method of the `Provider` implementation in AWS by using the EKS client instead of OpenTofu to fetch the kubeconfig. It also caches it inside the provider itself so subsequent calls don't have to fetch the credentials again. 

Using the EKS client over OpenTofu has the benefit of a much faster retrieval as it is one API call vs re-initializaing OpenTofu and refreshing the state. Also, caching it for a given combination of project name and region makes it more efficient so all the subsequent calls in a deployment (e.g. installing Longhorn and 
the AWS Load Balancer Controller) have access to the credentials without having to make the same API calls again.

The caching strategy has been implemented taking into account that #266 will be merged and it opens up new ways of users interacting with deployments in a way where making the caching thread safe, being able to cache multiple clusters using the same provider instance, and invalidating the cache after destroy make sense.

## How to Test

1. Deploy to AWS and verify OpenTofu is not being initialized after deploying the infrastructure. All the Helm Chart installs happening afterwards are using the EKS client and the cached kubeconfig.
2. Run `nic kubeconfig` on the deployed cluster and verify OpenTofu is not being initialized.


pkg/provider/aws/eks.go
```diff
@@ -0,0 +1,46 @@
...
+		Name: &clusterName,
+	})
+	if err != nil {
+		var notFound *ekstypes.ResourceNotFoundException
+		if errors.As(err, &notFound) {
+			return nil, fmt.Errorf("cluster %q not found in region %q: run 'deploy' first", clusterName, region)
+		}
+		return nil, fmt.Errorf("failed to describe EKS cluster: %w", err)
+	}
+
```
viniciusdc, line 59: Why do you need to pass the region on to the method again? The EKSClient is built in the factory method based on the passed-in region, so it already contains that information, doesn't it? And it seems that this is only used here for logging... could reduce the required parameter and re-use the existing data from the client

pkg/provider/aws/eks.go
```diff
@@ -0,0 +1,46 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/eks"
+	ekstypes "github.com/aws/aws-sdk-go-v2/service/eks/types"
+)
+
+// EKSClient defines the EKS operations needed to fetch cluster connection details.
+type EKSClient interface {
```
dcmcand, line 16: Interface scoped to exactly the one method you need - mockable, no aws-sdk types leak into test code. Textbook idiomatic Go.

pkg/provider/aws/eks.go
```diff
@@ -0,0 +1,46 @@
...
+	DescribeCluster(ctx context.Context, params *eks.DescribeClusterInput, optFns ...func(*eks.Options)) (*eks.DescribeClusterOutput, error)
+}
+
+func newEKSClient(ctx context.Context, region string) (EKSClient, error) {
```
dcmcand, line 20: **Blocker:** new `pkg/` function missing OTel span. AGENTS.md "OpenTelemetry Instrumentation" requires every new function in `pkg/` to be span-wrapped, and `pkg/provider/aws` is not on the exemption list.

Fix:
```go
tracer := otel.Tracer("nebari-infrastructure-core")
ctx, span := tracer.Start(ctx, "aws.newEKSClient")
defer span.End()
span.SetAttributes(attribute.String("region", region))
```
And `span.RecordError(err)` on each error return.

pkg/provider/aws/eks.go
```diff
@@ -0,0 +1,46 @@
+package aws
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/config"
```
dcmcand: **Stylistic:** the rest of the package imports this aliased as `awsconfig` (see `provider.go:9`). Align the alias here for grep-ability.

pkg/provider/aws/eks.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load AWS config: %w", err)
+	}
+	return eks.NewFromConfig(cfg), nil
+}
+
+// fetchEKSKubeconfig calls DescribeCluster and assembles a kubeconfig from
+// the returned endpoint and certificate authority data. ResourceNotFound is
+// translated into a friendly "run 'deploy' first" error.
+func fetchEKSKubeconfig(ctx context.Context, client EKSClient, clusterName, region string) ([]byte, error) {
```
dcmcand, line 37: **Blocker:** same OTel issue as `newEKSClient`. `DescribeCluster` is exactly the kind of network call spans surface best — a far more interesting trace span than the now-elided tofu invocation it replaces. Add a span here with `cluster_name` and `region` attributes, plus `span.RecordError(err)` on the failure paths.

pkg/provider/aws/eks.go
```diff
@@ -0,0 +1,46 @@
...
+	out, err := client.DescribeCluster(ctx, &eks.DescribeClusterInput{
+		Name: &clusterName,
+	})
+	if err != nil {
+		var notFound *ekstypes.ResourceNotFoundException
+		if errors.As(err, &notFound) {
+			return nil, fmt.Errorf("cluster %q not found in region %q: run 'deploy' first", clusterName, region)
```
dcmcand: Translating `ResourceNotFoundException` into "run 'deploy' first" is exactly the kind of error contextualization a user-facing CLI needs.

pkg/provider/aws/eks.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+		return nil, fmt.Errorf("failed to describe EKS cluster: %w", err)
+	}
+
+	if out.Cluster == nil || out.Cluster.Endpoint == nil || out.Cluster.CertificateAuthority == nil || out.Cluster.CertificateAuthority.Data == nil {
```
dcmcand, line 60: Explicit nil-guards on every level (`out.Cluster`, `Endpoint`, `CertificateAuthority`, `Data`) before dereferencing. AWS SDK pointer-heavy types are easy to get wrong; this is careful.

pkg/provider/aws/provider.go
```diff
@@ -37,11 +38,24 @@ const (
 )
 
 // Provider implements the AWS provider
-type Provider struct{}
+type Provider struct {
+	kubeconfigMu    sync.Mutex
```
dcmcand: **Stylistic:** two small Go-idiom notes since you asked:

1. Consider `sync.RWMutex` so concurrent reads after warm-up don't serialize. Tiny win, but it's the idiomatic Go choice for read-heavy maps.
2. Worth a one-liner doc above the struct: `// Provider is not safe to copy once constructed (embeds a mutex). Always pass *Provider.` — `go vet`/copylocks would catch a violation but the comment makes the contract obvious to readers.

pkg/provider/aws/provider.go
```diff
@@ -546,6 +569,11 @@ func (p *Provider) Destroy(ctx context.Context, projectName string, clusterConfi
 		return err
 	}
 
+	// Drop any cached kubeconfig so a follow-up Deploy with the same project name and
```
dcmcand: **Question:** invalidation runs after `tf.Destroy` succeeds but before `destroyStateBucket`. If bucket teardown fails and the user retries, the cache could repopulate. Not a correctness bug — `DescribeCluster` returns `NotFound` for a destroyed cluster and your code translates it — just defensive.

Consider moving into a `defer` at the top of `Destroy` so panics and early returns also clear it:
```go
defer p.invalidateKubeconfigCache(projectName, region)
```

pkg/provider/aws/provider.go
```diff
@@ -569,105 +600,38 @@ func (p *Provider) GetKubeconfig(ctx context.Context, projectName string, cluste
```
dcmcand, line 604: **Stylistic:** you build `kubeconfigCacheKey{...}` here and again at line 628 (write path). Build once near the top and reuse — also makes adding more cache attrs later a one-place change.

pkg/provider/aws/provider.go
```diff
@@ -569,105 +600,38 @@ func (p *Provider) GetKubeconfig(ctx context.Context, projectName string, cluste
...
-			span.RecordError(err)
-			return nil, fmt.Errorf("failed to get state bucket name: %w", err)
-		}
-	}
+	cacheKey := kubeconfigCacheKey{projectName: projectName, region: region}
 
 	span.SetAttributes(
 		attribute.String("provider", ProviderName),
 		attribute.String("cluster_name", clusterName),
 		attribute.String(attrKeyRegion, region),
```
dcmcand, line 612: **Question:** small TOCTOU window — the lock is released between the cache-check and the AWS call, so two concurrent callers can both miss and both call `DescribeCluster` (same data, same key — just wasted work). Today's CLI is single-threaded per AGENTS.md "Thread Safety", so this is fine *now*. The PR body mentions #266 may change that.

If you want to future-proof, `golang.org/x/sync/singleflight.Group` keyed on `cacheKey` is the idiomatic Go pattern: parallel callers share one in-flight call. Worth doing now if you're confident #266 will land; otherwise note as a follow-up.

pkg/provider/aws/provider.go
```diff
@@ -569,105 +600,38 @@ func (p *Provider) GetKubeconfig(ctx context.Context, projectName string, cluste
...
-		return nil, err
-	}
-
-	// Initialize terraform to read outputs from state
-	tf, err := tofu.Setup(ctx, tofuTemplates, awsCfg.toTFVars(projectName))
-	if err != nil {
-		span.RecordError(err)
-		return nil, fmt.Errorf("failed to setup terraform: %w", err)
+	p.kubeconfigMu.Lock()
+	if cached, ok := p.kubeconfigCache[cacheKey]; ok {
```
dcmcand, line 622: `attribute.Bool("cache_hit", ...)` makes the cache observable in traces. Nice — operators debugging "why is this deploy slow" can immediately see whether the cache is helping.

pkg/provider/aws/eks_test.go
```diff
@@ -0,0 +1,177 @@
...
+	if !strings.Contains(err.Error(), "not ready") {
+		t.Fatalf("expected 'not ready' error, got: %v", err)
+	}
+}
+
+func TestGetKubeconfig_ReturnsCachedValueWithoutHittingAWS(t *testing.T) {
+	p := NewProvider()
+	cached := []byte("preloaded-kubeconfig-bytes")
+	key := kubeconfigCacheKey{projectName: "proj", region: "us-west-2"}
+	p.kubeconfigCache[key] = cached
```
dcmcand: **Question:** these tests poke at `p.kubeconfigCache` directly without holding `kubeconfigMu` (also at line 153 and 166-167). Functionally fine in serial tests today, but `go test -race` could flag if tests are ever parallelized via `t.Parallel()` or `-parallel`.

Cheap fix: a small unexported test helper that takes the lock, or exercise the cache only through `GetKubeconfig` / `invalidateKubeconfigCache`. Not blocking — just a future footgun.

dcmcand (changes requested):
Really nice piece of work — switching `GetKubeconfig` from "spin up terraform-exec + Init + Output" to a single `DescribeCluster` is the right answer to #241, and the `EKSClient` interface scoped to the one method you need is textbook idiomatic Go. Tests cover the happy path plus four error shapes, and you got the AWS SDK pointer nil-guards right on the first try.

### Direct answer to your cache-on-struct question

Yes — it's idiomatic Go for this codebase, and you picked correctly here. Three things to know:

1. **Why it fits.** `cmd/nic/main.go` registers a single `*aws.Provider` instance at startup and reuses it across the run. Your struct field's lifetime is exactly one CLI invocation — which is what you want for a "don't re-fetch within a deploy" cache.
2. **What you're committing to.** Every other `*Provider` in `pkg/provider/{gcp,azure,local,hetzner,existing}` is a `struct{}` with no state. This PR introduces the first stateful provider. Not a problem, but if someone later proposes parallelizing deploy commands, they need to know `*aws.Provider` is now safe-for-concurrent-use only because of the mutex you added (and would benefit from `singleflight` — see inline).
3. **Alternatives you (correctly) didn't pick.** A `context.Value`-based cache has no precedent here and forces callers through a wrapper. CLI-layer memoization in `cmd/nic` works but requires changing every fan-out call site. Provider-struct caching keeps the optimization invisible to callers and works for any code path through `GetKubeconfig`. Right tradeoff for this codebase.

### Resolution status

Closes #241 — both manual test steps in the description map directly to the issue's perf complaint.

### Request changes

The only blocker is the missing OTel spans on `newEKSClient` and `fetchEKSKubeconfig` — AGENTS.md "OpenTelemetry Instrumentation" requires every new `pkg/` function to be span-wrapped, and `pkg/provider/aws` is not on the exemption list. A `DescribeCluster` call is exactly the kind of network-bound op spans are designed to surface. See inline for fix.

The questions and stylistic notes are all non-blocking. Worth a doc PR after this lands codifying the "providers may hold deploy-scoped in-memory caches" pattern in AGENTS.md so the next provider that wants the same trick has a recipe.

pkg/provider/aws/eks.go
marcelovilla: `region` is not only being used for the error message but it's also being pass to the `buildKubeconfig` function: https://github.com/nebari-dev/nebari-infrastructure-core/blob/1c85692808895b57d7cbba7cfd6dddfaff023b78/pkg/provider/aws/eks.go#L45

While it's true we could derive it from the client, that would break the `EKSClient` interface we use to allow for mocking in the tests. Do you ahve any strong feelings against passing this as an argument in the `fetchEKSKubeconfig` function?

pkg/provider/aws/eks.go
marcelovilla: Addressed in b80734c9c56499a07216be23a0c98c8a64096818

pkg/provider/aws/eks.go
marcelovilla: Addressed in e51ebcc9f66b476ae1f20ea152deab537c907e29

pkg/provider/aws/eks.go
marcelovilla: Addressed in 261af8ba57aff4e4235e1f23ddb069b68199bcba

pkg/provider/aws/provider.go
marcelovilla: Addressed in fb3d8a98d1f18e9286ae2dc5a2445f2a7dba6ef3

pkg/provider/aws/provider.go
marcelovilla: Addressed in 335adf3280389da674d593ed4bc00056858a8959

pkg/provider/aws/provider.go
marcelovilla: I'm not sure what you mean. It's build only once here:

https://github.com/nebari-dev/nebari-infrastructure-core/blob/335adf3280389da674d593ed4bc00056858a8959/pkg/provider/aws/provider.go#L607

pkg/provider/aws/provider.go
marcelovilla: If I understand this correctly, it would only happen in the case of concurrent deployments of the same `projectName` + `region` combination, right? I don't think that would realistically happen as one wold not want to deploy/update the same cluster at the same time.

pkg/provider/aws/eks_test.go
marcelovilla: Addressed in 34aa38ab8b8f402875ca78d2de081a818f15f9c1

pkg/provider/aws/eks.go
viniciusdc: No strong feelings; I do prefer it to be explicit, to be honest. Just caught my attention right away, since the EKSClient already contains the object indirectly and could've been used instead. For the tests, I don't see why that would be a major blocker/breaking change, since you need to build the client with a mocked entry point anyway. 

After looking again at the response object from the API, the describe object does not contain the region, as I assumed: https://docs.aws.amazon.com/boto3/latest/reference/services/eks/client/describe_cluster.html. So keeping as is sounds like the best scenario.

pkg/provider/aws/eks.go
```diff
@@ -0,0 +1,72 @@
...
+
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/eks"
+	ekstypes "github.com/aws/aws-sdk-go-v2/service/eks/types"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+// EKSClient defines the EKS operations needed to fetch cluster connection details.
+type EKSClient interface {
```
dcmcand, line 16: EKSClient as a tight consumer-defined interface with just DescribeCluster on it is exactly the right shape. Same pattern as S3Client and STSClient, and it keeps the test mock trivial.

pkg/provider/aws/eks.go
```diff
@@ -0,0 +1,72 @@
...
+	span.SetAttributes(
+		attribute.String("cluster_name", clusterName),
+		attribute.String(attrKeyRegion, region),
+	)
+
+	out, err := client.DescribeCluster(ctx, &eks.DescribeClusterInput{
+		Name: &clusterName,
+	})
+	if err != nil {
+		var notFound *ekstypes.ResourceNotFoundException
```
dcmcand, line 50: Good call using `errors.As` for `ResourceNotFoundException` and translating it into an actionable 'run deploy first' message. Beats string matching.

pkg/provider/aws/eks.go
```diff
@@ -0,0 +1,72 @@
...
+		if errors.As(err, &notFound) {
+			err := fmt.Errorf("cluster %q not found in region %q: run 'deploy' first", clusterName, region)
+			span.RecordError(err)
+			return nil, err
+		}
+		span.RecordError(err)
+		return nil, fmt.Errorf("failed to describe EKS cluster: %w", err)
+	}
+
+	if out.Cluster == nil || out.Cluster.Endpoint == nil || out.Cluster.CertificateAuthority == nil || out.Cluster.CertificateAuthority.Data == nil {
```
dcmcand, line 60: **Stylistic.** The nil-guard chain protects against nil pointers, but `*Endpoint == ""` or empty CA `Data` would still pass through to `buildKubeconfig`. EKS doesn't return empty strings in practice for a CREATING/ACTIVE cluster, so this is defense-in-depth at best. If you'd rather rely on the AWS guarantee, leave it; if you want belt-and-suspenders, the 'not ready' branch could also reject empty values.

pkg/provider/aws/provider.go
```diff
@@ -419,6 +443,14 @@ func (p *Provider) Destroy(ctx context.Context, projectName string, clusterConfi
 
 	region := awsCfg.Region
 
+	// Drop any cached kubeconfig for this cluster at the end on every
+	// exit path (success, error, panic), so callers reusing the same
+	// Provider after a destroy don't get a stale entry The call is
+	// idempotent: a missing entry is a no-op.
+	if !opts.DryRun {
+		defer p.invalidateKubeconfigCache(projectName, region)
```
dcmcand, line 451: Putting the cache invalidation in a `defer` covers success, error, and panic paths in one line. The `!opts.DryRun` guard is right since dry-run never writes to the cache.

pkg/provider/aws/provider.go
```diff
@@ -73,6 +88,15 @@ func containsSubstring(slice []string, substr string) bool {
 	return false
 }
 
+// invalidateKubeconfigCache removes any cached kubeconfig for the given
+// project/region. Called on Destroy so a follow-up Deploy in the same process
+// doesn't return stale data for a cluster that has been torn down and recreated.
+func (p *Provider) invalidateKubeconfigCache(projectName, region string) {
```
dcmcand, line 94: **Question.** CLAUDE.md says new code in `pkg/` should be span-instrumented, and `pkg/provider/aws` isn't on the exemption list. This helper is trivially short (one map delete under a mutex) and it's reached via `defer` from the already-instrumented `Destroy`, so trace coverage isn't really lost. Up to you whether to add an `aws.invalidateKubeconfigCache` span for symmetry with `extractAWSConfig`, or call this an implicit trivial-helper exemption. If the latter, might be worth a one-line note in CLAUDE.md so future reviews don't flag it again.

pkg/provider/aws/provider.go
```diff
@@ -554,13 +586,16 @@ func (p *Provider) Destroy(ctx context.Context, projectName string, clusterConfi
...
 
-// GetKubeconfig generates a kubeconfig file for the EKS cluster
+// GetKubeconfig generates a kubeconfig file for the EKS cluster.
+//
+// Results are cached in-memory per Provider instance, indexed by projectName and
+// region, so repeated calls within a single command invocation (e.g. ArgoCD, Longhorn,
+// and AWS LBC install) reuse the same fetched value.
 func (p *Provider) GetKubeconfig(ctx context.Context, projectName string, clusterConfig *config.ClusterConfig) ([]byte, error) {
 	tracer := otel.Tracer("nebari-infrastructure-core")
 	_, span := tracer.Start(ctx, "aws.GetKubeconfig")
```
dcmcand: **Question.** `_, span := tracer.Start(ctx, ...)` discards the new context, so `newEKSClient` and `fetchEKSKubeconfig` end up as siblings of `aws.GetKubeconfig` in the trace tree rather than children. The same pattern is in `Validate`/`Deploy`/`Destroy`, so this isn't new, but the PR adds two more spans that won't parent correctly. Want to switch to `ctx, span := ...` here and clean up the other methods in a follow-up, or punt the whole thing?

pkg/provider/aws/provider.go
```diff
@@ -569,105 +604,38 @@ func (p *Provider) GetKubeconfig(ctx context.Context, projectName string, cluste
...
-		span.RecordError(err)
-		return nil, err
-	}
-
-	// Initialize terraform to read outputs from state
-	tf, err := tofu.Setup(ctx, tofuTemplates, awsCfg.toTFVars(projectName))
-	if err != nil {
-		span.RecordError(err)
-		return nil, fmt.Errorf("failed to setup terraform: %w", err)
+	p.kubeconfigMu.RLock()
```
dcmcand, line 621: **Question.** Read-then-write under RWMutex is race-free, but two concurrent `GetKubeconfig` callers can both miss, both call `DescribeCluster`, and both write. Today the call sites are sequential so this is theoretical, but the PR description points at #266 opening up concurrent usage. Want to pull in `golang.org/x/sync/singleflight` keyed on `cacheKey` so a miss collapses to one upstream call, or are you fine accepting the duplicate fetch and documenting it?

pkg/provider/aws/eks_test.go
```diff
@@ -0,0 +1,192 @@
...
+	if m.DescribeClusterFunc != nil {
+		return m.DescribeClusterFunc(ctx, params, optFns...)
+	}
+	return &eks.DescribeClusterOutput{}, nil
+}
+
+// validCAData is a base64-encoded blob accepted by buildKubeconfig.
+var validCAData = base64.StdEncoding.EncodeToString([]byte("dummy-ca-bytes"))
+
+func successOutput() *eks.DescribeClusterOutput {
```
dcmcand, line 32: **Stylistic.** `successOutput` works fine as a single fixed blob. If more variations show up later, table-driving these would match the project convention (CLAUDE.md prefers table-driven tests for Go).

dcmcand (dismissed):
Approving. The EKS-client + per-Provider cache approach is the right shape for what #241 was asking for, and the unit tests cover the meaningful failure modes on the helper.

I left three questions on the span context, the trivial-helper instrumentation, and the concurrent-call behavior, plus a couple of small stylistic notes. The stylistic items are take-it-or-leave-it: address them in this PR, file follow-ups, or push back if you disagree. None of it blocks merge.

Nice work on closing out the slow GetKubeconfig path. That was real friction during deploys.

pkg/provider/aws/eks.go
marcelovilla: Thanks for raising this. I'm leaving it as is because (1) there's the AWS guarantee, and (2) if that is ever broken, this would still fail loudly downstream when `buildKubeconfig` is called.

pkg/provider/aws/provider.go
marcelovilla: Instrumenting this function means I'd need to pass `ctx` to it. I'm leaning towards leaving it like this but I'm wondering if you have any strong feelings about instrumenting everything, including functions that do not take a `ctx`.

pkg/provider/aws/provider.go
marcelovilla: Good catch, thanks. Addressed for this function in 037126421802ec42c2c156303151605f5c5da21b and opened https://github.com/nebari-dev/nebari-infrastructure-core/issues/329 as a follow up for the other methods.

pkg/provider/aws/provider.go
marcelovilla: I added a comment justifying leaving it as is in 9252ecc8b883549d9230f0887d1ecc8ba87a30a5

pkg/provider/aws/eks_test.go
marcelovilla: I agree. If more variations show up in the future, we should update these to be table-driven tests.

dcmcand (approved):
lgtm :rocket:

PR #324, opened 2026-05-21
feat(azure): implement Azure provider end-to-end via terraform-azurerm-aks-cluster

tylerpotts:
## Summary

Closes #327 

Brings the Azure provider to first-class status, on par with the AWS path. `nic deploy --config examples/azure-config.yaml` now provisions an AKS cluster end-to-end against a real subscription.

- New companion repo: https://github.com/nebari-dev/terraform-azurerm-aks-cluster (Track A, separate work). This PR pins the NIC shim to a git ref of that repo's `main` branch; will flip to the Terraform Registry version (`nebari-dev/aks-cluster/azurerm`) once Track A tags `v0.1.0`.
- `pkg/provider/azure/` is fully implemented: Config schema with validation, `(*Config).toTFVars()` for OpenTofu marshalling, embedded TF shim (5 files), `Deploy`/`Destroy`/`GetKubeconfig`/`Summary`/`InfraSettings`. Kubeconfig is fetched via the Azure SDK (`armcontainerservice`), not Terraform state.
- **State persistence via azurerm backend.** Bootstrapped Storage Account (deterministic name `nictfstate<14-hex>` from SHA-256 of subscription ID) + container, blob versioning enabled. Shared across clusters in a subscription, keyed by `<project_name>.tfstate`. Diverges from AWS (per-cluster S3 buckets) due to Azure's 24-char SA name limit + global namespace.
- `pkg/tofu` gained a small `SetExtraEnv` helper so `ARM_SUBSCRIPTION_ID` reaches the tofu child without mutating the parent process env. AWS provider unaffected.
- Tag-based resource discovery (`state.go`) + orphan reporting on destroy (`cleanup.go`). Auto-delete of orphans deferred to a follow-up.

## Design + plan artifacts (committed in this PR)

- `docs/superpowers/specs/2026-05-21-azure-aks-terraform-module-and-nic-provider-design.md` — full design spec
- `docs/superpowers/plans/2026-05-21-terraform-azurerm-aks-cluster.md` — Track A plan (26 tasks)
- `docs/superpowers/plans/2026-05-21-nic-azure-provider.md` — Track B plan (22 tasks)

## Test plan

- [x] CI green (`make check`: fmt/vet/lint/race-tested unit tests)
- [x] Manual: Configure Azure AD app w/ OIDC federation, set `AZURE_CLIENT_ID`/`AZURE_TENANT_ID`/`AZURE_SUBSCRIPTION_ID` secrets on the terraform repo
- [x] Manual: Tag Track A `v0.1.0`, verify Opentofu Registry auto-publishes at `nebari-dev/aks-cluster/azurerm`
- [x] Follow-up commit on this branch: flip `pkg/provider/azure/templates/main.tf` `source` from git-ref to `nebari-dev/aks-cluster/azurerm` + `version = "0.1.0"` (Track B Task 20)
- [x] Manual: `nic deploy --config examples/azure-config.yaml` against a real subscription end-to-end
- [x] Manual: `nic kubeconfig` writes a working kubeconfig; `kubectl get nodes` shows Ready
- [ ] Manual: `nic destroy` cleans up cluster + state SA stays (intentional — see commit `d3b0191`)

## Known follow-ups (deferred from MVP scope)

- Azure Files / `azurefile-csi` shared storage (EFS analog)
- Workload Identity beyond the kubelet identity
- Application Gateway Ingress Controller (ALB Controller analog)
- Key Vault encryption-at-rest
- Log Analytics Workspace integration
- Longhorn (intentionally skipped — AKS konnectivity tunnel makes EKS-style SG rules N/A)
- `DeployOptions.DryRun` / `DestroyOptions.Force` plumbing
- BYO networking validation (`existing_*` ids non-null when `create_* = false`)
- Auto-delete of orphan resources on destroy


pkg/provider/azure/provider.go
```diff
@@ -2,137 +2,310 @@ package azure
...
 		WithAction("validate").
 		WithMetadata("cluster_name", projectName))
 	return nil
 }
 
-// Deploy deploys Azure infrastructure (stub implementation)
+// Deploy provisions the Azure AKS cluster by invoking the embedded OpenTofu
+// shim. It writes terraform.tfvars.json from the parsed Config, runs
+// `tofu init` and `tofu apply`, and streams output through the status channel.
 func (p *Provider) Deploy(ctx context.Context, projectName string, clusterConfig *config.ClusterConfig, _ provider.DeployOptions) error {
```
marcelovilla: You're not using the `provider.DeployOptions`, which means running `nic` with the `--dry-run` flag is actually creating infrastructure.

Note that a `--dry-run` on a cluster that has not been deployed yet should use a local state to avoid creating a bucket. If `--dry-run` is used on an existing cluster, it should use the previously created remote state. This mimics the pattern we implemented in AWS.

pkg/provider/azure/provider.go
```diff
@@ -2,137 +2,310 @@ package azure
...
 	return nil
 }
 
-// Destroy tears down Azure infrastructure (stub implementation)
-func (p *Provider) Destroy(ctx context.Context, projectName string, _ *config.ClusterConfig, _ provider.DestroyOptions) error {
+// Destroy tears down the Azure AKS cluster by running `tofu destroy` against
+// the same state backend used by Deploy. After tofu completes, cleanupOrphans
+// reports any tagged resources tofu missed (e.g., AKS-managed MC_* siblings)
+// as a non-fatal warning so users can clean them up manually.
+func (p *Provider) Destroy(ctx context.Context, projectName string, clusterConfig *config.ClusterConfig, _ provider.DestroyOptions) error {
```
marcelovilla: You're not using the `provider.DeployOptions`, which means running nic with the --dry-run flag is actually destroying infrastructure.

marcelovilla (changes requested):
Thanks for this PR @tylerpotts! I just did a first pass trying several things, including a full deploy, updates on an existing one, and a full destroy. Things seem to be working fine with the exception of the `--dry-run` flag inline comments I left.

I'll need to to a second pass at it but I figured I would leave this first review here so you have the chance to implement the fixes in the meantime.

mtsokol:
Hi @tylerpotts,
Thank you for working on this! I already started using it in my work.
Is it possible to also include:

- NIC's tofu module hardcodes the dataplane to Azure's default (no `network_dataplane` field exposed in its Network config). Would it be possible to make it configurable so `network-dataplane cilium` is possible in its tofu module? Otherwise migrating network dataplane to cilium after deploy rolls all nodes ~5-15 min.
- Would it be possible to have NAP configurable as well? I need `--node-provisioning-mode Auto` which requires `cilium` in the first place. So running NAP requires a separate few commands after `nic deploy`.

pkg/provider/azure/provider.go
tylerpotts: Fixed in 4b1179e. `Deploy` now honors `opts.DryRun`: it runs `tofu plan` and returns without applying. To avoid creating a bucket-equivalent on a first-time dry run, it checks whether the state Storage Account already exists (`stateBackendExists`) — if not, it overrides the azurerm backend with a local backend (`WriteBackendOverride`) so no cloud state is bootstrapped; if the cluster already exists, it inits against the existing remote state. This mirrors the AWS local-vs-S3 pattern.

pkg/provider/azure/provider.go
tylerpotts: Fixed in 4b1179e. `Destroy` now honors `opts.DryRun`: it runs `tofu plan -destroy` and returns before any real `tofu destroy` or orphan sweep. Same backend handling as Deploy — if the state backend does not exist there is nothing to tear down and it falls back to a local backend so the plan still runs; otherwise it reads the existing remote state.

tylerpotts:
Hi @mtsokol — both are now implemented end to end.

**Module (Track A):** https://github.com/nebari-dev/terraform-azurerm-aks-cluster/pull/2 adds:
- `network_data_plane` wired into the cluster `network_profile` — set to `cilium` for Azure CNI Powered by Cilium (requires `network_plugin=azure` + overlay, which NIC already uses).
- `node_provisioning_mode` for NAP. Since the stable `azurerm` provider does not yet expose `nodeProvisioningProfile` (hashicorp/terraform-provider-azurerm#31418), the module PATCHes it on via the `azapi` provider after cluster creation. Validation enforces that `Auto` requires the cilium dataplane.

**NIC config:** in this PR you can now set them under `cluster.azure`:

```yaml
cluster:
  azure:
    network:
      dataplane: cilium
    node_provisioning_mode: Auto   # requires dataplane: cilium
```

`dataplane` defaults to `azure` and `node_provisioning_mode` to `Manual`, so existing configs are unaffected. Setting `dataplane: cilium` at create time avoids the node roll you hit when migrating after deploy.

These ride on the module being released as `v0.1.1` (the shim now pins `0.1.1`); that release is pending merge of the Track A PR above which I will finish up tomorrow.

pkg/provider/azure/INTEGRATION_TESTING.md
marcelovilla, line 1: As far as I can tell, there are no integration tests implemented in Azure so far. I would drop this file until we implement them. Otherwise, it's misleading.

WALKTHROUGH.md
```diff
@@ -226,6 +226,8 @@ Focus tests on:
 5. Example config in `examples/`
 6. Tests
 
+See `pkg/provider/aws/` for a native-SDK reference implementation, and `pkg/provider/azure/` for a Terraform-module-driven reference (OpenTofu wrapping an external module, with Azure SDK used only for kubeconfig retrieval and tag-based cleanup).
```
marcelovilla: This is not accurate: `pkg/provider/aws/` does not provide a native-SDK reference implementation.

WALKTHROUGH.md
```diff
@@ -9,7 +9,7 @@ A Go CLI for declarative cloud infrastructure management using native cloud SDKs
 - **No state files** — queries cloud APIs on every run for actual state
 - **Tag-based discovery** — all resources tagged `nic.nebari.dev/cluster-name` and `nic.nebari.dev/managed-by=nic`
 - **Reconciliation loop** — discover → diff → create/update/delete
-- **Native SDKs** — aws-sdk-go-v2 (GCP, Azure pending)
+- **Native SDKs / Terraform** — aws-sdk-go-v2 for AWS; Azure uses OpenTofu against the Track A module `nebari-dev/terraform-azurerm-aks-cluster` plus the Azure SDK (`armcontainerservice`) for kubeconfig retrieval. GCP pending.
```
marcelovilla: This is not accurate: AWS does not rely only on the SDKs anymore but also uses OpenTofu/Terraform. Also, I would ditch the _Track A_ wording here—it seems to be a leftover from the way you separated work but it shouldn't be user-facing.

pkg/provider/azure/templates/outputs.tf
```diff
@@ -0,0 +1,57 @@
...
+output "cluster_fqdn" {
+  value = module.aks_cluster.cluster_fqdn
+}
+
+output "host" {
+  value     = module.aks_cluster.host
+  sensitive = true
+}
+
+# The Track A module exposes kube_config_raw (local cluster-admin), not
```
marcelovilla: I find the term _Track A_ confusing. It seems to come from the way you separated the work for implementing the Azure provider—create the terraform module (Track A) vs wiring it here (Track B). I don't think that should be in the codebase as it's not clear to a reader where it's coming from.

pkg/provider/azure/tofu.go
```diff
@@ -0,0 +1,175 @@
+package azure
+
+import "embed"
+
+// tofuTemplates contains the OpenTofu shim that calls the Track A module.
```
marcelovilla: Same comment as https://github.com/nebari-dev/nebari-infrastructure-core/pull/324/changes#r3322716233 regarding the usage of _Track A_.

ARCHITECTURE.md
```diff
@@ -173,16 +173,33 @@ func (p *Provider) someFunction(ctx context.Context, clients *Clients, cfg *conf
 | `provider.go` | Stub provider that prints operations |
 | `config.go` | GCP-specific config types: `Config`, `NodeGroup`, `Taint`, `GuestAccelerator` |
 
-### Azure Provider (pkg/provider/azure/) - **Stub**
+### Azure Provider (pkg/provider/azure/) - **Fully Implemented**
 
 **Location:** `pkg/provider/azure/`
 
-**Status:** Stub implementation - prints config as JSON
+**Status:** Complete implementation that drives the Track A Terraform module `nebari-dev/terraform-azurerm-aks-cluster` via OpenTofu. Implements all `Provider` methods: `Validate`, `Deploy`, `Destroy`, `GetKubeconfig`, `Summary`, and `InfraSettings`.
```
marcelovilla: Same comment as https://github.com/nebari-dev/nebari-infrastructure-core/pull/324/changes#r3322716233 regarding the usage of _Track A_.

ARCHITECTURE.md
```diff
@@ -173,16 +173,33 @@ func (p *Provider) someFunction(ctx context.Context, clients *Clients, cfg *conf
...
+
+**Module sourcing:** Currently pinned to the module via a git ref; flips to the Terraform Registry once Track A tags `v0.1.0`.
```
marcelovilla: Same comment as https://github.com/nebari-dev/nebari-infrastructure-core/pull/324/changes#r3322716233 regarding the usage of _Track A_.

Furthermore, the module already is tagged as v0.1.0 so this is obsolete.

ARCHITECTURE.md
```diff
@@ -173,16 +173,33 @@ func (p *Provider) someFunction(ctx context.Context, clients *Clients, cfg *conf
...
+**Kubeconfig:** Fetched via the Azure SDK (`armcontainerservice`) rather than read from Terraform state, so it works even when the local tfstate is absent.
 
 | File | Purpose |
 |------|---------|
-| `provider.go` | Stub provider that prints operations |
+| `provider.go` | Provider implementation: orchestrates Validate/Deploy/Destroy/Summary/InfraSettings over OpenTofu |
 | `config.go` | Azure-specific config types: `Config`, `NodeGroup`, `Taint` |
+| `state.go` | Tag-based discovery and orphan cleanup helpers |
+| `state_backend.go` | Bootstraps the `azurerm` Terraform backend (resource group, storage account, container) |
+| `tofu.go` | OpenTofu invocation: init/plan/apply/destroy against the Track A module |
```
marcelovilla: Same comment as https://github.com/nebari-dev/nebari-infrastructure-core/pull/324/changes#r3322716233 regarding the usage of _Track A_.

ARCHITECTURE.md
```diff
@@ -173,16 +173,33 @@ func (p *Provider) someFunction(ctx context.Context, clients *Clients, cfg *conf
...
+| `kubeconfig.go` | Retrieves cluster admin kubeconfig via `armcontainerservice` |
+| `cleanup.go` | Post-destroy resource sweep (tag-based) |
+| `interfaces.go` | Azure SDK client interfaces for mocking |
+| `version.go` | Module version pin |
+| `templates/` | Rendered Terraform root module that wraps the Track A module |
```
marcelovilla: Same comment as https://github.com/nebari-dev/nebari-infrastructure-core/pull/324/changes#r3322716233 regarding the usage of _Track A_.

pkg/provider/azure/templates/main.tf
marcelovilla, line 1: You're missing a `.terraform.lock.hcl` pinning the dependencies for all different platforms. While the `nebari-dev/aks-cluster/azurerm` itself is pinned here, there's no guarantee other dependencies, such as the azure provider, won't drift between different runs of the same version of `nic`.

You have to run the following commands to generate the lockfile:

```bash
tofu init -backend=false

tofu providers lock \
  -platform=linux_amd64 \
  -platform=linux_arm64 \
  -platform=darwin_amd64 \
  -platform=darwin_arm64 \
  -platform=windows_amd64
```

pkg/provider/azure/templates/outputs.tf
marcelovilla, line 1: Non blocking and feel free to ignore this, but outputs have descriptions on the AWS provider but don't have them here. Just flagging the asymmetry.

pkg/provider/azure/cleanup.go
```diff
@@ -0,0 +1,56 @@
...
+	ids, err := listTaggedResources(ctx, client, projectName)
+	if err != nil {
+		return err
+	}
+	if len(ids) == 0 {
+		return nil // happy path: tofu destroy was thorough.
+	}
+	mc, others := classifyMC(ids)
+	return fmt.Errorf("found %d orphaned resources after destroy (MC: %d, other: %d); run `az resource delete --ids %s` to clean up",
+		len(ids), len(mc), len(others), strings.Join(ids, " "))
```
marcelovilla, line 55: This is not really cleaning up any resources but rather just returning an error telling the user to run `az resource delete`. Is that the expected behavior? Seems inconsistent with the function name.

pkg/provider/azure/interfaces.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// managedClusterVersionsAPI exposes the AKS-supported-versions lookup.
+type managedClusterVersionsAPI interface {
+	ListKubernetesVersions(
+		ctx context.Context,
+		location string,
+		options *armcontainerservice.ManagedClustersClientListKubernetesVersionsOptions,
+	) (armcontainerservice.ManagedClustersClientListKubernetesVersionsResponse, error)
+}
```
marcelovilla: This is dead production code that's only used in this function: https://github.com/nebari-dev/nebari-infrastructure-core/blob/4b1179ec1cb4a85182b06bb582a14d90b6350d51/pkg/provider/azure/version.go#L8-L20


which is only called in the tests:

https://github.com/nebari-dev/nebari-infrastructure-core/blob/4b1179ec1cb4a85182b06bb582a14d90b6350d51/pkg/provider/azure/version_test.go#L25-L37

pkg/provider/azure/version.go
```diff
@@ -0,0 +1,20 @@
...
+		return nil, fmt.Errorf("list AKS versions in %s: %w", location, err)
+	}
+	out := make([]string, 0, len(resp.Values))
+	for _, v := range resp.Values {
+		if v != nil && v.Version != nil {
+			out = append(out, *v.Version)
+		}
+	}
+	return out, nil
+}
```
marcelovilla: This is dead production code that only gets called in the tests:

https://github.com/nebari-dev/nebari-infrastructure-core/blob/4b1179ec1cb4a85182b06bb582a14d90b6350d51/pkg/provider/azure/version_test.go#L25-L37

pkg/provider/azure/interfaces.go
```diff
@@ -0,0 +1,42 @@
...
+		ctx context.Context,
+		resourceGroupName, resourceName string,
+		options *armcontainerservice.ManagedClustersClientListClusterAdminCredentialsOptions,
+	) (armcontainerservice.ManagedClustersClientListClusterAdminCredentialsResponse, error)
+
+	Get(
+		ctx context.Context,
+		resourceGroupName, resourceName string,
+		options *armcontainerservice.ManagedClustersClientGetOptions,
+	) (armcontainerservice.ManagedClustersClientGetResponse, error)
```
marcelovilla: This is dead production code as far as I can tell.

pkg/provider/azure/kubeconfig_test.go
```diff
@@ -0,0 +1,60 @@
...
+		},
+	}, nil
+}
+
+func (f *fakeManagedClusters) Get(_ context.Context, _, _ string, _ *armcontainerservice.ManagedClustersClientGetOptions) (armcontainerservice.ManagedClustersClientGetResponse, error) {
+	if f.getResult == nil {
+		return armcontainerservice.ManagedClustersClientGetResponse{}, nil
+	}
+	return armcontainerservice.ManagedClustersClientGetResponse{ManagedCluster: *f.getResult}, nil
+}
```
marcelovilla: This is dead code as far as I can tell.

pkg/provider/azure/tofu.go
```diff
@@ -0,0 +1,175 @@
...
+	return out
+}
+
+func mergeTags(user map[string]string, projectName string) map[string]string {
+	out := make(map[string]string, len(user)+2)
+	for k, v := range user {
+		out[k] = v
+	}
+	out[tagClusterName] = projectName
+	out[tagManagedBy] = "nic"
```
marcelovilla: Just a small stylistic comment: `nic` is being hardcoded here as well as in

https://github.com/nebari-dev/nebari-infrastructure-core/blob/4b1179ec1cb4a85182b06bb582a14d90b6350d51/pkg/provider/azure/state_backend.go#L159

and in 

https://github.com/nebari-dev/nebari-infrastructure-core/blob/4b1179ec1cb4a85182b06bb582a14d90b6350d51/pkg/provider/azure/state_backend.go#L197.

This could be an additional constant here:

https://github.com/nebari-dev/nebari-infrastructure-core/blob/4b1179ec1cb4a85182b06bb582a14d90b6350d51/pkg/provider/azure/tofu.go#L85-L88

marcelovilla (changes requested):
@tylerpotts just did a second pass here

pkg/provider/azure/INTEGRATION_TESTING.md
tylerpotts: Dropped `INTEGRATION_TESTING.md` in 28b0f58 — agreed it was misleading with no integration tests implemented yet.

WALKTHROUGH.md
tylerpotts: Reworded in 28b0f58: now describes `pkg/provider/aws/` and `pkg/provider/azure/` as reference implementations that both combine cloud SDKs with OpenTofu, rather than calling AWS a native-SDK implementation.

WALKTHROUGH.md
tylerpotts: Fixed in 28b0f58: the bullet now states both providers combine cloud SDKs with OpenTofu/Terraform, and the "Track A" wording is gone.

pkg/provider/azure/templates/outputs.tf
tylerpotts: Removed the "Track A" wording in 28b0f58 (comment now says "the aks_cluster module").

pkg/provider/azure/tofu.go
tylerpotts: Removed "Track A" here in 28b0f58 (now "the external AKS module").

ARCHITECTURE.md
tylerpotts: Removed "Track A" in 28b0f58.

ARCHITECTURE.md
tylerpotts: Fixed in 28b0f58: dropped "Track A" and replaced the obsolete git-ref/v0.1.0 note with "Consumes the published `nebari-dev/aks-cluster/azurerm` module from the OpenTofu Registry, pinned by version in templates/main.tf."

ARCHITECTURE.md
tylerpotts: Removed "Track A" in 28b0f58.

ARCHITECTURE.md
tylerpotts: Removed "Track A" in 28b0f58.

pkg/provider/azure/templates/main.tf
tylerpotts: Added `pkg/provider/azure/templates/.terraform.lock.hcl` in 28b0f58, generated exactly as you described (`tofu init -backend=false` then `tofu providers lock` for linux/darwin/windows amd64+arm64). It pins azurerm 4.74.0 and azapi 2.10.0 with checksums for all five platforms. The lockfile is embedded via `//go:embed all:templates` and `tofu init` runs without `-upgrade`, so it's honored at deploy time.

pkg/provider/azure/templates/outputs.tf
tylerpotts: Added descriptions to all outputs in 28b0f58, matching the wording used in the module's own outputs for consistency.

pkg/provider/azure/cleanup.go
tylerpotts: Good catch — the name overpromised. Renamed `cleanupOrphans`/`cleanupOrphansWithClient` to `reportOrphans`/`reportOrphansWithClient` in 28b0f58 to match the report-only behavior. Auto-deletion is still intentionally deferred (it returns a non-fatal warning listing the orphaned IDs and the `az resource delete` command); the rename makes that explicit.

pkg/provider/azure/interfaces.go
tylerpotts: Removed in 28b0f58: deleted the `managedClusterVersionsAPI` interface along with `version.go` and its test.

pkg/provider/azure/version.go
tylerpotts: Deleted `version.go` and `version_test.go` in 28b0f58.

pkg/provider/azure/interfaces.go
tylerpotts: Removed the unused `Get` method from the `managedClustersAPI` interface in 28b0f58 (only `ListClusterAdminCredentials` is actually called in production).

pkg/provider/azure/kubeconfig_test.go
tylerpotts: Removed the dead `Get` fake and its `getResult` field from `fakeManagedClusters` in 28b0f58.

pkg/provider/azure/tofu.go
tylerpotts: Extracted into a `managedByValue = "nic"` constant in tofu.go (next to the tag-name constants) and used it at tofu.go and both state_backend.go sites in 28b0f58.

marcelovilla (approved):
@tylerpotts thanks for addressing all the comments and for the PR! It's very exciting to have support for Azure now :rocket:

mtsokol:
Thank you both for working on this! 🎉

PR #328, opened 2026-05-22
feat(longhorn): expose UI through Envoy Gateway with Keycloak OIDC

### LINKED ISSUE #268: [SPIKE] - Backup Solution for Nebari Components
## Description

For Nebari ecosystem products to be used in critical production workflows, we need a end-to-end backup and restore solution that guarantees that we can recover user data up to a point in time ("RPO") fairly quickly ("RTO").

Because of the distributed, modular nature of the system we do not expect a "unified" backup/restore option, but a "shared responsibility" between the Nebari Core and installed software packs with some of the following principles.

- Nebari core layer protects minimal data (Keycloak DB, certificates, app configurations, etc.)
- Each software pack is responsible for the workflow for backup/restore of its own data
- (Possibly) the core architecture defines an interface for software packs to implement whereby the core can orchestrate backup/restore of individual components

The purpose of this story is to:
1) Define the architecture for the backup/restore ecosystem for Nebari, adhering to compliance best practices such as those described below
2) Create downstream implementation issues for the core components (NIC, operator, etc.)
3) Create additional issues for building the software pack-level backup/restore features for the highest priority software packs - 'data science' and 'nebi'

## Use Case

This is needed for compliance and security approvals in production environments.

## Reference

These are a few of the NIST 800-53 (Rev 5) controls that are relevant.

CP-9: System Backup: This is the primary control for backups. It mandates that organizations conduct backups of user-level information, system-level information, and security documentation. Crucially, Revision 5 places a heavy emphasis on protecting the confidentiality, integrity, and availability of these backups (e.g., protecting them against ransomware).

CP-10: System Recovery and Reconstitution: This control dictates how a system is restored to its original, known-working state after a disruption or failure. It covers the actual processes and mechanisms used for recovery.

CP-6: Alternate Storage Site: Requires organizations to establish an alternate storage site (including agreements to ensure access) to house system backups, ensuring the site is geographically separate from the primary site to prevent a single event from destroying both.

CP-4: Contingency Plan Testing: Mandates that backup and recovery mechanisms aren't just theoretical; they must be regularly tested to ensure systems can actually be restored within the required timeframe.

CP-2: Contingency Plan: Requires the creation of the overarching plan that defines the Recovery Point Objectives (RPO) and Recovery Time Objectives (RTO) that your backup and restore operations must meet.

tylerpotts:
## Summary

Exposes the Longhorn UI at `https://longhorn.<domain>` through the existing Envoy Gateway, gated by Keycloak OIDC via an Envoy Gateway `SecurityPolicy`, and restricted to members of the `longhorn-admins` Keycloak group. Routes / policies / cert SAN / Keycloak client are only provisioned when the provider has Longhorn enabled (AWS / Hetzner default-on; Azure / GCP / Local off).

- **Per-provider gate**: new `provider.InfraSettings.LonghornEnabled` populated from existing per-provider `LonghornEnabled()` config methods. Threaded through `TemplateData` to gate manifest rendering.
- **Dual-namespace OIDC secret**: foundational install writes `longhorn-oidc-client-secret` into both `keycloak` (consumed by realm-setup-job to register the OIDC client) and `longhorn-system` (consumed by the SecurityPolicy). No `ReferenceGrant` needed.
- **New manifests** (all conditionally rendered on `LonghornEnabled`):
  - `manifests/networking/routes/longhorn-httproute.yaml` — HTTPRoute `longhorn` in `longhorn-system` → `longhorn-frontend:8080`, attaches to `nebari-gateway/https`
  - `manifests/networking/policies/longhorn-securitypolicy.yaml` — `SecurityPolicy longhorn-oidc` enforcing Keycloak OIDC (clientID `longhorn`, redirect `https://longhorn.<domain>/oauth2/callback`), plus a JWT provider and an `authorization` rule that Allows only when the `groups` claim contains `longhorn-admins` (everything else 403)
  - New ArgoCD `Application securitypolicies` so the new `policies/` directory is reconciled
- **Cert SAN**: `gateway-certificate.yaml` adds `longhorn.<domain>` to `dnsNames` when enabled.
- **Keycloak realm setup**: `realm-setup-job.yaml` registers the `longhorn` OIDC client, creates `longhorn-admins` / `longhorn-viewers` groups, attaches `groups` scope, and adds the realm admin to `longhorn-admins`.

Authorization gates the UI to members of the **`longhorn-admins` Keycloak group**. Non-members get HTTP 403 from the gateway. The `longhorn-viewers` group is created but currently a no-op (Longhorn has no read-only mode).

## Test Plan

Unit tests added (all passing under `make check` + `go test -race ./...`):

- [x] `TestProvider_InfraSettings_LonghornEnabled` on AWS and Hetzner (default + explicit-disabled)
- [x] `TestCreateLonghornSecrets` — dual-namespace Secret creation; no-op when `ClientSecret` empty
- [x] `TestCreateKeycloakAndLonghornSecrets_Together` — coexistence with the existing ArgoCD secret
- [x] `TestWriteAllToGit_LonghornHTTPRoute` — positive + negative render
- [x] `TestWriteAllToGit_LonghornSecurityPolicy` — positive + negative render; positive asserts forwardAccessToken, jwt provider, authorization Deny-by-default + Allow-longhorn-admins rule
- [x] `TestWriteAllToGit_GatewayCertIncludesLonghorn` — SAN added/omitted
- [x] `TestWriteAllToGit_RealmSetupRegistersLonghornClient` — kcadm block + env var renders/omits

Operator-driven acceptance (NOT automated; run after deploy):

- [x] Deploy with `domain: <real-domain>` + Longhorn enabled (default for AWS): `./nic deploy --config ...`
- [x] `kubectl -n envoy-gateway-system get certificate nebari-gateway-cert -o jsonpath='{.spec.dnsNames}'` includes `longhorn.<domain>`
- [x] `kubectl -n longhorn-system get httproute longhorn` shows `Accepted=True`
- [x] `kubectl -n longhorn-system get securitypolicy longhorn-oidc` shows `Accepted=True`
- [x] `kubectl -n keycloak logs job/keycloak-realm-setup | grep -i longhorn` shows the new client + groups created
- [x] Sign in as the realm admin user (already in `longhorn-admins`) → land on the Longhorn UI
- [x] Create a fresh Keycloak user, do NOT add to `longhorn-admins`, sign in → expect HTTP 403 from the gateway
- [x] Add that user to `longhorn-admins` via the Keycloak UI, sign out and back in → land on the Longhorn UI
- [x] Set `aws.longhorn.enabled: false`, redeploy → HTTPRoute / SecurityPolicy / Keycloak client / cert SAN all absent

## Follow-ups (out of scope)

- Client-secret rotation on redeploy (same pre-existing divergence risk as today's ArgoCD client)


viniciusdc:
I am looking at this @tylerpotts, thanks

pkg/argocd/templates/apps/securitypolicies.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: argoproj.io/v1alpha1
```
dcmcand, line 1: **Blocker (see summary):** this Application has no `LonghornEnabled` gate, but the only manifest under its target `policies/` directory (`longhorn-securitypolicy.yaml`) renders empty when Longhorn is disabled. Combined with `allowEmpty: false` below, that means a Longhorn-disabled cluster gets an Application targeting zero resources, and a Longhorn-enabled->disabled toggle leaves the SecurityPolicy un-prunable. Mirror the MetalLB skip pattern in `writer.go` (`isMetalLBPath`): skip writing both this app file and the policies manifest when `!settings.LonghornEnabled`.

dcmcand (dismissed):
Thanks for this - the architecture is sound and the enabled path is solid, well-tested work. Requesting changes on one Blocker around the disabled/teardown path.

## Highlights

- **Provider consistency is excellent.** AWS/Hetzner track `Config.LonghornEnabled()` (default-on); Azure/GCP/local hardcode `false` with clear `not yet wired` comments. The default-then-refine pattern matches the existing `StorageClass` handling exactly, and `LonghornEnabled` flows through `provider.InfraSettings` -> `TemplateData` with no provider-name switches anywhere. Textbook use of the abstraction boundary.
- **Dual-namespace secret handling is clean** (`createLonghornSecrets`): writes the OIDC client secret to both `keycloak` (for realm-setup-job) and `longhorn-system` (for the SecurityPolicy), with a comment explaining why, and no-ops on empty.
- **`generateSecurePassword` is reused, not reinvented** - crypto `rand.Reader`, generated only when `LonghornEnabled`.
- **Strong test coverage**: both enabled and disabled branches asserted for the HTTPRoute, SecurityPolicy, realm-setup-job, and gateway cert.
- **The group-claim path-form comment** documents a real cross-repo contract with nebari-operator instead of silently working around it - good defensive documentation.
- No shell-injection risk in realm-setup-job: `$LONGHORN_CLIENT_SECRET` comes from a `secretKeyRef` and is quoted.

## Blocker

**The `securitypolicies` Application is created unconditionally, but its only manifest is conditionally empty - this diverges from the MetalLB precedent and breaks the Longhorn-disable path.**

`apps/securitypolicies.yaml` has no `LonghornEnabled` gate and uses `allowEmpty: false`, while `manifests/networking/policies/` contains *only* `longhorn-securitypolicy.yaml`, which renders empty when Longhorn is disabled.

The established pattern for "a provider doesn't need this" is MetalLB: `writer.go` (`isMetalLBPath`) skips writing **both** `apps/metallb*.yaml` **and** `manifests/metallb*`, so no Application is ever created. This PR instead leaves the `securitypolicies` Application pointing at a directory that yields zero resources whenever Longhorn is off (Azure, GCP, local - the primary dev/test path - and AWS/Hetzner with `longhorn_enabled: false`).

Concrete failure mode: with `allowEmpty: false`, when a cluster goes from Longhorn-enabled to disabled (config toggle + redeploy), ArgoCD refuses to prune the SecurityPolicy to zero, leaving an orphaned policy and an `OutOfSync` app that `selfHeal` cannot reconcile.

**Suggested fix:** follow the MetalLB precedent - generalize the skip mechanism (e.g. an `isLonghornPath` check or a small skip-list) so that when `!settings.LonghornEnabled` the writer skips both `apps/securitypolicies.yaml` and `manifests/networking/policies/longhorn-securitypolicy.yaml`. That removes the empty committed file and the orphan-on-disable problem in one move.

**One thing I could not verify without a live ArgoCD:** the exact sync *status* of a fresh non-Longhorn deploy (target=0, live=0) - it may show Synced/Healthy harmlessly rather than erroring. Can you confirm what you observed on a local/Azure deploy? Even if the fresh case is benign, the toggle-off orphan case and the divergence from the deliberate MetalLB skip pattern still stand.

## Question

**deploy.go gating vs Keycloak:** `longhornClientSecret` is generated on `infraSettings.LonghornEnabled` alone, but `createLonghornSecrets` only runs inside `if foundationalCfg.Keycloak.Enabled`. The writer comment says Keycloak is mandatory infrastructure, so `LonghornEnabled && !Keycloak.Enabled` shouldn't occur - is that invariant guaranteed, or relying on `deploy.go` always setting `Keycloak.Enabled: true`? If the latter, a one-line comment at the generation site would prevent a future dangling `secretKeyRef`.

pkg/argocd/templates/apps/securitypolicies.yaml
tylerpotts: Fixed in c9e4450: mirrored the MetalLB skip pattern — added `isLonghornOnlyPath` in `writer.go` that skips writing both `apps/securitypolicies.yaml` and everything under `manifests/networking/policies/` when `LonghornEnabled` is false, so a Longhorn-disabled cluster never gets the empty Application and an enabled->disabled toggle removes both files from the gitops repo (letting root-app pruning delete the Application and cascade-prune the SecurityPolicy). Also removed the now-redundant `{{- if .LonghornEnabled }}` guard from the manifest template and updated the tests to assert the files are absent when disabled.

pkg/providers/cluster/aws/state.go
```diff
@@ -119,7 +123,7 @@ func ensureStateBucket(ctx context.Context, client S3Client, region, bucketName
 	createInput := &s3.CreateBucketInput{
 		Bucket: aws.String(bucketName),
 	}
-	if region != "us-east-1" {
+	if region != awsRegionUSEast1 {
```
viniciusdc: Why is there a need to define this as a new constant?

tylerpotts: Good catch — it was only used in this one place, so the constant added indirection without much benefit. Inlined the `"us-east-1"` literal and moved the explanation about S3 special-casing the default region into a comment at the usage site (9ebffdb).

viniciusdc:
Ran this through the diff plus an architecture pass and a second adversarial pass, and confirmed the main thing on a throwaway k3d deploy. The shape is good and the tests are thorough, and the `LonghornEnabled` threading through `InfraSettings` is clean. One thing gates merge for me, plus a few I don't think the tests cover.

The gate is ArgoCD RBAC. The shared `groups` mapper gets flipped to `full.path=true` in `realm-setup-job.yaml`, and that flip isn't behind the `LonghornEnabled` guard, so every client's group claim goes to the path form (`/argocd-admins`). But `pkg/argocd/config.go` still matches the bare `g, argocd-admins`, so an ArgoCD SSO user ends up with an empty app list. I confirmed it on a k3d deploy with Longhorn off: the generated realm-setup renders `full.path=true`, kcadm shows the applied mapper is `full.path=true` with groups resolving to `/argocd-admins`, and `config.go` is still bare. #385 makes `config.go` match both forms, so I think this just needs to land with or after #385, not before. Two nuances: the `kcadm create ... || true` can't flip an existing mapper, so it only bites fresh installs (with or without ds-pack), not existing clusters; and the inline comment credits nebari-operator for the reconcile, but it's actually the data-science-pack chart job (`files/keycloak_rbac_bootstrap.py`), worth fixing so it doesn't mislead the next reader.

A few things outside the test surface I'd want a second look at.

The Longhorn OIDC client secret is create-if-absent in both the K8s Secret and `kcadm create client`, with no update path. If `longhorn-system` ever gets recreated (Longhorn uninstall/reinstall), a redeploy writes a fresh random secret into just that namespace while Keycloak keeps the old one, and login fails with nothing surfaced. There's also no rotation story.

`LONGHORN_CLIENT_SECRET`'s `secretKeyRef` in the realm-setup job isn't `optional: true`. If someone regenerates the manifests with `LonghornEnabled=true` against a cluster where the secret wasn't (re)created, the PostSync realm-setup pod hits `CreateContainerConfigError` and the keycloak app's hook fails. We regenerate manifests without a full deploy fairly often, so this one is easy to hit.

Disabling Longhorn later leaves `apps/securitypolicies.yaml` and `longhorn-securitypolicy.yaml` behind in the gitops repo, since `WriteAllToGit` never removes files, so the SecurityPolicy and the Keycloak client stick around. MetalLB and trust-manager have the same latent behavior, but this instance carries an auth resource, which makes me a bit more nervous about it.

Smaller stuff:

The SecurityPolicy points `jwt.remoteJWKS` at the in-cluster Keycloak but `oidc.provider.issuer` at the external URL with no in-cluster override, so Envoy's OIDC discovery has to reach `https://keycloak.<domain>` and trust its cert. Fine with a real cert, but it breaks on a staging/self-signed issuer, which can still happen on the providers where this actually renders.

`LabelKeyManagedBy`/`LabelValueManagedBy` are byte-identical to the existing `ManagedByLabel`/`NebariManagedByValue`, and the PR moves call sites onto the new ones while leaving the old defined. Should we just collapse to one set?

The `existing` provider actually installs Longhorn when you set `cluster.existing.longhorn` (`existing/config.go` plus `PostProvision`'s `longhorn.Install`), but its `InfraSettings` never sets `LonghornEnabled`, so the UI never gets exposed even when Longhorn really is there. Should `existing` derive `LonghornEnabled` from its `longhorn.enabled` the way AWS and Hetzner do?

The design itself is sound and #385 ordering is the real gate. IMO it'd help to write the groups-claim-format contract down somewhere (`full.path=true`, who reconciles it, that all consumers must match the path form), since that's the crux here and it's undocumented right now.

For what it's worth, I checked these and they're fine: the `longhorn-frontend:80` backendRef matches the chart service, the same-namespace targetRef is valid, the wave-3 app is ok with the Envoy Gateway CRDs at wave 1, and the PostSync hook sidesteps Job immutability.

pkg/argocd/foundational.go
```diff
@@ -330,6 +372,37 @@ func createKeycloakSecrets(ctx context.Context, client kubernetes.Interface, key
 	return nil
 }
 
+// createLonghornSecrets writes the OIDC client secret used to protect the
+// Longhorn UI into both the keycloak namespace (for realm-setup-job) and the
+// longhorn-system namespace (for the Envoy Gateway SecurityPolicy). When
+// longhornSSO.ClientSecret is empty, nothing is created.
+func createLonghornSecrets(ctx context.Context, client kubernetes.Interface, longhornSSO LonghornSSOConfig) error {
```
dcmcand, line 458: **Blocker: cross-namespace secret divergence on a partial-failure retry.**

`createLonghornSecrets` writes `longhorn-oidc-client-secret` into both `keycloak` and `longhorn-system` via `createSecret`, which is create-if-not-exists and never updates. The value is a fresh random secret generated on every `Deploy()` (`deploy.go:222`), and an install failure is downgraded to a warning rather than failing the deploy (`deploy.go:259-262`).

Failure sequence:
1. First deploy: the `keycloak` secret is created (value A); the `longhorn-system` create fails transiently. The error becomes a warning and the deploy appears to succeed.
2. Re-run `nic deploy`: a new random value B is generated. The loop finds the `keycloak` secret already present (keeps A) and creates `longhorn-system` with B.
3. The Keycloak client is registered with A while the Envoy SecurityPolicy presents B, so every login fails with an opaque `invalid_client` and nothing in the logs points back to the namespace mismatch.

This is on the default-on path for AWS and Hetzner, and the two new tests only cover the clean-slate case.

Suggested fix: before writing, check whether the secret already exists in either namespace and treat that value as canonical for the missing namespace, rather than using the freshly generated value whenever a target namespace happens to be empty. That needs `createSecret` (or a sibling helper) to return the existing value so both namespaces reconcile to a single value regardless of which succeeded on a prior attempt.

pkg/argocd/foundational.go
```diff
@@ -33,11 +33,29 @@ const (
 	// NebariFoundationalPartOf is the value of the app.kubernetes.io/part-of label for foundational resources.
 	NebariFoundationalPartOf = "nebari-foundational"
 
+	// LabelKeyPartOf is the standard Kubernetes label key identifying the higher-level app a resource belongs to.
+	LabelKeyPartOf = "app.kubernetes.io/part-of"
+
+	// LabelKeyManagedBy is the standard Kubernetes label key identifying the controller/tool that manages a resource.
+	LabelKeyManagedBy = "app.kubernetes.io/managed-by"
```
dcmcand: Stylistic: `LabelKeyManagedBy` / `LabelValueManagedBy` duplicate the existing `ManagedByLabel` / `NebariManagedByValue` (same string values, still used at `certificate.go:158`). The package now has two names for each of two identical constants. Consider consolidating onto one pair, either reusing the existing ones here or migrating `certificate.go` to the new names and deleting the old ones.

pkg/argocd/templates/manifests/networking/policies/longhorn-securitypolicy.yaml
```diff
@@ -0,0 +1,51 @@
...
+  targetRefs:
+    - group: gateway.networking.k8s.io
+      kind: HTTPRoute
+      name: longhorn
+  oidc:
+    provider:
+      issuer: "https://keycloak.{{ .Domain }}{{ .KeycloakBasePath }}/realms/nebari"
+    clientID: longhorn
+    clientSecret:
+      name: longhorn-oidc-client-secret
```
dcmcand: Stylistic: the secret name is hardcoded here and again in `realm-setup-job.yaml:44`, while the Go side has a `LonghornOIDCClientSecretName` constant that is never threaded into `TemplateData`. Compare `KeycloakAdminSecretName`, which is passed through `TemplateData` specifically to keep the Go constant and the rendered YAML in sync. Threading this one the same way would make a future rename mechanically safe instead of relying on grep discipline.

pkg/argocd/writer_test.go
```diff
@@ -657,6 +657,10 @@ func TestServiceHTTPRoutes_TargetHTTPSListener(t *testing.T) {
 
 			output := string(processed)
 
+			if strings.TrimSpace(output) == "" {
+				t.Skipf("skipping %s: empty render with default test data", name)
```
dcmcand, line 667: Stylistic: this `t.Skipf` skips any route template that renders empty with the default (zero-value) test data, not just `longhorn-httproute.yaml`. That is fine today since `TestWriteAllToGit_LonghornHTTPRoute` independently pins the "targets https listener" property for Longhorn, but a future conditionally-rendered route would silently drop out of this generic check. A short comment noting that expectation would keep it from being lost.

pkg/argocd/templates/manifests/networking/policies/longhorn-securitypolicy.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - groups
+    forwardAccessToken: true
+  jwt:
+    providers:
+      - name: keycloak
+        issuer: "https://keycloak.{{ .Domain }}{{ .KeycloakBasePath }}/realms/nebari"
+        remoteJWKS:
+          uri: "{{ .KeycloakServiceURL }}/realms/nebari/protocol/openid-connect/certs"
+  authorization:
+    defaultAction: Deny
```
dcmcand, line 35: Deny-by-default with a single explicit `allow-longhorn-admins` rule is the right authorization posture here: anything whose `groups` claim does not match gets a 403 at the gateway rather than falling through. Good.

pkg/argocd/templates/manifests/networking/policies/longhorn-securitypolicy.yaml
```diff
@@ -0,0 +1,51 @@
...
+    rules:
+      - name: allow-longhorn-admins
+        action: Allow
+        principal:
+          jwt:
+            provider: keycloak
+            claims:
+              # nebari-operator's data-science-pack RBAC bootstrap reconciles
```
dcmcand: Really appreciate this comment. The `full.path=true` reconciliation in nebari-operator is exactly the kind of non-obvious cross-component behavior that costs someone hours, and documenting it right where the leading-slash match matters (with the matching note in `realm-setup-job.yaml`) is excellent point-of-use documentation.

dcmcand (dismissed):
**Re-review of #328** (head `e1b4d9d`). Two passes this time: a design-conformance pass and an infra-Go correctness pass. Both prior review items are resolved, so this focuses on what is new.

## Prior review items: both resolved

- **SecurityPolicy Application gating (@dcmcand's blocker): resolved.** `isLonghornOnlyPath` in `writer.go` skips both `apps/securitypolicies.yaml` and the whole `manifests/networking/policies/` directory when Longhorn is disabled, mirroring `isMetalLBPath`/`isTrustBundlePath`. The now-redundant in-template guard was removed, and on an enable-to-disable toggle the prune cascades correctly through `root.yaml`'s finalizer. Covered by the positive and negative cases in `TestWriteAllToGit_LonghornSecurityPolicy`.
- **`aws/state.go` constant (@viniciusdc's question): resolved.** Inlined to a `"us-east-1"` literal with an accurate comment about S3 special-casing the default region.

## Overall

Nice work. The design is clean: the new capability flows through `InfraSettings.LonghornEnabled` and is mirrored into `TemplateData`, with no provider-name switches and no concrete-provider imports crossing the CLI/library boundary. Provider wiring is consistent across aws/azure/gcp/hetzner/local, and `go build`, `go vet`, `golangci-lint run`, and `go test ./... -race` all pass at head.

There is one blocker (secret divergence on a partial-failure retry, left inline on `foundational.go`), one genuine design question about ADR-0006 scope (below), and a few stylistic notes inline. Positive highlights are inline too.

## Question: is conditional-manifest gating in scope for ADR-0006?

This PR gates GitOps-rendered manifests (HTTPRoute, SecurityPolicy, cert SAN, Keycloak client) with an `InfraSettings` flag plus a `writer.go` skip predicate. ADR-0006 (Proposed) flags "conditional foundational software" and calls the flag-plus-skip pairing out as not scaling, but its worked examples are all Helm chart installs, and its enforcement mechanism (#349) does not exist yet. The live pattern for conditional manifests, including `trust-bundle` which merged after ADR-0006 was proposed, still uses this approach. So: is UI-exposure manifest gating meant to be in or out of ADR-0006's scope? If out, a one-line scope note in the ADR would let future reviews tell the two cases apart. Happy to open that as a follow-up if useful.

## Suggested follow-ups (not blocking)

1. `InstallFoundationalServices` derives its Kubernetes client internally rather than accepting a `kubernetes.Interface`, so the secret path is not directly unit-testable, which is exactly why the blocker scenario slips past the two new tests. Refactoring it to take an interface would make the reconciliation path testable.
2. Document the ADR-0006 scope distinction (charts vs rendered manifests), per the question above.
3. Consolidate the duplicate managed-by label constants package-wide (see inline on `foundational.go`).
4. Toggling `longhorn.enabled: false` prunes the SecurityPolicy but leaves the two imperatively-created OIDC secrets orphaned. This matches existing bootstrap behavior, but it is worth either cleaning up or documenting.
5. This is the first Envoy Gateway `SecurityPolicy` using `oidc:`/`jwt:` in the repo, which adds an EG-to-Keycloak discovery path and an in-cluster JWKS hairpin. Worth confirming it holds up under restricted egress before the pattern gets reused (relevant to the enterprise CA-trust epic).

Requesting changes for the blocker only. Everything else is a question or optional.

pkg/argocd/foundational.go
tylerpotts: Fixed in 7e281ac. `createLonghornSecrets` now reconciles instead of create-if-absent: it first reads the secret from both namespaces, and any existing value is canonical over the freshly generated one (keycloak checked first, since realm-setup-job registers the Keycloak client from that copy). Missing copies are created with the canonical value and diverged copies are updated to it, so your failure sequence converges to value A on the re-run. A transient Get error now fails the deploy rather than being mistaken for "absent" and seeding a fresh value. Added two regression tests: the partial-failure retry (keycloak has A, longhorn-system missing, new value B generated -> both end at A) and the diverged case (A/B -> both reconciled to A).

pkg/argocd/foundational.go
tylerpotts: Consolidated in the merge commit 23ad9df: main had meanwhile standardized on `PartOfLabel`/`ManagedByLabel`/`NebariManagedByValue`, so the branch-local `LabelKeyPartOf`/`LabelKeyManagedBy`/`LabelValueManagedBy` set is deleted and all call sites (including `createLonghornSecrets`) use the existing pair.

pkg/argocd/templates/manifests/networking/policies/longhorn-securitypolicy.yaml
tylerpotts: Fixed in 7e281ac: added `TemplateData.LonghornOIDCSecretName`, populated from the `LonghornOIDCClientSecretName` constant in `NewTemplateData` (same pattern as `KeycloakAdminSecretName`), and both `longhorn-securitypolicy.yaml` and `realm-setup-job.yaml` now render `{{ .LonghornOIDCSecretName }}` instead of the hardcoded string.

pkg/argocd/writer_test.go
tylerpotts: Added in 7e281ac: a comment above the `t.Skipf` noting that it skips any route rendering empty with zero-value data, and that each conditionally rendered route therefore needs its own test pinning the https-listener property with its gate enabled (pointing at `TestWriteAllToGit_LonghornHTTPRoute` as the example).

tylerpotts:
@viniciusdc thanks for the thorough pass - addressed in 23ad9df and 7e281ac:

- **#385 ordering**: agreed, that's the real gate. No code change here; this should merge with or after https://github.com/nebari-dev/nebari-infrastructure-core/pull/385 (still open as of now).
- **Comment attribution**: fixed in both `realm-setup-job.yaml` and `longhorn-securitypolicy.yaml` - the reconcile is credited to the data-science-pack chart's RBAC bootstrap job (`files/keycloak_rbac_bootstrap.py`), not nebari-operator.
- **Create-if-absent secret with no update path**: `createLonghornSecrets` now reconciles. An existing value in either namespace is canonical over the freshly generated one (keycloak first, since the Keycloak client is registered from that copy), missing copies are created with it and diverged copies updated to it. That covers your longhorn-system-recreated scenario: the redeploy now converges to the keycloak value instead of writing a fresh one into just longhorn-system. Rotation is still an open follow-up - there's deliberately no path that changes an established value yet.
- **`optional: true`**: done, plus the job script now skips Longhorn client setup with an explicit log line when `LONGHORN_CLIENT_SECRET` is unset, so a regenerated manifest without a full deploy neither hits `CreateContainerConfigError` nor registers the client with an empty secret.
- **Files left behind on disable**: you were right - `WriteAllToGit` only skipped, never removed. It now deletes previously written gated output when the gate is off (new `removeStaleTemplate`), and this applies uniformly to Longhorn, MetalLB, trust-manager, and certificate templates, so the latent MetalLB/trust-manager variant is fixed too. Regression test covers the enable->disable toggle. Caveat: on an already-bootstrapped repo the writer only runs with `--regen-apps` (existing behavior, unchanged).
- **Issuer URL vs in-cluster JWKS on self-signed certs**: real limitation, agreed. Envoy's OIDC discovery goes to the external issuer URL and must trust its cert. I'd rather handle that alongside the enterprise CA-trust work than special-case it here - noting it as a follow-up.
- **Label constants**: consolidated onto `ManagedByLabel`/`NebariManagedByValue` while merging latest main (main had already standardized on those).
- **existing/openshift**: still silently false - giving those providers an opt-in needs a config surface decision, so I'd keep it out of this PR and open a follow-up issue.
- **Groups-claim contract**: written down in `docs/keycloak-groups-claim.md` - `full.path=true`, who reconciles it, and that all consumers must match the path form (with the #385 compatibility note).

tylerpotts:
@dcmcand re-review addressed in 23ad9df (merge with latest main) and 7e281ac - the blocker fix and stylistic items are answered on the inline threads.

**On the ADR-0006 scope question**: I read the ADR's target as Helm chart installs (its examples, and the enforcement mechanism in https://github.com/nebari-dev/nebari-infrastructure-core/issues/349, are all chart-shaped). Gating rendered manifests with an `InfraSettings` flag plus a writer skip predicate is the established live pattern (MetalLB, trust-bundle, now Longhorn UI), and I think it should stay out of ADR-0006's scope - a one-line scope note in the ADR distinguishing charts from rendered manifests would keep future reviews from relitigating it. Please do open that follow-up if you're willing.

**On the suggested follow-ups**:
1. `createLonghornSecrets` itself takes `kubernetes.Interface`, and the new divergence tests exercise the reconciliation path directly with a fake clientset - that's how the blocker scenario is now covered. Agreed that `InstallFoundationalServices` accepting an interface would make the whole secret path testable end-to-end; good follow-up.
2. Covered by the scope note above.
3. Done - consolidated while merging main.
4. Partially improved: `WriteAllToGit` now removes gated template output on toggle-off (so the SecurityPolicy prunes), but the two imperatively created OIDC secrets and the Keycloak client do still linger. Follow-up material.
5. Agreed - the EG-to-Keycloak discovery path and JWKS hairpin under restricted egress should be validated before this pattern is reused; it overlaps with the enterprise CA-trust epic and Vini's self-signed-issuer point.

pkg/argocd/templates/manifests/networking/policies/longhorn-securitypolicy.yaml
tylerpotts: Thanks - deny-by-default felt right for an admin UI sitting behind the gateway; anything new has to be allowed explicitly.

pkg/argocd/templates/manifests/networking/policies/longhorn-securitypolicy.yaml
tylerpotts: Thanks. One correction that came out of viniciusdc's pass: the reconciler is actually the data-science-pack chart's RBAC bootstrap job (files/keycloak_rbac_bootstrap.py), not nebari-operator - the comment now credits the right component (7e281ac). The contract is also written down in docs/keycloak-groups-claim.md so it lives somewhere greppable beyond the two point-of-use comments.

PR #330, opened 2026-05-27
feat(aws): make load balancer scheme configurable

dcmcand:
## Summary

Adds `cluster.aws.load_balancer_scheme` so operators can override the hardcoded `internet-facing` scheme that NIC stamps onto the Gateway's NLB. Defaults preserve existing behavior; valid values are `internet-facing` (default) and `internal`.

Without this, a private-only VPC (no public subnets, common in GovCloud and regulated environments) causes the AWS Load Balancer Controller to look for public subnets it cannot find, and the deploy hangs in the LB endpoint wait.

Closes #305.

## Changes

- `pkg/provider/aws/config.go` - new `LoadBalancerScheme` field + `LoadBalancerSchemeOrDefault()` helper + scheme constants
- `pkg/provider/aws/provider.go` - `InfraSettings()` substitutes the configured value into the scheme annotation; `Validate()` rejects values other than `internal` / `internet-facing` so typos fail fast
- `pkg/provider/aws/provider_test.go` - new table tests covering default + both explicit values + three rejection cases
- `examples/aws-config.yaml` - documents the new optional field with the private-VPC use case

No Terraform module changes needed - the LBC reads the annotation off the Service at runtime.

## Test plan

- [x] Unit tests cover default + both explicit values
- [x] Unit tests cover validation rejection (typo, arbitrary string, mixed case)
- [x] `go test ./...` passes
- [x] `golangci-lint run` clean
- [x] Manual: deploy `dcmcand-305` with `load_balancer_scheme: internal` - confirmed end-to-end:
  - Service annotation: `service.beta.kubernetes.io/aws-load-balancer-scheme: internal`
  - AWS NLB: `Scheme: internal`, `Type: network`
  - LB DNS resolves to RFC1918 only (`10.10.156.89`)
- [ ] Manual: regression check on a normal cluster with the field unset - confirm NLB is still internet-facing (lower priority; the default path is unit-tested and the diff for that path is one constant assignment)
- [ ] Manual (gated on #283 landing): deploy into a fully private VPC built by NIC and confirm the full original failure mode is fixed end-to-end

### DIFF
--- a/pkg/provider/aws/config.go
+++ b/pkg/provider/aws/config.go
@@ -27,6 +27,22 @@ type Config struct {
 	EFS                       *EFSConfig                       `yaml:"efs,omitempty"`
 	Longhorn                  *longhorn.Config                 `yaml:"longhorn,omitempty"`
 	AWSLoadBalancerController *AWSLoadBalancerControllerConfig `yaml:"aws_load_balancer_controller,omitempty"`
+	LoadBalancerScheme        string                           `yaml:"load_balancer_scheme,omitempty"`
+}
+
+const (
+	LoadBalancerSchemeInternetFacing = "internet-facing"
+	LoadBalancerSchemeInternal       = "internal"
+)
+
+// LoadBalancerSchemeOrDefault returns the configured AWS load balancer scheme,
+// defaulting to "internet-facing" when unset. Values are validated at config
+// load time, so callers can trust the result is one of the supported schemes.
+func (c *Config) LoadBalancerSchemeOrDefault() string {
+	if c.LoadBalancerScheme == "" {
+		return LoadBalancerSchemeInternetFacing
+	}
+	return c.LoadBalancerScheme
 }
 
 type AWSLoadBalancerControllerConfig struct {
--- a/pkg/provider/aws/config.go
+++ b/pkg/provider/aws/config.go
@@ -31,16 +31,21 @@ type Config struct {
 }
 
 const (
-	LoadBalancerSchemeInternetFacing = "internet-facing"
-	LoadBalancerSchemeInternal       = "internal"
+	loadBalancerSchemeInternetFacing = "internet-facing"
+	loadBalancerSchemeInternal       = "internal"
 )
+
+var validLoadBalancerSchemes = []string{
+	loadBalancerSchemeInternetFacing,
+	loadBalancerSchemeInternal,
+}
 
 // LoadBalancerSchemeOrDefault returns the configured AWS load balancer scheme,
 // defaulting to "internet-facing" when unset. Values are validated at config
 // load time, so callers can trust the result is one of the supported schemes.
 func (c *Config) LoadBalancerSchemeOrDefault() string {
 	if c.LoadBalancerScheme == "" {
-		return LoadBalancerSchemeInternetFacing
+		return loadBalancerSchemeInternetFacing
 	}
 	return c.LoadBalancerScheme
 }
--- a/pkg/provider/aws/provider.go
+++ b/pkg/provider/aws/provider.go
@@ -139,11 +139,9 @@ func (p *Provider) Validate(ctx context.Context, projectName string, clusterConf
 	}
 
 	// Validate load_balancer_scheme if specified
-	if awsCfg.LoadBalancerScheme != "" &&
-		awsCfg.LoadBalancerScheme != LoadBalancerSchemeInternetFacing &&
-		awsCfg.LoadBalancerScheme != LoadBalancerSchemeInternal {
-		err := fmt.Errorf("invalid load_balancer_scheme %q: must be %q or %q",
-			awsCfg.LoadBalancerScheme, LoadBalancerSchemeInternetFacing, LoadBalancerSchemeInternal)
+	if awsCfg.LoadBalancerScheme != "" && !contains(validLoadBalancerSchemes, awsCfg.LoadBalancerScheme) {
+		err := fmt.Errorf("invalid load_balancer_scheme %q (must be one of: %v)",
+			awsCfg.LoadBalancerScheme, validLoadBalancerSchemes)
 		span.RecordError(err)
 		return err
 	}
@@ -731,7 +729,7 @@ func (p *Provider) Summary(clusterConfig *config.ClusterConfig) map[string]strin
 func (p *Provider) InfraSettings(clusterConfig *config.ClusterConfig) provider.InfraSettings {
 	sc := longhorn.StorageClassName
 	var efsSC string
-	lbScheme := LoadBalancerSchemeInternetFacing
+	lbScheme := loadBalancerSchemeInternetFacing
 
 	rawCfg := clusterConfig.ProviderConfig()
 	if rawCfg != nil {
--- a/pkg/provider/aws/provider.go
+++ b/pkg/provider/aws/provider.go
@@ -138,6 +138,16 @@ func (p *Provider) Validate(ctx context.Context, projectName string, clusterConf
 		}
 	}
 
+	// Validate load_balancer_scheme if specified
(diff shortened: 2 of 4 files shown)


oren-openteams:
Tested against private-only GovCloud VPC (us-gov-west-1, no IGW/NAT — egress via TGW). Built `nic` from PR head `d1fc0f3` and deployed against an existing VPC (`existing_vpc_id`) with `load_balancer_scheme: internal`. Results:

- ✅ `service.beta.kubernetes.io/aws-load-balancer-scheme: internal` rendered onto the Envoy Gateway Service.
- ✅ AWS NLB came up with `Scheme: internal`, `Type: network`.
- ✅ NLB private IPs are in the VPC CIDR (10.20.0.0/22).
- ✅ Full NIC bootstrap completed in a private-only VPC — no hang waiting for an internet-facing endpoint that would have nowhere to land. This is the original failure mode and the fix resolves it.

LGTM, ready to merge from my side.

oren-openteams (approved):
Verified end-to-end against a private-only GovCloud VPC — see comment above for details.

pkg/provider/aws/config.go
```diff
@@ -27,6 +27,27 @@ type Config struct {
...
+
+var validLoadBalancerSchemes = []string{
+	loadBalancerSchemeInternetFacing,
+	loadBalancerSchemeInternal,
+}
+
+// LoadBalancerSchemeOrDefault returns the configured AWS load balancer scheme,
+// defaulting to "internet-facing" when unset. Values are validated at config
+// load time, so callers can trust the result is one of the supported schemes.
+func (c *Config) LoadBalancerSchemeOrDefault() string {
```
tylerpotts, line 46: Nice docstring — explicitly stating *why* callers can trust the return value (the validation invariant established at config-load time) is the right level of comment. Easy to skip; you didn't.

pkg/provider/aws/provider.go
```diff
@@ -138,6 +138,14 @@ func (p *Provider) Validate(ctx context.Context, projectName string, clusterConf
 		}
 	}
 
+	// Validate load_balancer_scheme if specified
+	if awsCfg.LoadBalancerScheme != "" && !contains(validLoadBalancerSchemes, awsCfg.LoadBalancerScheme) {
```
tylerpotts, line 142: Good call reusing the existing `contains()` helper rather than an inline equality chain — matches the convention already set by the taint-effect validator at `provider.go:191-196`. Style-consistent.

pkg/provider/aws/provider.go
```diff
@@ -138,6 +138,14 @@ func (p *Provider) Validate(ctx context.Context, projectName string, clusterConf
...
+		err := fmt.Errorf("invalid load_balancer_scheme %q (must be one of: %v)",
```
tylerpotts, line 143: **Nit (non-blocking):** `%v` on `validLoadBalancerSchemes` renders as `[internet-facing internal]`. Quoting (`%q`) or a comma-joined string reads a little nicer in CLI output. Truly nitpick territory.

pkg/provider/aws/provider.go
```diff
@@ -714,12 +722,14 @@ func (p *Provider) Summary(clusterConfig *config.ClusterConfig) map[string]strin
...
-// aws-load-balancer-scheme=internet-facing, LBC creates an internal NLB by
-// default, which is not reachable from outside the VPC.
+// Controller (type=external) and request an IP-targeted NLB. The
+// aws-load-balancer-scheme annotation defaults to "internet-facing" so the NLB
+// is reachable from outside the VPC; operators with private-only VPCs can
+// override this to "internal" via cluster.aws.load_balancer_scheme.
 func (p *Provider) InfraSettings(clusterConfig *config.ClusterConfig) provider.InfraSettings {
 	sc := longhorn.StorageClassName
 	var efsSC string
+	lbScheme := loadBalancerSchemeInternetFacing
```
tylerpotts, line 732: **Nit (non-blocking):** the fallback to `loadBalancerSchemeInternetFacing` here duplicates the default established by `LoadBalancerSchemeOrDefault()` at `config.go:48`. Not a bug — this branch only fires when `ProviderConfig()` is nil or unmarshal fails, and line 744 overrides it on the happy path. But initializing `lbScheme` to `""` and always going through `LoadBalancerSchemeOrDefault()` would centralize the default in one place. Tiny.

pkg/provider/aws/provider.go
```diff
@@ -714,12 +722,14 @@ func (p *Provider) Summary(clusterConfig *config.ClusterConfig) map[string]strin
 // EFSStorageClass is set when EFS is enabled.
 //
 // LoadBalancerAnnotations route the Gateway's Service to the AWS Load Balancer
```
tylerpotts, line 724: Glad you updated this docstring — the old "Without ... internet-facing, LBC creates an internal NLB by default" rationale would have been outright wrong once both modes are supported. Easy to forget.

pkg/provider/aws/provider_test.go
```diff
@@ -55,3 +57,77 @@ func TestInfraSettings(t *testing.T) {
 		})
 	}
 }
+
+// TestValidate_LoadBalancerScheme covers the rejection path only. Valid
```
tylerpotts, line 61: Appreciated — the comment pre-empts the obvious "why don't you test happy paths through `Validate()`?" review question with the real reason (credentials check needs live AWS). Saves a round-trip.

pkg/provider/aws/provider_test.go
```diff
@@ -55,3 +57,77 @@ func TestInfraSettings(t *testing.T) {
...
+// environment. The default and valid-value paths are exercised by
+// TestInfraSettings and TestInfraSettings_LoadBalancerScheme.
+func TestValidate_LoadBalancerScheme(t *testing.T) {
+	tests := []struct {
+		name   string
+		scheme string
+	}{
+		{name: "typo is rejected", scheme: "internet_facing"},
+		{name: "arbitrary string is rejected", scheme: "public"},
+		{name: "mixed case is rejected", scheme: "Internal"},
```
tylerpotts, line 73: Thoughtful edge case. The AWS API is case-sensitive on this annotation — silently accepting `"Internal"` would produce a misconfigured LB at deploy time. Catching it at `Validate()` is exactly right.

tylerpotts (approved):
## Review: feat(aws): make load balancer scheme configurable

### Issue Resolution: ✅ Fully Resolved

Issue #305 has four DoD items — all four are met:

| DoD item | Status |
|---|---|
| `load_balancer_scheme` field with `internal`/`internet-facing` validation | ✅ `pkg/provider/aws/config.go:30,38-41`, `provider.go:141-147` |
| `InfraSettings()` uses the configured value | ✅ `provider.go:744,755` |
| Unit tests cover both modes | ✅ `provider_test.go:99-133` plus rejection cases at `:66-97` |
| Example config / docs reference the new field | ✅ `examples/aws-config.yaml:86-90` |
| Manual verification | ✅ PR description confirms NLB scheme + RFC1918 DNS on a live cluster |

The PR description also correctly calls out that the fully-private-VPC end-to-end test is gated on #283 — honest framing; this PR delivers exactly the unit of work issue #305 asked for.

### General Notes

Tight, well-scoped PR. ~30 lines of code plus tests, defaults preserve current behavior, validation fails fast and case-sensitively, the example config is updated with the *why*, and end-to-end manual verification on a live cluster is in the description. The tightening commit (`d1fc0f3`) shows the author responds well to feedback — unexporting constants, switching to the `contains()` helper, splitting Validate vs. InfraSettings test coverage.

### Supply-chain / scope check

Clean. No new deps, no `init()`, no `os/exec`, no `unsafe`, no embedded files, no network calls in this diff. The four changed files match exactly what the PR claims to do.

Approving. Two minor nits as inline comments — neither blocking.

PR #331, opened 2026-05-27
feat(otel): software-pack extension point for collector config + cluster RBAC

tylerpotts:
## Summary

Adds a positional extension point to NIC's OpenTelemetry collector so that Nebari software packs (starting with [nebari-lgtm-pack](https://github.com/nebari-dev/nebari-lgtm-pack)) can supply additional collector configuration without modifying NIC or sharing a Kubernetes resource with NIC.

**Extension point** (added to the chart's helm values in `pkg/argocd/templates/apps/opentelemetry-collector.yaml`):

- `extraVolumes` mounts an optional ConfigMap named `opentelemetry-collector-overrides` (must live in `monitoring`).
- An `ensure-overrides` init container copies `relay.yaml` from that ConfigMap into an emptyDir, or writes `{}` if the ConfigMap doesn't exist.
- `command.extraArgs` adds `--config=/conf/overrides/relay.yaml` so the collector deep-merges the override config with its base config at startup.

**Behavior:**

- Fresh cluster, no software pack installed → init container writes `{}`, collector runs with NIC's default debug exporters. Unchanged from before.
- Cluster with nebari-lgtm-pack installed → pack renders `opentelemetry-collector-overrides`, init container resolves it, collector routes logs/traces/metrics to Loki/Tempo/Mimir.

**Also in this PR (independent fix surfaced in the same investigation):**

- `clusterRole.create: true` with rules giving the collector's ServiceAccount `pods`/`nodes`/`services`/`endpoints` list+watch. Without this the prometheus receiver's `kubernetes_sd_configs` (role: pod) silently discovers nothing — caught when Mimir was empty even with the LGTM exporters wired.

Companion PR: https://github.com/nebari-dev/nebari-lgtm-pack/pull/10

Closes: companion to https://github.com/nebari-dev/nebari-lgtm-pack/issues/8

## Test plan

- [x] `go test ./pkg/argocd/...` — `TestWriteApplication_OtelCollector_OverridesExtensionPoint` locks in the new structure (extraVolumes, init container, --config arg) and asserts the old-design fragments (`ignoreDifferences:`, `RespectIgnoreDifferences=true`) are absent.
- [x] `golangci-lint run ./pkg/argocd/...` — clean.
- [x] `helm template` of the OTel chart with NIC's values renders the init container, optional configMap volume, emptyDir resolved-overrides volume, and `--config=/conf/overrides/relay.yaml` in the DaemonSet args.
- [x] End-to-end on a Hetzner k3s cluster with nebari-lgtm-pack@50cc920 installed: collector pods rolled, override CM resolved, Mimir received metrics from 8 scrape targets, logs flowing through Loki. Triggered a separate OTel app sync afterwards and confirmed override CM remained intact.

Refs: https://github.com/nebari-dev/nebari-lgtm-pack/issues/8


pkg/argocd/templates/apps/opentelemetry-collector.yaml
```diff
@@ -58,6 +78,49 @@ spec:
                         regex: ([^:]+)(?::\d+)?;(\d+)
                         replacement: $1:$2
                         target_label: __address__
+
+                  # cAdvisor — per-container CPU/memory/network metrics
+                  # exposed by the kubelet. Required for the Kubernetes
+                  # Views / Pods + Namespaces dashboards (container_cpu_*,
+                  # container_memory_*). Routed via the apiserver proxy so
+                  # it works regardless of kubelet network reachability.
+                  - job_name: 'kubernetes-cadvisor'
```
dcmcand, line 98: [BLOCKER] In daemonset mode there is one collector pod per node, and `role: node` service discovery returns every node in the cluster. So every collector pod scrapes every node's cAdvisor (and kubelet, in the job below) through the apiserver proxy: on a 10-node cluster that is 100 scrape sets, all with identical label sets after the labelmap relabeling. Mimir will either reject the duplicate samples or silently overwrite them depending on remote-write config. A small e2e cluster would not surface this.

Two ways out:

- Use the chart's `presets.kubeletMetrics.enabled: true`, which uses the kubeletstats receiver pinned to `${env:K8S_NODE_IP}` so each pod only scrapes its own node.
- If the apiserver-proxy + prometheus-format approach is deliberate, add a `keep` relabel rule on `__meta_kubernetes_node_name` matching `${env:K8S_NODE_NAME}` so each pod drops every node but its own (the env var may need adding via `extraEnvs` with a `fieldRef` to `spec.nodeName` if no preset already injects it).

pkg/argocd/templates/apps/opentelemetry-collector.yaml
```diff
@@ -58,6 +78,49 @@ spec:
...
+                    scheme: https
+                    tls_config:
+                      ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+                      insecure_skip_verify: true
```
dcmcand: [BLOCKER] `insecure_skip_verify: true` and `ca_file` contradict each other, and skip-verify wins: the CA file is loaded and then ignored, so these scrapes run TLS without server authentication. Since the target is `kubernetes.default.svc`, whose cert is signed by exactly the CA in the mounted `ca.crt`, verification should just work. Drop `insecure_skip_verify: true` here and in the kubernetes-kubelet job below. Skip-verify is only needed when scraping kubelets directly with their self-signed certs, which the apiserver-proxy design specifically avoids.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
```diff
@@ -58,6 +78,49 @@ spec:
...
+                      - role: node
+                    relabel_configs:
+                      - action: labelmap
+                        regex: __meta_kubernetes_node_label_(.+)
+                      - target_label: __address__
+                        replacement: kubernetes.default.svc:443
+                      - source_labels: [__meta_kubernetes_node_name]
+                        regex: (.+)
+                        target_label: __metrics_path__
+                        replacement: /api/v1/nodes/$${1}/proxy/metrics/cadvisor
```
dcmcand, line 116: The `$${1}` escaping is correct here, and easy to get wrong: the collector's confmap expansion turns `$$` into a literal `$` before prometheus ever sees the replacement string. Nice. One consistency note: the pre-existing kubernetes-pods job above uses bare `$1:$2`, which happens to pass through unexpanded on current collector versions but is the deprecated form. Worth normalizing it to `$$1:$$2` in a follow-up.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
```diff
@@ -103,6 +166,49 @@ spec:
...
+        # Software-pack extension point. A Nebari software pack (e.g.
+        # nebari-lgtm-pack) creates a ConfigMap named
+        # `opentelemetry-collector-overrides` containing `relay.yaml` — a
+        # partial OTel config that's deep-merged with the base config above
+        # at collector startup. The init container copies the override file
+        # into an emptyDir, falling back to an empty mapping `{}` if no
+        # software pack has provided one. We do NOT try to share the chart's
+        # main ConfigMap because ArgoCD's `ignoreDifferences` does not survive
+        # sync (upstream issue argoproj/argo-cd#7478) — separate ConfigMaps
+        # mean NIC and software packs never write to the same resource.
```
dcmcand: [BLOCKER] This comment block is currently the only documentation of a cross-repo API: the ConfigMap name, the `monitoring` namespace, the `relay.yaml` key, the mount path, and the absent-means-`{}` fallback are all load-bearing for nebari-lgtm-pack and any future pack. That contract needs a real home before external repos depend on it. An ADR (0005-otel-collector-software-pack-override-point or similar) would capture the contract surface, the argo-cd#7478 rationale for separate ConfigMaps, the considered alternatives, and how this relates to the ADR-0003 pack-integration model, so we end up with one coherent extension story instead of two.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
```diff
@@ -103,6 +166,49 @@ spec:
...
+        extraVolumes:
+          - name: overrides-src
+            configMap:
+              name: opentelemetry-collector-overrides
+              optional: true
```
dcmcand, line 200: [BLOCKER] `optional: true` makes the contract's failure mode completely silent: a pack that ships the wrong ConfigMap name, namespace, or key just gets `{}` merged in and the collector runs happily on defaults. Nothing logs, nothing fails, and the pack author finds out when their dashboards are empty. At minimum the init container should log which path it took (found `relay.yaml` vs wrote the fallback) so it is diagnosable from pod logs; better, the doc/ADR that defines this contract should spell out the failure mode and how a pack author verifies their override actually landed.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
```diff
@@ -24,6 +24,24 @@ spec:
...
+        clusterRole:
+          create: true
+          rules:
+            - apiGroups: [""]
+              resources: ["pods", "nodes", "services", "endpoints"]
+              verbs: ["get", "list", "watch"]
+            - apiGroups: [""]
+              resources: ["nodes/metrics", "nodes/proxy"]
+              verbs: ["get"]
+            - nonResourceURLs: ["/metrics"]
```
dcmcand: [STYLISTIC] Two smaller things on the RBAC block:

- `nodes/proxy` get is inherently broad: the kubelet proxy reaches more than `/metrics`, and RBAC cannot restrict it by path. Worth a sentence in the comment acknowledging that limitation so a future reviewer does not read it as an oversight.
- The `nonResourceURLs: ["/metrics"]` rule is not used by any current scrape job (the apiserver-proxy paths are resource URLs covered by `nodes/proxy`). I would drop it until an apiserver scrape config actually exists, in line with the least-privilege intent of the rest of the block.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
```diff
@@ -103,6 +166,49 @@ spec:
...
+              optional: true
+          - name: overrides-resolved
+            emptyDir: {}
+        extraVolumeMounts:
+          - name: overrides-resolved
+            mountPath: /conf/overrides
+            readOnly: true
+        initContainers:
+          - name: ensure-overrides
+            image: busybox:1.37
```
dcmcand: [STYLISTIC] Tag-pinned image: `busybox:1.37` can be repointed upstream without any change visible here. Low blast radius since this container only touches its own emptyDir, but a digest pin is cheap insurance for something that runs on every node at every pod restart.

pkg/argocd/writer_test.go
```diff
@@ -739,3 +739,51 @@ func TestEnvoyGatewayBeforeCertManager(t *testing.T) {
 		t.Errorf("envoy-gateway (%d) must have a lower sync-wave than cert-manager (%d)", envoyWaveNum, certWaveNum)
 	}
 }
+
+func TestWriteApplication_OtelCollector_OverridesExtensionPoint(t *testing.T) {
```
dcmcand, line 743: [STYLISTIC] This project's convention is table-driven tests. A `{name, fragment string, wantPresent bool}` table with `t.Run` subtests would merge the required and forbidden loops and give each fragment its own named failure, which makes CI output much easier to read when one fragment drifts.

pkg/argocd/writer_test.go
```diff
@@ -739,3 +739,51 @@ func TestEnvoyGatewayBeforeCertManager(t *testing.T) {
...
+	// `opentelemetry-collector-overrides` containing `relay.yaml`; the init
+	// container resolves it (or falls back to `{}`) into an emptyDir that the
+	// collector reads via an extra --config flag. This sidesteps the upstream
+	// ArgoCD ignoreDifferences-during-sync bug (argoproj/argo-cd#7478) by
+	// keeping the base CM and the override CM completely separate.
+	requiredFragments := []string{
+		"extraVolumes:",
+		"name: overrides-src",
+		"name: opentelemetry-collector-overrides",
+		"optional: true",
```
dcmcand: [STYLISTIC] `"optional: true"` and `"emptyDir: {}"` are generic enough that any future YAML in this template could satisfy them, so these assertions could keep passing long after the thing they guard is gone. Composite fragments that pin the context are sturdier, e.g. asserting `name: opentelemetry-collector-overrides` together with its indented `optional: true` as one string. That ties the test to indentation, but in a template test the indentation is load-bearing anyway.

dcmcand (dismissed):
Nice work overall: the separate-ConfigMap design with the init container fallback is the right call given argoproj/argo-cd#7478, the inline rationale comments will save the next person real time, the `$$` escaping in the new scrape jobs is correct per the prometheus receiver docs, and the negative assertions guarding against a regression to the old `ignoreDifferences` design are a good idea. The e2e test plan on real hardware is appreciated too.

Requesting changes on four things, detailed inline:

1. The cadvisor/kubelet scrape jobs duplicate metrics once per node in daemonset mode (every collector pod scrapes every node).
2. `insecure_skip_verify: true` defeats the `ca_file` sitting right next to it.
3. The software-pack override contract is a cross-repo API that currently lives only in a YAML comment; it needs an ADR, and reconciling with the ADR-0003 integration model.
4. The contract's failure mode is fully silent (wrong name/namespace/key just means `{}`); it needs at least diagnosability, ideally documented verification steps for pack authors.

One inline question about config merge semantics for pipelines (last-write-wins vs additive), since that defines what pack authors have to re-declare.

Also worth considering: the RBAC + scrape-job change and the extension point share no dependency and revert independently. The PR description already calls the RBAC piece an independent fix. Splitting them would let the extension-point contract discussion proceed without holding up the metrics fix, and both correctness blockers happen to be in the metrics half.

Stylistic notes inline on test structure, fragment specificity, busybox pinning, and RBAC scope.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
tylerpotts: Fixed: Added `K8S_NODE_NAME` env var via `extraEnvs` with `fieldRef` to `spec.nodeName`, and added a `keep` relabel rule on `__meta_kubernetes_node_name` matching `${env:K8S_NODE_NAME}` to both the kubernetes-cadvisor and kubernetes-kubelet jobs. Each daemonset pod now scrapes only its own node, avoiding the N×N scrape explosion.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
tylerpotts: Fixed: Removed `insecure_skip_verify: true` from both kubernetes-cadvisor and kubernetes-kubelet jobs. The apiserver proxy serves the kubernetes.default.svc cert signed by the cluster CA, so verification with the mounted `ca.crt` works correctly.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
tylerpotts: Acknowledged — good catch on the kubernetes-pods job using bare `$1:$2`. Will normalize to `$$1:$$2` in a follow-up PR to keep this one focused on the extension point.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
tylerpotts: Fixed: Created ADR-0005 (0005-otel-collector-software-pack-override-point.md) documenting the full contract surface: ConfigMap name (`opentelemetry-collector-overrides`), namespace (`monitoring`), key (`relay.yaml`), merge behavior via OTel confmap, failure modes, and how to diagnose issues. Updated the inline comment to reference the ADR.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
tylerpotts: Fixed: Added explicit logging to the init container so it reports which path it took:
- `"found /src/relay.yaml from software pack ConfigMap, using override config"` when the override exists
- `"no /src/relay.yaml found (ConfigMap absent or missing key), using empty fallback {}"` when falling back

The ADR documents how pack authors can verify their override landed by checking init container logs.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
tylerpotts: Fixed both:
- Added comment explaining that `nodes/proxy` grants access to the full kubelet proxy, not just /metrics paths, and that RBAC cannot restrict by URL path.
- Removed the unused `nonResourceURLs: ["/metrics"]` rule — no current scrape job uses non-resource URLs (apiserver-proxy paths are resource URLs covered by `nodes/proxy`).

pkg/argocd/templates/apps/opentelemetry-collector.yaml
tylerpotts: Fixed: Pinned busybox to digest `busybox@sha256:9532d8c39891ca2ecde4d30d7710e01fb739c87a8b9299685c63704296b16028` (the 1.37 digest). This ensures reproducibility even if the tag is repointed upstream.

pkg/argocd/writer_test.go
tylerpotts: Fixed: Converted to table-driven test with `{name, fragment, wantPresent}` structure and `t.Run` subtests. Each fragment now has its own named failure, making CI output clearer.

pkg/argocd/writer_test.go
tylerpotts: Fixed: Switched to composite fragments that pin the context. For example, the `optional: true` assertion is now:
```
"name: overrides-src\n            configMap:\n              name: opentelemetry-collector-overrides\n              optional: true"
```
This ties the test to the indentation (which is load-bearing for YAML templates) and will fail if the structure drifts.

docs/adr/0005-otel-collector-software-pack-override-point.md
```diff
@@ -0,0 +1,156 @@
...
+          exporters: [otlphttp/mimir]
+```
+
+### Merge Behavior
+
+The collector is started with two `--config` flags:
+1. The base configuration (from the Helm chart's ConfigMap)
+2. `/conf/overrides/relay.yaml` (from this extension point)
+
+The OTel Collector's confmap system deep-merges these configurations. Later configs override earlier ones for scalar values; arrays and maps are merged recursively.
```
Adam-D-Lewis: Arrays aren't deep merged.  I gave a suggested change, but feel free to restate.

```suggestion
The OTel Collector's confmap system merges these configurations as follows: **maps are deep-merged** (keys defined only in the base config are preserved; later sources override matching keys), but **lists/arrays are not merged — the last `--config` source wins and replaces the list wholesale**. (Appending lists is opt-in via the experimental `confmap.enableMergeAppendOption` feature gate, which NIC does not enable.) Practical consequence for pack authors: specify only the map keys you want to change, but any list you touch — e.g. a pipeline's `exporters` — must be re-stated in full. Setting a map key to `null` deletes that component from the base, so use `{}` for an empty map. See the [confmap merge docs](https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/README.md).
```

Adam-D-Lewis (commented):
Reviewed the OTel collector software-pack override point. The architecture is sound — exposing a separate, pack-owned `opentelemetry-collector-overrides` ConfigMap merged via a second `--config` is the right way to extend an ArgoCD-managed (selfHeal) foundational collector without fighting sync. I confirmed by `helm template`-ing chart 0.143.0 with this PR's values that the chart appends `command.extraArgs` after its own `--config=/conf/relay.yaml`, so the override correctly wins on merge.

Two things look like blockers (both reproduced from that render):

**1. The override is mounted inside the read-only base-config volume.** The chart mounts its base config as a read-only configMap volume at `/conf`; this PR mounts the override emptyDir at `/conf/overrides`, nested inside it:

```yaml
volumeMounts:
  - mountPath: /conf            # configMap volume (read-only)
    name: opentelemetry-collector-configmap
  - mountPath: /conf/overrides  # emptyDir, nested inside /conf
    name: overrides-resolved
```

Mounting a volume into a subdirectory of a read-only configMap volume typically fails container creation (`mkdir /conf/overrides: read-only file system`). Suggest mounting outside `/conf` (e.g. `/conf-overrides`) and pointing `--config` there. Worth a quick `kubectl apply` to confirm on a real cluster.

**2. `$${env:K8S_NODE_NAME}` is over-escaped (lines 105, 128) → cAdvisor/kubelet scrape nothing.** OTel collapses `$$`→`$` ([docs](https://opentelemetry.io/docs/collector/configuration/): "Use `$$` to indicate a literal `$`"), so this resolves to the literal text `${env:K8S_NODE_NAME}`, not the node name — the `keep` relabel matches no node and both new jobs stay empty. Should be single-`$` `${env:K8S_NODE_NAME}`. (The `$${1}` on lines 114/137 is correctly escaped — that one must reach Prometheus literally.)

**Integration note:** end-to-end this is a no-op until the LGTM pack actually creates the `opentelemetry-collector-overrides` ConfigMap (ns `monitoring`, key `relay.yaml`); today the pack only ships an `examples/` file, and it's in the old `config:`-wrapper values format rather than the partial-config format this contract expects.

One doc correction on the merge semantics inline below.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
```diff
@@ -58,6 +86,55 @@ spec:
...
+                  - job_name: 'kubernetes-cadvisor'
+                    scheme: https
+                    tls_config:
+                      ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+                    bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
+                    kubernetes_sd_configs:
+                      - role: node
+                    relabel_configs:
+                      - source_labels: [__meta_kubernetes_node_name]
+                        regex: $${env:K8S_NODE_NAME}
```
Adam-D-Lewis: This is over-escaped. We *want* the OTel confmap resolver to expand the env var here, but `$$` collapses to a literal `$`, so this resolves to the string `${env:K8S_NODE_NAME}` rather than the node name — the `keep` then matches no node and the job scrapes nothing. Single-`$` so OTel substitutes the actual node name before Prometheus sees the regex:

```suggestion
                        regex: ${env:K8S_NODE_NAME}
```

(The `$${1}` in the `__metrics_path__` replacement below is correct as-is — that one must reach Prometheus literally.)

pkg/argocd/templates/apps/opentelemetry-collector.yaml
```diff
@@ -58,6 +86,55 @@ spec:
...
+                  - job_name: 'kubernetes-kubelet'
+                    scheme: https
+                    tls_config:
+                      ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+                    bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
+                    kubernetes_sd_configs:
+                      - role: node
+                    relabel_configs:
+                      - source_labels: [__meta_kubernetes_node_name]
+                        regex: $${env:K8S_NODE_NAME}
```
Adam-D-Lewis: Same over-escaping as the cadvisor job: `$${env:K8S_NODE_NAME}` collapses to the literal `${env:K8S_NODE_NAME}`, so the `keep` matches no node and kubelet metrics come back empty. Single-`$` so OTel expands it to the node name:

```suggestion
                        regex: ${env:K8S_NODE_NAME}
```

docs/adr/0005-otel-collector-software-pack-override-point.md
```diff
@@ -0,0 +1,156 @@
...
+**Good:**
+- NIC and software packs never write to the same Kubernetes resource
+- ArgoCD sync cannot overwrite pack configuration
+- Init container logs show which path was taken (override vs fallback)
+- Contract is explicit and documented
+
+**Bad:**
+- Pack authors must create a ConfigMap with the exact name/namespace/key
+- Silent fallback to `{}` if ConfigMap is misconfigured (mitigated by init container logging)
+- Additional complexity compared to a single ConfigMap
```
Adam-D-Lewis, line 52:
```suggestion
- Additional complexity compared to a single ConfigMap
- Only one software pack can own the override ConfigMap; supporting multiple collector-customizing packs is an explicit non-goal (see [Single-Owner Constraint](#single-owner-constraint))
```

docs/adr/0005-otel-collector-software-pack-override-point.md
```diff
@@ -0,0 +1,156 @@
...
+```bash
+kubectl logs -n monitoring <collector-pod> -c ensure-overrides
+```
+
+Expected output when override is applied:
+```
+ensure-overrides: found /src/relay.yaml from software pack ConfigMap, using override config
+```
+
+## Options Detail
```
Adam-D-Lewis, line 158: Nit: Document that only a single OTEL backend is supported.  Pretty wordy, so feel free to ignore or make more concise.

```suggestion
### Single-Owner Constraint

**Supporting more than one collector-customizing software pack is an explicit
non-goal of this design.** The override point is a **singleton**: it is keyed to
one fixed ConfigMap (`opentelemetry-collector-overrides`) and the collector
merges exactly two configs (base + this override). Exactly **one** pack may own
the override.

This is an unusual case in practice — a cluster has a single observability
backend — so the contract optimizes for that rather than for composition. For
the record, here is what happens if two packs both try to customize the
collector:

- **Resource collision.** Each pack is its own ArgoCD Application, so both
  render a ConfigMap with the same name/namespace. ArgoCD raises a
  `SharedResourceWarning` and, with `selfHeal: true`, the two Applications
  thrash — each sync reverts the other's `relay.yaml`. (Via plain Helm, the
  second `helm install` fails: the ConfigMap already exists and is owned by
  another release.)
- **Silent pipeline clobbering (if forced to coexist).** `confmap` deep-merges
  maps but **replaces lists** — the last `--config` source wins (see
  [Merge Behavior](#merge-behavior)). Exporter *definitions* from both packs
  union fine under `exporters:`, but a pipeline's `exporters:` **list** (e.g.
  `service.pipelines.metrics.exporters`) keeps only the last writer, so one
  pack's backend is defined but never wired into the pipeline and receives no
  telemetry — with no error. The `ensure-overrides` init container cannot
  detect this; it is only visible at the ArgoCD layer.

**Guidance:**

- Reserve the override for the **single telemetry-backend pack** (e.g.
  `nebari-lgtm-pack`) that defines *where* telemetry is exported.
- **Producer** packs that merely generate telemetry must use the collector's
  ingest paths instead: send OTLP to the collector's `4317`/`4318` endpoints,
  or annotate pods with `prometheus.io/scrape`. They must never write the
  override ConfigMap.

Supporting multiple collector-customizing packs is therefore **out of scope**.
If it ever becomes a real requirement, the contract would have to change — each
pack creating a uniquely-named override ConfigMap, the collector loading all of
them via additional `--config` flags (a controller regenerating the daemonset's
volume mounts and args dynamically), and the `confmap.enableMergeAppendOption`
feature gate enabled so component lists append-and-dedupe instead of last-wins.
That work is deferred until a concrete multi-backend need exists.

## Options Detail
```

Adam-D-Lewis (commented):
**Follow-up — design scope.** Multi-pack collector customization is an unusual case (a cluster has one observability backend), and this design intentionally does **not** support it. Two packs that both customize the collector collide rather than compose:

- **Resource collision:** both are separate ArgoCD Applications rendering the same-named override ConfigMap → `SharedResourceWarning` + `selfHeal` thrash (each sync reverts the other); via plain Helm the second install just fails (`already exists, owned by another release`).
- **Silent pipeline clobbering (if forced to coexist):** confmap deep-merges maps but **replaces lists**, so a pipeline's `exporters:` list keeps only the last writer — one pack's backend ends up defined but never wired in, silently receiving no telemetry.

Rather than build for it, I've suggested documenting this and marking it an **explicit non-goal** in the ADR (inline below): a bullet in the **Bad:** consequences, and a new **Single-Owner Constraint** subsection recording the failure behavior, the single-owner guidance (telemetry-backend pack owns the override; producer packs use OTLP/scrape ingest), and what real multi-backend support would take if it's ever needed.

Adam-D-Lewis:
A few of the issues surfaced in review only bite at runtime — not in static review or the unit test. Specifically, the nested `/conf/overrides` mount (likely a `CreateContainerError` on a real kubelet) and the `$$`-over-escaping (the cAdvisor/kubelet jobs scrape nothing). And `TestWriteApplication_OtelCollector_OverridesExtensionPoint` is a structural string-match that would still pass with both of those present.

So before merging, I'd suggest we actually deploy this to a cluster and confirm it works end-to-end rather than relying on review alone:

- [ ] Collector daemonset pods reach `Running` — rules out the read-only-mount `CreateContainerError`
- [ ] `ensure-overrides` init container logs the expected line and the override merges (e.g. the override's exporter shows up in the collector's effective config)
- [ ] `kubernetes-cadvisor` and `kubernetes-kubelet` targets are **up** with non-zero samples — confirms the node-name relabel actually matches
- [ ] End-to-end: with the LGTM pack creating the override ConfigMap, `container_cpu_*` / `container_memory_*` land in Mimir and the dashboards populate (needs the pack-side wiring too)

docs/adr/0005-otel-collector-software-pack-override-point.md
tylerpotts: Fixed: Applied your suggested wording. The merge-behavior paragraph now states that maps deep-merge but lists are replaced wholesale by the last `--config` source, notes the `confmap.enableMergeAppendOption` gate is not enabled, and covers the `null`-deletes-component gotcha with a link to the confmap merge docs.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
tylerpotts: Fixed: Applied the suggestion — changed to single-`$` `${env:K8S_NODE_NAME}` so the confmap resolver expands it to the actual node name before Prometheus sees the regex. Left `$${1}` in the `__metrics_path__` replacement as-is since that one must reach Prometheus literally.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
tylerpotts: Fixed: Same fix as the cadvisor job — single-`$` `${env:K8S_NODE_NAME}` so OTel expands it to the node name and the `keep` rule actually matches the pod's own node.

docs/adr/0005-otel-collector-software-pack-override-point.md
tylerpotts: Fixed: Applied the suggestion — added the cons bullet noting only one pack can own the override ConfigMap, linking to the new Single-Owner Constraint section.

docs/adr/0005-otel-collector-software-pack-override-point.md
tylerpotts: Fixed: Applied the suggested Single-Owner Constraint section verbatim — it documents the singleton nature of the override point, what happens if two packs collide (ArgoCD SharedResourceWarning/thrash, silent pipeline clobbering from list replacement), guidance for producer vs backend packs, and what a multi-pack design would require if it ever becomes a real need.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
```diff
@@ -27,13 +27,39 @@ spec:
...
+        # approach for scraping kubelet metrics.
+        clusterRole:
+          create: true
+          rules:
+            - apiGroups: [""]
+              resources: ["pods", "nodes", "services", "endpoints"]
+              verbs: ["get", "list", "watch"]
+            - apiGroups: [""]
+              resources: ["nodes/metrics", "nodes/proxy"]
+              verbs: ["get"]
```
pmeier: Superseded by #387

pkg/argocd/templates/apps/opentelemetry-collector.yaml
```diff
@@ -27,13 +27,39 @@ spec:
...
             cpu: 100m
             memory: 128Mi
           limits:
             cpu: 500m
             memory: 512Mi
+        extraEnvs:
+          - name: K8S_NODE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
```
pmeier: Might be superseded by #387 as well but needs checking.

docs/adr/0005-otel-collector-software-pack-override-point.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Single-Owner Constraint
+
+**Supporting more than one collector-customizing software pack is an explicit
+non-goal of this design.** The override point is a **singleton**: it is keyed to
+one fixed ConfigMap (`opentelemetry-collector-overrides`) and the collector
+merges exactly two configs (base + this override). Exactly **one** pack may own
+the override.
+
+This is an unusual case in practice — a cluster has a single observability
```
pmeier:
```suggestion
This is an usual case in practice — a cluster has a single observability
```

or do I misread this?

docs/adr/0005-otel-collector-software-pack-override-point.md
pmeier, line 1: Is there any official guidance of where packs should send their data to? Meaning if I have the chat pack and the LTGM pack installed, should the chat apps send their traces to OTEL collector which then forwards to LGTM or should it connect to the LGTM stack directly?

pkg/argocd/templates/apps/opentelemetry-collector.yaml
tylerpotts: Not fully superseded — checked the chart (v0.143.0): the `kubernetesAttributes` preset grants only `pods`/`namespaces` (+`replicasets`), not `nodes` or `nodes/proxy`, and custom `clusterRole.rules` are *merged with* the preset rules rather than replacing them. The cAdvisor/kubelet jobs need `nodes` (for `role: node` SD) and `nodes/proxy` (apiserver-proxy scraping), so the block is still required.

But you're right that part of it was redundant: dropped `pods` (preset covers it) and `services`/`endpoints`/`nodes/metrics` (unused speculative perms), slimming the block to just `nodes` + `nodes/proxy`. Done in 4d1dab8.

pkg/argocd/templates/apps/opentelemetry-collector.yaml
tylerpotts: Confirmed superseded — verified in the chart's `_pod.tpl`: the `kubernetesAttributes` preset injects `K8S_NODE_NAME` via `fieldRef: spec.nodeName` whenever `kubernetesAttributes.enabled && mode == daemonset` (both true here). Removed the duplicate `extraEnvs` block; the `${env:K8S_NODE_NAME}` relabel rules now resolve off the preset-injected var. Added a comment on the preset documenting the dependency. Done in 4d1dab8.

docs/adr/0005-otel-collector-software-pack-override-point.md
tylerpotts: Good catch — you read it right, "unusual" was backwards (a single backend is exactly the case we optimize for). Fixed, reworded to "This is the common case in practice" to also avoid the "a/an usual" article issue. Done in 4d1dab8.

docs/adr/0005-otel-collector-software-pack-override-point.md
tylerpotts: Yes — the ADR's [Single-Owner Constraint](https://github.com/nebari-dev/nebari-infrastructure-core/blob/otel-lgtm-configmap-handoff/docs/adr/0005-otel-collector-software-pack-override-point.md#single-owner-constraint) section covers this. Producer packs (e.g. the chat pack) send telemetry *to* the collector — OTLP on `4317`/`4318`, or `prometheus.io/scrape` annotations — and the collector forwards to the backend. They must never connect to LGTM directly or write the override ConfigMap. The override is reserved for the single backend pack (LGTM) that defines *where* telemetry is exported. So: chat → collector → LGTM, not chat → LGTM.

pmeier (approved):

viniciusdc (approved):
LGTM, thanks @tylerpotts @Adam-D-Lewis @pmeier

Issue #31, opened 2026-01-08
Keycloak needs to be run on FIPS compliant software


tylerpotts:
### Feature description

Keycloak uses encryption algorithms that are not compliant with FIPS by default. Modify the keycloak deployment so that it is FIPS compliant.

See https://www.keycloak.org/server/fips

May want to enable specifying a custom docker image for keycloak

PR #335, opened 2026-05-28
feat(aws): add enable_irsa config field to skip EKS OIDC provider

oren-openteams:
## Status

**Ready for review.** The blocker (terraform-aws-eks-cluster v0.5.0) shipped on 2026-05-28 carrying [nebari-dev/terraform-aws-eks-cluster#31](https://github.com/nebari-dev/terraform-aws-eks-cluster/pull/31) — the `enable_irsa` variable this PR plumbs through. The module pin in this PR is already at the registry-tagged `v0.5.0`, so no source-tree edits or branch refs are needed.

Behavior was field-tested in a real cluster for the past week via the source-tree workaround described under "Verification on a live deployment" below: `terraform-aws-eks-cluster` source pointed at PR #31's branch, hardcoded `enable_irsa = false` in `main.tf`. The cluster came up successfully, no OIDC provider, no `tls_certificate` fetch, full deploy through Argo + foundational apps. The only un-validated path is "config field flows through to terraform without source patches" — that's what this PR provides.

## Summary

Adds a new `cluster.aws.enable_irsa` YAML knob (`*bool`, optional). When unset, the upstream module default (`true`) applies and existing deployments are unaffected. Set `enable_irsa: false` to skip creation of the EKS OIDC provider entirely.

Two motivating scenarios:

1. **EKS Pod Identity-only deployments.** The wrapper module already uses Pod Identity for the EBS CSI driver, EFS CSI driver, and AWS Load Balancer Controller. When the cluster doesn't need IRSA for any additional workloads, the OIDC provider is pure dead weight — extra IAM resource, extra `tls_certificate` fetch on every apply, destroy-ordering hazard at teardown.

2. **Private deployments where DNS can't resolve `oidc.eks.<region>.amazonaws.com`.** In a fully-private VPC (no IGW, no NAT, egress via a central network appliance), the VPC resolver returns NXDOMAIN for that hostname — AWS doesn't ship a VPC interface endpoint for the EKS OIDC host. The upstream module fetches the OIDC issuer's certificate thumbprint via `data "tls_certificate"`, which dies with `dial tcp: lookup oidc.eks.<region>.amazonaws.com on <vpc-resolver>:53: no such host`. The cluster, node groups, and addons all create successfully; the deploy aborts only at the thumbprint fetch — recovery requires reading the module source and either disabling IRSA out-of-band or working around the DNS gap.

```yaml
cluster:
  aws:
    region: us-gov-west-1
    kubernetes_version: "1.34"
    enable_irsa: false   # new — skip OIDC provider creation
    # ...
```

## Changes

- `pkg/provider/aws/config.go`: add `EnableIRSA *bool` to the AWS `Config` struct
- `pkg/provider/aws/tofu.go`: add `EnableIRSA *bool` to `TFVars` (`omitempty`); only set when the user provided a value, so the upstream default propagates when unset
- `pkg/provider/aws/templates/variables.tf`: declare `variable "enable_irsa"` with default `true` (mirrors upstream)
- `pkg/provider/aws/templates/main.tf`: pass `enable_irsa = var.enable_irsa`; bump module pin `0.4.0` → `0.5.0`
- `pkg/provider/aws/tofu_test.go`: cover the three cases (unset, explicit false, explicit true)

## Test plan

- [x] `go test ./pkg/provider/aws/...` — passes (new + existing)
- [x] `go vet ./pkg/provider/aws/` — clean
- [x] Field-tested via the workaround documented in "Verification on a live deployment" — cluster comes up with no OIDC provider, no `tls_certificate` fetch
- [ ] End-to-end `nic deploy` with `enable_irsa: false` on this PR's exact code (registry-pinned v0.5.0, no source patches) — confirms the config-driven path works without manual main.tf edits. This is the last delta vs. the field-tested workaround.

## Verification on a live deployment

Behavior was verified end-to-end against a real EKS cluster by patching `pkg/provider/aws/templates/main.tf` to source `terraform-aws-eks-cluster` directly from #31's branch and hardcoding `enable_irsa = false`. With that combination:

- `Plan: 0 to add, 1 to change, 0 to destroy` (only an unrelated CloudWatch log group drift)
- No `tls_certificate` data source in the refresh
- No `aws_iam_openid_connect_provider` in the plan
- `Apply complete!` — deploy converged past the OIDC fetch that previously aborted the run

This PR is the productionized version of that workaround: a config knob instead of a source-tree edit.

### DIFF
--- a/pkg/provider/aws/config.go
+++ b/pkg/provider/aws/config.go
@@ -41,6 +41,12 @@ type Config struct {
 	// field once trust-manager (the in-pod half of nebari-dev/nebari-infrastructure-core#307)
 	// lands; keeping it provider-scoped here matches the current Provider interface.
 	TrustBundle *TrustBundleConfig `yaml:"trust_bundle,omitempty"`
+	// EnableIRSA toggles creation of the EKS OIDC provider for IAM Roles for
+	// Service Accounts. When unset, the upstream module default (true) applies.
+	// Set false when the cluster relies exclusively on EKS Pod Identity, or
+	// when the VPC cannot resolve oidc.eks.<region>.amazonaws.com (a fully
+	// private deployment with no public DNS resolution for AWS hostnames).
+	EnableIRSA *bool `yaml:"enable_irsa,omitempty"`
 }
 
 // TrustBundleConfig specifies the source of an extra CA bundle. Exactly one of
--- a/pkg/provider/aws/templates/main.tf
+++ b/pkg/provider/aws/templates/main.tf
@@ -31,4 +31,5 @@ module "eks_cluster" {
   efs_kms_key_arn                          = var.efs_kms_key_arn
   node_security_group_additional_rules     = var.node_security_group_additional_rules
   extra_ca_bundle                          = var.extra_ca_bundle
+  enable_irsa                              = var.enable_irsa
 }
--- a/pkg/provider/aws/templates/variables.tf
+++ b/pkg/provider/aws/templates/variables.tf
@@ -132,3 +132,8 @@ variable "enable_cluster_autoscaler_pod_identity" {
   type    = bool
   default = true
 }
+
+variable "enable_irsa" {
+  type    = bool
+  default = true
+}
--- a/pkg/provider/aws/tofu.go
+++ b/pkg/provider/aws/tofu.go
@@ -45,7 +45,8 @@ type TFVars struct {
 	ExtraCABundle                 *string              `json:"extra_ca_bundle,omitempty"`
 	// No omitempty: a false value must be emitted so it overrides the module's
 	// `true` default when the autoscaler is disabled.
-	EnableClusterAutoscalerPodIdentity bool `json:"enable_cluster_autoscaler_pod_identity"`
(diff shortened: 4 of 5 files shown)


dcmcand (approved):
Approving. Clean, correctly-scoped change: it follows the established `*bool` + `omitempty` + only-set-the-TFVar-when-the-user-provided-a-value pattern, so the upstream module default (`true`) is preserved when `enable_irsa` is unset. Good 3-case test coverage (unset/false/true) and a well-documented motivation (EKS Pod-Identity-only clusters, and fully private VPCs that can't resolve `oidc.eks.<region>.amazonaws.com`).

Transparency: I rebased this branch onto `main` to clear the conflict with #352. The resolution kept the module pin at main's `0.6.0` (which already carries `enable_irsa` from upstream v0.5.0 / PR #31, so the bump in this PR was superseded), and unioned the two new TFVars fields and terraform variables (`enable_cluster_autoscaler_pod_identity` + `enable_irsa`). The lockfile regenerated drift-clean. `go build`, `go vet`, `go test ./...`, and the drift regeneration loop all pass locally.

Issue #307, opened 2026-05-20
Epic: Enterprise CA bundle propagation for TLS-inspected egress


dcmcand:
## Goal

Make a NIC deploy succeed end-to-end in environments where all outbound HTTPS is intercepted by an enterprise TLS-inspecting proxy that presents an org-CA-signed certificate. Scope and problem statement are defined in nebari-dev/nebari-infrastructure-core#277.

## Proposed solution

Two complementary mechanisms, plus a small config surface to drive them.

1. **Node-level trust:** the org CA is installed into the OS trust store on every worker node via the node bootstrap script (EKS launch template `userData` on AWS). This covers containerd image pulls, kubelet, and any Go-based workload that reads the system store. Without this, the cluster never bootstraps because ImagePullBackOff happens before anything else gets a chance to run.

2. **In-pod trust:** cert-manager `trust-manager` is deployed as a foundational app, configured with a `Bundle` resource sourcing the org CA. trust-manager projects the bundle as a ConfigMap into each target namespace. Foundational apps that do not consult the OS trust store (ArgoCD repo-server, Keycloak, anything Python/Node/JVM) get the bundle mounted and the relevant env vars set (`REQUESTS_CA_BUNDLE`, `SSL_CERT_FILE`, `NODE_EXTRA_CA_CERTS`, etc.) via Helm value overrides.

3. **Config surface:** a top-level `trust_bundle` field (path on operator's machine, inline PEM, or existing K8s secret ref) drives both of the above. Shape as proposed in #277.

The operator host running `nic` is expected to have the CA installed at the OS level by the operator's normal workstation provisioning, since Go reads system trust on Linux and macOS.

## Scope

Tracked as child issues (populated after creation):

- [ ] Config schema: `trust_bundle` field, parsing, validation
- [ ] AWS node bootstrap: inject CA into EKS launch template userData (lives in `terraform-aws-eks-cluster`)
- [ ] Deploy trust-manager as a foundational app and create the org-CA `Bundle`
- [ ] Patch ArgoCD foundational app to consume the projected bundle
- [ ] Patch Keycloak foundational app to consume the projected bundle
- [ ] JupyterHub spawner integration (lives in `nebari-data-science-pack`)
- [ ] Integration verification against a mock TLS-inspecting proxy
- [ ] Documentation under `docs/operations/`

## Node trust for non-AWS providers

[...]

PR #336, opened 2026-05-28
feat(aws): plumb trust_bundle config into terraform-aws-eks-cluster

### LINKED ISSUE #308: Add trust_bundle config field
## Goal

Add the `trust_bundle` top-level config field that the rest of the CA propagation work keys off of.

## Proposed shape

Per #277:

```yaml
trust_bundle:
  ca_certificates: /path/to/org-ca-bundle.pem
  # OR:
  # inline_pem: |
  #   -----BEGIN CERTIFICATE-----
  #   ...
  # OR:
  # secret_ref:
  #   name: org-ca-bundle
  #   namespace: kube-system
```

Exactly one of the three forms is required when the block is present.

This issue also covers the small AWS-side variable plumbing through `pkg/provider/aws/templates/main.tf` so the bundle reaches the EKS module once nebari-dev/terraform-aws-eks-cluster lands the matching variable.

## Definition of done

- `TrustBundle` struct in `pkg/config/` with parsing for all three forms.
- Validation: file path exists and is readable, inline PEM is parseable, `secret_ref` has both fields. Validation surfaces from `nic validate`.
- Table-driven unit tests for each form and each failure mode.
- AWS template wired to pass the bundle to the upstream module variable.
- Example config snippet in `examples/`.

## Parent

#307

### LINKED ISSUE #334: Wire trust_bundle config into AWS provider and pass it to terraform-aws-eks-cluster as extra_ca_bundle
## Why

The AWS half of node-level trust from #307. EKS worker nodes need the org CA in the OS trust store before kubelet/containerd start, otherwise image pulls fail through a TLS-inspecting proxy and the cluster never bootstraps. The upstream module support is in nebari-dev/terraform-aws-eks-cluster#32 (closes terraform-aws-eks-cluster#29).

## Changes

- Add a top-level `trust_bundle` config field per the shape proposed in #307 (path on operator's machine, inline PEM, or existing k8s secret ref). For this issue, only the base64-PEM-string output matters; the in-pod / trust-manager half is a separate child.
- Resolve `trust_bundle` to a base64-encoded PEM string at config-load time so providers consume a single normalized form.
- AWS provider:
  - Add `ExtraCABundle *string` to `TFVars` in `pkg/provider/aws/tofu.go`, populated only when the bundle is set (so `omitempty` keeps it out of tfvars otherwise).
  - Add `extra_ca_bundle` variable to `pkg/provider/aws/templates/variables.tf` and forward it to the `cluster` module in `templates/main.tf`.
  - Bump the `terraform-aws-eks-cluster` module ref in `templates/main.tf` to the tag containing terraform-aws-eks-cluster#32.

## Tests

- `toTFVars`: unset -> field omitted from marshaled JSON; set -> base64 string present.
- Config parse round-trip for the new top-level field, including inline-PEM and file-path resolution branches.

## Definition of done

- With `trust_bundle` set in `config.yaml`, `nic deploy` against AWS produces an EKS launch template whose userData installs the cert (AL2023 cloud-init pre-nodeadm or Bottlerocket `settings.pki.org-ca`, depending on `ami_type`).
- Unset -> no field in tfvars, no diff in any launch template on subsequent deploys.
- Behavior verified for at least one AL2023 node group; Bottlerocket path covered by module-level tests in terraform-aws-eks-cluster#32.

## Depends on

- nebari-dev/terraform-aws-eks-cluster#32 merged and tagged before the module ref can be bumped to a stable release. For end-to-end validation during development the module ref can point at the `feat/extra-ca-bundle` branch.

## Parent

- #307

dcmcand:
**Draft - will move to ready-for-review once nebari-dev/terraform-aws-eks-cluster#32 is merged and tagged.**

closes  #308
 
While the upstream module PR is open, the module source in `pkg/provider/aws/templates/main.tf` points at the `feat/extra-ca-bundle` branch on `terraform-aws-eks-cluster` instead of the registry. Before this PR lands, that source needs to be reverted to the registry form with the tagged version that includes #32.

## Summary

- Adds an optional `cluster.aws.trust_bundle` config block (`path` or `inline` PEM) and resolves it to a base64-encoded PEM string.
- Forwards it to the EKS cluster module as `extra_ca_bundle`, so the bundle is installed into the OS trust store of every EKS worker node before kubelet starts (required for clusters whose egress runs through a TLS-inspecting proxy).
- The temporary branch-based module source lets us test end-to-end ahead of the upstream merge.

## Test plan

- [x] `go test ./pkg/provider/aws/...` covering both `TrustBundle.ResolveBase64` (path / inline / both-set error / missing-file error / no-PEM error / nil no-op) and the `toTFVars` wiring (omitted vs base64 round-trip).
- [x] End-to-end on AWS: deployed a real EKS cluster (`dcmcand-333` in `us-west-2`) with a test CA passed via `trust_bundle.path`. Verified the cert lands in `/etc/pki/ca-trust/source/anchors/org-ca.crt` and the extracted bundle on the running AL2023 node, fingerprint match.
- [x] Pending: bump the module source back to the registry with the tagged version after [#32](https://github.com/nebari-dev/terraform-aws-eks-cluster/pull/32) merges.

Closes #334. Refs #307.


dcmcand:
Tested end-to-end on AWS.

Deployed `dcmcand-333` in `us-west-2` with this branch's `nic` binary against a config that sets:

```yaml
cluster:
  aws:
    trust_bundle:
      path: /tmp/nic-ca-test/ca.pem
```

`ca.pem` was a self-signed cert generated with `openssl req -x509 ... -subj "/CN=NIC-Trust-Bundle-Test-1779968788"` and SHA256 fingerprint `9A:10:47:AF:E7:F4:18:89:9A:25:92:7A:7B:01:65:2B:21:1A:91:5A:C5:95:E0:4D:33:5E:C1:F6:06:44:C7:02`.

On the running AL2023 node:

- `/etc/pki/ca-trust/source/anchors/org-ca.crt` matches the generated cert byte for byte (same CN, same fingerprint), so the cloud-init pre-nodeadm script wrote the bundle as expected.
- `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` contains a cert with that same fingerprint, so `update-ca-trust extract` ran before kubelet/containerd came up.
- `/etc/pki/tls/certs/ca-bundle.crt` is a symlink to the extracted bundle, so anything reading the AL2023 system trust store sees the org CA - containerd image pulls, kubelet, and any Go binary using crypto/x509.

Bottlerocket path was not exercised in this deploy (no Bottlerocket node group); coverage there comes from the module-level `tofu test` in nebari-dev/terraform-aws-eks-cluster#32.

pkg/provider/aws/config.go
```diff
@@ -152,3 +172,40 @@ func (c *Config) EFSStorageClassName() string {
...
+	if pathSet {
+		data, err := os.ReadFile(t.Path)
+		if err != nil {
+			return "", fmt.Errorf("trust_bundle: read %s: %w", t.Path, err)
+		}
+		pem = data
+	} else {
+		pem = []byte(t.Inline)
+	}
+	if !strings.Contains(string(pem), "-----BEGIN CERTIFICATE-----") {
```
tylerpotts, line 201: **Question:** PEM validation here is a substring check — it accepts anything containing the `-----BEGIN CERTIFICATE-----` literal without parsing the cert. `encoding/pem.Decode` (stdlib) would validate structure. A tell: the test fixture `samplePEM` isn't a real certificate, which a stricter check would have forced. Given the module re-validates base64 and `update-ca-trust` rejects junk on the node, shallow may be intentional — just flagging it as a deliberate call.

pkg/provider/aws/config.go
```diff
@@ -152,3 +172,40 @@ func (c *Config) EFSStorageClassName() string {
...
+
+// ResolveBase64 returns the configured CA bundle as a base64-encoded PEM string,
+// suitable for passing straight to the terraform-aws-eks-cluster module's
+// extra_ca_bundle input. Returns an empty string when the bundle is unset.
+func (t *TrustBundleConfig) ResolveBase64() (string, error) {
+	if t == nil {
+		return "", nil
+	}
+	pathSet := t.Path != ""
+	inlineSet := strings.TrimSpace(t.Inline) != ""
```
tylerpotts, line 184: **Stylistic:** asymmetric emptiness checks — `inlineSet` trims whitespace but `pathSet` (line 183) uses a bare `!= ""`, so a whitespace-only path is treated as set and fails later at read. Minor.

pkg/provider/aws/config.go
```diff
@@ -152,3 +172,40 @@ func (c *Config) EFSStorageClassName() string {
...
+		data, err := os.ReadFile(t.Path)
+		if err != nil {
+			return "", fmt.Errorf("trust_bundle: read %s: %w", t.Path, err)
+		}
+		pem = data
+	} else {
+		pem = []byte(t.Inline)
+	}
+	if !strings.Contains(string(pem), "-----BEGIN CERTIFICATE-----") {
+		return "", fmt.Errorf("trust_bundle: no PEM certificate found in %s",
```
tylerpotts, line 202: **Stylistic:** the inline `func() string {...}()` to pick the error subject reads awkwardly; computing a local `subject` variable before the `if` would be clearer.

pkg/provider/aws/tofu.go
```diff
@@ -126,6 +127,14 @@ func (c *Config) toTFVars(projectName string) TFVars {
 		}
 	}
 
+	caBundle, err := c.TrustBundle.ResolveBase64()
```
tylerpotts, line 130: **Question:** `toTFVars` now does disk I/O via `ResolveBase64`, which also runs the mutual-exclusion / missing-file / no-PEM validation. Those checks only fire at Deploy/Destroy — `nic validate` won't catch a misconfigured `trust_bundle`, even though sibling checks (LB scheme, CIDR, node groups) live in `Provider.Validate`. Lifting the resolve/validate into `Provider.Validate` would surface errors at validate time and keep `toTFVars` pure. Ambiguous because the file read is arguably a deploy-time side effect — your call.

pkg/provider/aws/templates/variables.tf
```diff
@@ -122,3 +122,8 @@ variable "node_security_group_additional_rules" {
   type    = any
   default = {}
 }
+
+variable "extra_ca_bundle" {
```
tylerpotts, line 126: Matches the file's conventions exactly (`string`, `default = null`, no description like the other variables). Clean passthrough.

pkg/provider/aws/templates/main.tf
```diff
@@ -29,4 +29,5 @@ module "eks_cluster" {
   efs_encrypted                            = var.efs_encrypted
   efs_kms_key_arn                          = var.efs_kms_key_arn
   node_security_group_additional_rules     = var.node_security_group_additional_rules
+  extra_ca_bundle                          = var.extra_ca_bundle
```
tylerpotts, line 32: Clean single-line passthrough; the module bump to `0.5.0` picks up the upstream `extra_ca_bundle` input.

tylerpotts (approved):
## Architecture + Go + IaC review — Approved

Approving. No blockers across architecture, Go, or Terraform. Nice work — the `toTFVars` error threading is clean, `ResolveBase64` is nil-safe, and the table tests cover every branch (nil / empty / both-set / inline / path / missing-file / no-PEM, plus the tfvars round-trip).

**One theme worth your attention — validation placement.** The `trust_bundle` resolve + validation currently runs inside `toTFVars` at Deploy/Destroy time, so `nic validate` gives a misconfigured bundle a clean bill of health even though sibling config checks (LB scheme, CIDR, node groups) live in `Provider.Validate`. Moving the resolve/validate into `Provider.Validate` would surface errors earlier and keep `toTFVars` pure. Flagged inline — non-blocking, but the most substantive point.

**Provider-scoping vs epic #307.** Placing the field at `cluster.aws.trust_bundle` is the right call for this slice and violates no current boundary — it genuinely is an AWS module input, and the godoc already notes the likely future move to top-level. Heads-up only: epic #307 also needs the same bundle delivered in-pod via trust-manager across providers, so an operator on a multi-provider story would eventually configure the same PEM twice. Worth capturing the migration path (a top-level field the AWS provider reads as a fallback) in #307 so this field can be deprecated cleanly.

Issue resolution: #308 and #334 fully resolved; #307 partial by design (node-level half only).

Minor note: the PR description is now stale — it still says `main.tf` points at the `feat/extra-ca-bundle` branch and needs reverting to the registry form, but that's already done (registry `0.5.0`).

pkg/provider/aws/templates/main.tf
```diff
@@ -1,6 +1,6 @@
 module "eks_cluster" {
   source  = "nebari-dev/eks-cluster/aws"
-  version = "0.4.0"
+  version = "0.5.0"
```
viniciusdc, line 3: This is already flipped to the registry source at `version = "0.5.0"` — which is the tag that actually shipped `extra_ca_bundle` (terraform-aws-eks-cluster v0.5.0) 

Heads up that the PR description is now stale relative to the code: it still says the module source "points at the `feat/extra-ca-bundle` branch ... instead of the registry" and lists "bump the module source back to the registry with the tagged version" as a Pending item in the test plan. That's done as of `d640d9c`. 

Worth updating the body and ticking that box so reviewers know the upstream-merge blocker is already cleared

viniciusdc:
Going to merge this one due to the tracked itens above and approvals, will create a follow up issue afterwards with low priority and nice to have fixes mentioned.

PR #346, opened 2026-06-01
feat(argocd): deploy trust-manager and project org CA bundle in-pod

tylerpotts:
## Summary

Closes https://github.com/nebari-dev/nebari-infrastructure-core/issues/309. This is the in-pod half of epic https://github.com/nebari-dev/nebari-infrastructure-core/issues/307 (the node-level half shipped in #336 via `extra_ca_bundle`).

- Deploys cert-manager **trust-manager** as a foundational ArgoCD app (`pkg/argocd/templates/apps/trust-manager.yaml`, Helm chart `v0.22.1` from `https://charts.jetstack.io`, sync-wave 3).
- Renders a **Bundle** (`trust.cert-manager.io/v1alpha1`) that sources the org CA inline and projects it as a ConfigMap (`ca-certificates.crt`) into every namespace via an empty `namespaceSelector` (sync-wave 4).
- Adds a **top-level `trust_bundle`** config (`path` or inline PEM) in `pkg/config/trust_bundle.go` with `ResolvePEM`/`ResolveBase64`/`Validate`.
- Plumbs the resolved bundle to providers via `Deploy`/`DestroyOptions` (the provider boundary deliberately doesn't see `NebariConfig`, so resolution happens at the orchestration layer in `pkg/nic`).
- The provider-scoped `cluster.aws.trust_bundle` introduced in #336 is **removed**: the top-level `trust_bundle` key is now the only way to configure the bundle, and it drives both the node-level `extra_ca_bundle` input and in-pod trust via trust-manager.

## Test plan

- [x] `golangci-lint run` on changed packages — 0 issues
- [x] `go test -short` on argocd/config/nic/aws — all pass
- [x] New tests: config resolution/validation, AWS tfvars bundle pass-through, Bundle template renders valid YAML, WriteAllToGit skips/writes trust-manager based on config
- [x] Verify trust-manager Bundle propagates the CA ConfigMap to namespaces in a live cluster


tylerpotts:
## Verified in a local kind cluster

Deployed end-to-end via `make localkind-up` (local provider) with a top-level `trust_bundle` set to a throwaway self-signed CA (`CN = Test Org Root CA`). This exercises the real NIC path: `WriteAllToGit` gates trust-manager on the top-level `trust_bundle` field, renders the app + Bundle manifests into the GitOps repo, and ArgoCD syncs them.

**Results:**

- `WriteAllToGit` rendered `apps/trust-manager.yaml`, `apps/trust-bundle.yaml`, and `manifests/security/trust-bundle/bundle.yaml`. The Bundle is valid YAML with the PEM correctly indented under `inLine: |-` (the `indent` template func works).
- ArgoCD apps both `Synced` / `Healthy`:
  ```
  trust-bundle    Synced  Healthy
  trust-manager   Synced  Healthy
  ```
- trust-manager pod (`v0.22.1`) Running 1/1 in `cert-manager`.
- Bundle `nebari-trust-bundle` status: `Synced: True` — "Successfully synced Bundle to all namespaces".
- The target ConfigMap `nebari-trust-bundle` (data key `ca-certificates.crt`) was projected into **all 13 namespaces** (argocd, cert-manager, default, envoy-gateway-system, keycloak, kube-system, metallb-system, monitoring, nebari-operator-system, nebari-system, …).
- **Content verified**: the projected cert matches the source CA byte-for-byte.

  | | Subject | SHA-256 fingerprint |
  |---|---|---|
  | Source (GitOps repo) | `CN = Test Org Root CA` | `91:61:BF:5A:…:98:95:F2` |
  | Projected (default ns) | `CN = Test Org Root CA` | `91:61:BF:5A:…:98:95:F2` |

**Scope note:** this validates the **in-pod** half — trust-manager projecting the CA as a ConfigMap into every namespace for pods to consume. It does **not** cover the **node-level** OS trust store, which is the EKS `extra_ca_bundle` path from #336 and requires AWS (not exercisable on kind).

pkg/config/trust_bundle.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil, nil
+	}
+
+	var (
+		pem     []byte
+		subject string
+	)
+	if pathSet {
+		subject = t.Path
+		data, err := os.ReadFile(t.Path)
```
dcmcand: **[BLOCKER]** Whitespace-padded path is detected as set but read un-trimmed. Line 61 uses `strings.TrimSpace(t.Path) != ""` to decide the path is set, but this read uses the original `t.Path`. A value like `"  /real/path  "` is treated as set, then `os.ReadFile` is called with the padded string and fails with a confusing `trust_bundle: read   /real/path  : no such file or directory`. The previous `aws/config.go` implementation used `t.Path != ""` (no trim), so this is a new behavior change. Fix: trim once and use the trimmed value for both the set-check and the read. Worth a table case exercising `{Path: "  /real/path  "}` - the current tests only cover the all-whitespace (`"   "` -> unset) case.

dcmcand (commented):
## Review summary

In-pod half of epic #307: deploys cert-manager trust-manager via ArgoCD and projects the org CA ConfigMap into every namespace, promotes `TrustBundleConfig` to a top-level config type (AWS package keeps a backward-compatible alias), and plumbs the resolved bundle through `Deploy`/`DestroyOptions`. Architecture is fully conformant - correct import direction (config knows no provider types), no provider-name switches, instrumentation and the status/logging seam respected.

**Findings:** 1 Blocker, 2 Questions, 7 Stylistic. Inline comments below.

- **Blocker:** whitespace-padded `trust_bundle.path` is detected as set but read un-trimmed (`pkg/config/trust_bundle.go`).
- **Questions** (genuine design calls for you): whether `nic validate` should touch disk, and whether `destroy` should tolerate a missing PEM file.
- **Stylistic:** resolve the PEM once instead of twice per deploy, split structural validation from disk I/O, reject `PRIVATE KEY` blocks defensively, a few small clarity nits.

Nice work overall - the alias-based migration, the precedence short-circuit, and the trailing-newline handling in `indentLines` are all clean.

pkg/config/trust_bundle.go
```diff
@@ -0,0 +1,90 @@
...
+	if len(pem) == 0 {
+		return "", nil
+	}
+	return base64.StdEncoding.EncodeToString(pem), nil
+}
+
+// Validate checks that the bundle is well-formed without returning its contents.
+// Safe to call at validate time so misconfigurations surface before deploy.
+func (t *TrustBundleConfig) Validate() error {
+	_, err := t.resolve()
```
dcmcand: **[QUESTION]** `Validate()` -> `resolve()` -> `os.ReadFile`, so `nic validate` fails when a `path:`-based bundle's file isn't present, even if the config is structurally correct (e.g. in CI, or on a machine that isn't the operator's). Is that intended? It's consistent with effectful validation of some other fields, but if validate should stay environment-independent, the structural-vs-I/O split suggested below is the fix. Flagging because only you know the intended contract for `nic validate`.

pkg/nic/destroy.go
```diff
@@ -121,10 +121,17 @@ func (c *Client) Destroy(ctx context.Context, cfg *config.NebariConfig, opts Des
 		}
 	}
 
+	caBundle, err := cfg.TrustBundle.ResolveBase64()
```
dcmcand, line 130: **[QUESTION]** Destroy now re-resolves the bundle from disk so the destroy plan matches deploy. If a `path:`-based PEM was deleted or moved after deploy, destroy fails here before reaching OpenTofu - even though the applied value already lives in TF state. Should destroy be tolerant of a missing file (e.g. read the value from state outputs), or is requiring the file to remain until destroy an acceptable contract? Genuine design call - whichever way you decide, a note on the config field comment would help operators.

pkg/argocd/writer.go
```diff
@@ -256,14 +264,26 @@ func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.Nebari
 	workDir := gitClient.WorkDir()
 	data := NewTemplateData(cfg, gitConfig, settings)
 
+	// Resolve the top-level trust bundle for trust-manager. Disk I/O lives here
+	// rather than in the pure NewTemplateData constructor.
+	trustPEM, err := cfg.TrustBundle.ResolvePEM()
```
dcmcand: **[STYLISTIC]** A `path:`-based bundle is read from disk twice per deploy: here via `ResolvePEM()` and again in `pkg/nic/deploy.go` via `ResolveBase64()`. Consider resolving once at the orchestration layer and passing the resolved PEM into `WriteAllToGit`. Eliminates the redundant read and closes a small TOCTOU window between the two reads.

pkg/config/trust_bundle.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	if len(pem) == 0 {
+		return "", nil
+	}
+	return base64.StdEncoding.EncodeToString(pem), nil
+}
+
+// Validate checks that the bundle is well-formed without returning its contents.
+// Safe to call at validate time so misconfigurations surface before deploy.
+func (t *TrustBundleConfig) Validate() error {
```
dcmcand, line 60: **[STYLISTIC]** Two small things on `Validate()`:

1. Consider splitting structural validation (mutual exclusion of path/inline + the PEM-marker check) from effectful resolution. If `Validate()` did only the structural check and left `os.ReadFile` to `ResolvePEM`/`ResolveBase64`, validation would be environment-independent and usable in config-linting contexts where the file isn't present. This is also the concrete fix if you decide the QUESTION above (validate touching disk) should be a no.
2. Add an explicit `if t == nil { return nil }` guard at the top to match `ResolvePEM`/`ResolveBase64`. Nil-safety currently lives only in the private `resolve()` helper, so a reader has to follow the call to confirm it.

pkg/config/trust_bundle.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("trust_bundle: read %s: %w", t.Path, err)
+		}
+		pem = data
+	} else {
+		subject = "inline value"
+		pem = []byte(t.Inline)
+	}
+
+	if !strings.Contains(string(pem), "-----BEGIN CERTIFICATE-----") {
```
dcmcand: **[STYLISTIC]** Defense-in-depth: consider rejecting PEM input that contains a private-key block. The resolved bundle lands in OpenTofu state *and* in a ConfigMap projected to every namespace via the GitOps repo, so a user who accidentally points `trust_bundle` at a combined cert+key file would silently distribute the key cluster-wide and commit it to git. A cheap guard alongside the existing CERTIFICATE check, e.g. reject when `strings.Contains(pem, "PRIVATE KEY")`, keeps an honest mistake from having severe consequences.

pkg/argocd/writer.go
```diff
@@ -339,6 +367,35 @@ func isMetalLBPath(relPath string) bool {
...
+func isTrustBundlePath(relPath string) bool {
+	return relPath == "apps/trust-manager.yaml" ||
+		relPath == "apps/trust-bundle.yaml" ||
+		strings.HasPrefix(relPath, "manifests/security/trust-bundle")
+}
+
+// templateFuncs provides the small set of helpers templates may call. indent and
+// nindent mirror the common Helm helpers for embedding multi-line values (e.g. a
+// PEM bundle) at a fixed YAML indentation.
+var templateFuncs = template.FuncMap{
```
dcmcand, line 424: **[STYLISTIC]** `templateFuncs` is package-level but only consumed by `processTemplate`. A one-line comment noting it's the central extension point for template helpers (or moving it adjacent to `processTemplate`) would keep a reader from assuming it's a broader public extension surface.

pkg/argocd/templates/apps/trust-bundle.yaml
```diff
@@ -0,0 +1,36 @@
...
+    path: {{ if .GitPath }}{{ .GitPath }}/{{ end }}manifests/security/trust-bundle
+
+  destination:
+    server: https://kubernetes.default.svc
+
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
+      allowEmpty: false
```
dcmcand, line 33: **[STYLISTIC]** `allowEmpty: false` here means that if template rendering ever produces an empty `manifests/security/trust-bundle/` directory, ArgoCD refuses to prune and could leave a stale Bundle CR pointing at a deleted PEM. That may well be the safer choice, but the intent differs from `trust-manager.yaml` (where a Helm chart is never empty). A short comment on why `allowEmpty: false` is correct for the git-backed Bundle would help future readers.

pkg/argocd/templates/manifests/security/trust-bundle/bundle.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: trust.cert-manager.io/v1alpha1
```
dcmcand, line 5: **[STYLISTIC]** This uses the alpha API `trust.cert-manager.io/v1alpha1`, which can change across trust-manager minor versions. The chart is pinned to `v0.22.1` so there's no immediate risk, but a comment here (or a line in the upgrade runbook) noting the alpha-API dependency would make the coupling explicit when the chart is bumped.

pkg/provider/aws/config.go
```diff
@@ -33,22 +29,19 @@ type Config struct {
...
-// TrustBundleConfig specifies the source of an extra CA bundle. Exactly one of
-// Path or Inline must be set. Path is a filesystem path to a PEM file on the
-// operator's machine; Inline is the PEM text itself.
-type TrustBundleConfig struct {
-	Path   string `yaml:"path,omitempty"`
-	Inline string `yaml:"inline,omitempty"`
-}
+// TrustBundleConfig is an alias for the canonical top-level type. Retained so the
+// deprecated cluster.aws.trust_bundle field continues to unmarshal.
+type TrustBundleConfig = config.TrustBundleConfig
```
dcmcand: Nice - aliasing the AWS type to the canonical `config.TrustBundleConfig` keeps a single source of truth while letting existing `cluster.aws.trust_bundle` configs keep unmarshaling with zero migration. Idiomatic use of a Go type alias for exactly this backward-compat case.

pkg/provider/aws/tofu.go
```diff
@@ -127,10 +127,15 @@ func (c *Config) toTFVars(projectName string) (TFVars, error) {
 		}
 	}
 
+	// Provider-scoped trust_bundle takes precedence; fall back to the top-level
+	// bundle resolved by the orchestration layer.
 	caBundle, err := c.TrustBundle.ResolveBase64()
 	if err != nil {
 		return TFVars{}, err
 	}
+	if caBundle == "" {
```
dcmcand: Good precedence handling: the error from `ResolveBase64()` above returns immediately rather than falling through to `fallbackCABundle`, so a misconfigured explicit value can never silently degrade to a different bundle. The provider-scoped-wins-else-fallback logic is clear and well-commented.

pkg/argocd/writer.go
```diff
@@ -339,6 +367,35 @@ func isMetalLBPath(relPath string) bool {
...
+	"indent":  indentLines,
+	"nindent": func(spaces int, s string) string { return "\n" + indentLines(spaces, s) },
+}
+
+// indentLines prefixes every non-empty line of s with the given number of spaces.
+func indentLines(spaces int, s string) string {
```
dcmcand, line 430: `indentLines` handles the common trailing-newline case correctly: a PEM ending in `\n` yields a trailing empty element from `strings.Split`, which is left unindented, and the `|-` block scalar then strips it - so the rendered `inLine` round-trips as valid YAML. The parse test confirms it.

pkg/argocd/writer.go
```diff
@@ -256,14 +264,26 @@ func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.Nebari
...
+
 	span.SetAttributes(
 		attribute.String("work_dir", workDir),
 		attribute.String("domain", data.Domain),
 		attribute.String("git_repo_url", data.GitRepoURL),
 	)
 
 	// Walk all files in the templates directory
-	err := fs.WalkDir(templates, templateDir, func(path string, d fs.DirEntry, err error) error {
+	err = fs.WalkDir(templates, templateDir, func(path string, d fs.DirEntry, err error) error {
```
dcmcand: Correct switch from `:=` to `=` here - `err` is already declared by the `trustPEM, err :=` resolution above, so this avoids shadowing and there's no stale-err leak into the WalkDir error check. Easy to get wrong; good catch.

pkg/provider/aws/provider.go
```diff
@@ -170,6 +170,19 @@ func (p *Provider) Validate(ctx context.Context, projectName string, clusterConf
...
 
+	// Validate the deprecated provider-scoped trust_bundle here so misconfigs
+	// surface at validate time rather than mid-deploy.
+	if awsCfg.TrustBundle != nil {
+		if err := awsCfg.TrustBundle.Validate(); err != nil {
+			span.RecordError(err)
+			return err
+		}
+		status.Send(ctx, status.NewUpdate(status.LevelWarning,
+			"cluster.aws.trust_bundle is deprecated; set trust_bundle at the top level of the Nebari config instead").
```
dcmcand: Good placement: the deprecation warning is emitted during `Validate` (i.e. `nic validate`), before any infrastructure changes, so operators see it without having to trigger a deploy - and it goes through `status.Send` rather than a direct `slog` call, respecting the package's logging seam.

dcmcand (changes requested):
Requesting changes on the blocker from my detailed review above (#pullrequestreview-4409840945): whitespace-padded `trust_bundle.path` is detected as set but read un-trimmed in `pkg/config/trust_bundle.go`, which produces a confusing read error. The two Questions and the stylistic items there are non-blocking - happy to re-review once the blocker is addressed.

pkg/config/trust_bundle.go
tylerpotts: Fixed in b82d9f6. `resolve()` now trims `Path` once into a local and uses that trimmed value for both the set-check and `os.ReadFile`, so `"  /real/path  "` reads the real file. Added a `{Path: "  "+pemPath+"  "}` table case to `TestTrustBundleResolveBase64`.

pkg/config/trust_bundle.go
tylerpotts: Decided `nic validate` should stay environment-independent. `Validate()` is now structural-only (mutual-exclusion + inline PEM-marker checks) and no longer touches disk, so it works in CI / off the operator's box. The file read and the path PEM-marker check moved to resolve time (deploy/destroy). See the structural/effectful split reply below. b82d9f6.

pkg/nic/destroy.go
tylerpotts: Made destroy tolerant. `destroy.go` now downgrades a missing path: PEM (`fs.ErrNotExist`) to an empty bundle with a warning instead of failing, since the applied value already lives in TF state. Threaded the same tolerance into the deprecated provider-scoped path via a `tolerateMissingBundle` flag on `toTFVars` (true on destroy, false on deploy). b82d9f6.

pkg/argocd/writer.go
tylerpotts: Done. The top-level bundle is now resolved once in `Deploy` (`ResolvePEM`), its base64 form derived for the provider, and the PEM threaded through `bootstrapGitOps` into `WriteAllToGit` via a new `trustBundlePEM` param. `WriteAllToGit` no longer re-reads from disk, closing the TOCTOU window. b82d9f6.

pkg/config/trust_bundle.go
tylerpotts: Both done. (1) `Validate()` is now structural-only and no longer calls `resolve()`/`os.ReadFile`. (2) Added explicit `if t == nil { return nil }` guards to `Validate`, `ResolvePEM`, and `ResolveBase64`. b82d9f6.

pkg/config/trust_bundle.go
tylerpotts: Added. A new `checkPEM` helper rejects input containing a `PRIVATE KEY` block alongside the existing CERTIFICATE check, with a test case covering an inline cert+key bundle. b82d9f6.

pkg/argocd/writer.go
tylerpotts: Added a comment noting `templateFuncs` is the single extension point, consumed only by `processTemplate` rather than a broader public surface. b82d9f6.

pkg/argocd/templates/apps/trust-bundle.yaml
tylerpotts: Added a comment explaining why `allowEmpty: false` is correct for the git-backed Bundle (prevents pruning to a stale CR pointing at a deleted PEM) and contrasting it with the never-empty Helm chart in trust-manager.yaml. b82d9f6.

pkg/argocd/templates/manifests/security/trust-bundle/bundle.yaml
tylerpotts: Added a comment flagging the `v1alpha1` alpha-API dependency and pointing at the chart pin (`v0.22.1`) in apps/trust-manager.yaml to revisit whenever that pin is bumped. b82d9f6.

pkg/argocd/writer.go
tylerpotts: Heads up: the orchestration-layer refactor (resolving the PEM once in `Deploy` and threading it in) removed the earlier `trustPEM, err :=` declaration in `WriteAllToGit`, so this line is back to `err := fs.WalkDir(...)`. `err` is no longer pre-declared at that point, so there is no shadowing or stale-err concern either way now. b82d9f6.

viniciusdc:
#336 is merged to `main` now, so per the Stacking note this is good to retarget from `feat/trust-bundle-aws` to `main`.

pkg/provider/aws/provider.go
dcmcand: **Change requested:** this deprecation warning is in `Validate`, but `nic deploy` calls config-level `cfg.Validate(...)`, not `clusterProvider.Validate`, so it never fires on a plain `nic deploy` - the deprecated `cluster.aws.trust_bundle` is used silently and takes precedence over the top-level. Please emit the deprecation from a path that runs on deploy/destroy (e.g. `toTFVars`, or `Deploy`/`Destroy` after extracting config) so users see the nudge before the field is removed.

pkg/argocd/templates/apps/trust-manager.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: trust-manager
+  namespace: argocd
+  labels:
+    app.kubernetes.io/part-of: nebari-foundational
+    app.kubernetes.io/managed-by: nebari-infrastructure-core
+  annotations:
+    argocd.argoproj.io/sync-wave: "3"
```
dcmcand, line 10: Question (non-blocking): trust-manager's validating webhook needs a cert-manager-issued cert, so it depends on cert-manager being healthy, not just created. Sync-waves order creation (cert-manager wave 2, this wave 3) but not readiness, so on a first deploy the webhook can fail to start until cert-manager finishes issuing, causing a multi-minute retry. ArgoCD's retry recovers it, so this is likely fine - just confirming it's expected rather than something that warrants a readiness gate.

pkg/argocd/templates/manifests/security/trust-bundle/bundle.yaml
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  sources:
+    - inLine: |-
+{{ indent 8 .TrustBundlePEM }}
+  target:
+    configMap:
+      key: ca-certificates.crt
+    # Empty selector projects the bundle ConfigMap into every namespace,
+    # covering foundational and software-pack namespaces alike.
+    namespaceSelector: {}
```
dcmcand, line 29: Question (non-blocking): the empty `namespaceSelector` projects this ConfigMap into every namespace, including `kube-system` and `default`. That's intentional and documented here, but in clusters running OPA/Kyverno policies that restrict writes to system namespaces it can trip violations. Confirming this is a conscious tradeoff for the target audience.

pkg/argocd/templates/manifests/security/trust-bundle/bundle.yaml
```diff
@@ -0,0 +1,21 @@
+# trust.cert-manager.io/v1alpha1 is an alpha API and may change across
+# trust-manager minor versions. It is safe while the chart is pinned (see the
+# pinned version in apps/trust-manager.yaml); revisit this manifest whenever that
+# pin is bumped.
+apiVersion: trust.cert-manager.io/v1alpha1
```
dcmcand, line 5: Minor: confirm `trust.cert-manager.io/v1alpha1` is the API served by trust-manager v0.22.1. It is for the v0.x line (the Bundle CRD hasn't graduated past v1alpha1), so this is just a re-check-when-the-chart-pin-bumps note, which the comment already gestures at.

pkg/nic/destroy.go
```diff
@@ -121,10 +123,27 @@ func (c *Client) Destroy(ctx context.Context, cfg *config.NebariConfig, opts Des
 		}
 	}
 
+	// Re-resolve the bundle so the destroy plan matches what was deployed. The
+	// applied value already lives in TF state, so a source PEM that was deleted
+	// or moved after deploy must not block teardown: downgrade a missing file to
+	// an empty bundle and warn rather than fail.
+	caBundle, err := cfg.TrustBundle.ResolveBase64()
+	if errors.Is(err, fs.ErrNotExist) {
```
dcmcand, line 131: Question (non-blocking): on destroy, a since-deleted bundle file is tolerated and `caBundle` becomes empty, and `toTFVars` tolerates the provider-scoped one too. If the deployed state had an `extra_ca_bundle` and both sources now resolve empty, does OpenTofu's destroy plan care about that diff, or does it only act on the resources being deleted? Just confirming there's no plan/apply mismatch on destroy.

dcmcand (dismissed):
Thanks for this - it's a well-built foundation for the trust_bundle epic, and the security thinking is strong: the `checkPEM` private-key guard (with the why-comment about state/GitOps distribution), resolving the bundle once at the orchestration layer to avoid a TOCTOU window, and the table-tested deprecation precedence matrix. Build, vet, gofmt, and tests all pass.

Requesting changes on two items, both about the deprecated-field migration and the GitOps record rather than the core feature:

1. The `cluster.aws.trust_bundle` deprecation warning only fires in `Validate`, which `nic deploy` doesn't call (deploy uses config-level `cfg.Validate`, not `clusterProvider.Validate`). So the deprecated field is used silently on deploy and even takes precedence over the top-level. Please emit the warning from a path that runs on deploy/destroy.
2. `scrubbedConfig` leaves `trust_bundle` (path or inline) in the committed `nic-config.yaml`; for the path case that bakes a machine-local file path into the GitOps record. Resolve-to-inline or strip it, matching how the auth fields are scrubbed.

Neither is a security issue - the CA is public and `Validate` rejects `PRIVATE KEY` blocks - they're about migration UX and record hygiene.

A few non-blocking questions are inline (trust-manager/cert-manager readiness across sync-waves, `namespaceSelector: {}` into system namespaces, destroy-with-missing-file plan, Bundle `v1alpha1` for the pinned chart). I confirmed the Bundle name/key/namespaceSelector match what the stacked consumers #350 and #353 expect.

Issue #309 is delivered. Happy to re-review once the two items above are addressed.

pkg/provider/aws/provider.go
tylerpotts: Resolved by removing the deprecated field entirely in fb93708. `cluster.aws.trust_bundle`, the precedence/fallback logic in `toTFVars`, the validate-time deprecation warning, and the `tolerateMissingBundle` flag are all gone — the bundle now flows exclusively through the top-level `trust_bundle` key, so there is no deprecation path left to warn about. One caveat worth noting: provider config unmarshals non-strictly, so a config still carrying `cluster.aws.trust_bundle` is now silently ignored rather than erroring. Since this is pre-1.0 and the field only existed on this branch plus #336, that seemed acceptable over keeping a tombstone check.

pkg/argocd/templates/apps/trust-manager.yaml
tylerpotts: Confirmed expected. Sync waves order creation only, and on a first deploy the trust-manager webhook pod can crash-loop briefly until cert-manager finishes issuing its serving cert. ArgoCD's retry/self-heal recovers it without intervention, and nothing downstream consumes the Bundle before wave 4, so a readiness gate did not seem worth the extra machinery. Will revisit if first-deploy convergence time becomes a real complaint.

pkg/argocd/templates/manifests/security/trust-bundle/bundle.yaml
tylerpotts: Yes, conscious tradeoff. The target audience is enterprises behind TLS-inspecting egress proxies where any pod in any namespace may need the org CA, and predicting which namespaces is harder than projecting everywhere. Clusters with OPA/Kyverno policies restricting system-namespace writes would need to exempt the trust-manager ServiceAccount or narrow the selector; if that becomes a common ask we can expose `namespaceSelector` as config later without breaking the default.

pkg/argocd/templates/manifests/security/trust-bundle/bundle.yaml
tylerpotts: Confirmed: trust-manager v0.22.x serves Bundle exclusively at `trust.cert-manager.io/v1alpha1` — the CRD has not graduated past v1alpha1 anywhere in the v0.x line. The comment next to the chart pin flags re-checking this whenever the pin is bumped.

pkg/nic/destroy.go
tylerpotts: Confirmed no mismatch. `tofu destroy` deletes every resource tracked in state regardless of input-variable drift — variables only need to evaluate so provider/module config can be built; a changed `extra_ca_bundle` input would surface as a diff on plan/apply, but destroy does not reconcile inputs against state. Also note the provider-scoped tolerance half of this is gone as of fb93708: the deprecated `cluster.aws.trust_bundle` was removed entirely, so the only tolerance left is the top-level one in `pkg/nic/destroy.go`.

viniciusdc:
Looking over it today, thanks @tylerpotts

viniciusdc:
Ran this branch with #353 stacked on top on a fresh k3d cluster. The projection works, trust-manager lands the org CA in every app namespace and the repo-server clones through the TLS-inspecting proxy fine (that half I checked over on #353).

One thing I'd want to sort out before merge though. The Bundle] writes key `ca-certificates.crt` from `inLine` with just the org CA and no `useDefaultCAs: true`, so what lands in the namespaces is a one-cert file. 

https://github.com/nebari-dev/nebari-infrastructure-core/blob/213f4552d82a284217ea4314f70c05379cb07475/pkg/argocd/templates/manifests/security/trust-bundle/bundle.yaml#L13-L18

I checked on the cluster and the projected CM has exactly that, only the org CA. Anything that points `SSL_CERT_FILE` or `REQUESTS_CA_BUNDLE` at it loses every public root.

I don't think this is a judgment call, the parents already settled it should be system + org. #307 wires the in-pod trust through `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` / `NODE_EXTRA_CA_CERTS`, and those replace the system pool rather than add to it. Oren went through this on #310 and showed `--ca-file` / `GIT_SSL_CAINFO` swap the whole pool out, so an org-only file falls over the moment something is allow-listed past the proxy or redirects to a publicly-signed host. The pip-install-from-PyPI line in #307's DoD needs the public roots present too. And #353 already builds system ∪ org for the repo-server for this exact reason, so `useDefaultCAs: true` (or a separate key) would line the Bundle up with it.

The repo-server itself is fine either way since #353 builds its own combined bundle, so this really only bites the other consumers, keycloak, the packs, singleuser.

Separate small thing: the scrubbed `nic-config.yaml` that gets written to git still has `trust_bundle.path` in it, which is a path on my machine. Doesn't hurt anything but it means nothing in the repo, so it should probably get stripped along with the auth fields.

tylerpotts:
Both requested changes from @dcmcand's review and both points from @viniciusdc's k3d run are now addressed:

1. **Deprecation warning never fired on deploy** (dcmcand) — resolved in fb93708 by removing the deprecated `cluster.aws.trust_bundle` field entirely (see the inline thread on `pkg/provider/aws/provider.go` for details).
2. **`trust_bundle` in the committed `nic-config.yaml`** (dcmcand, and viniciusdc's "separate small thing") — resolved in d04baa4. `scrubbedConfig` now rewrites any configured bundle to its resolved inline PEM (reusing the value already resolved once in `Deploy`, no extra disk read), and strips it entirely when it resolved empty. A machine-local `path:` no longer lands in the GitOps record, and the committed config stays self-contained and reflects the deployed value. Added table cases covering path-to-inline rewrite, empty-resolution strip, unset passthrough, and a marshalled-output check that the local path never appears.
3. **Org-only projected bundle loses every public root** (viniciusdc) — resolved in 6837147. Agreed this isn't a judgment call: #307 wires in-pod trust through `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` / `NODE_EXTRA_CA_CERTS`, which replace the default pool rather than extend it, and https://github.com/nebari-dev/nebari-infrastructure-core/issues/310 already demonstrated the swap-the-whole-pool failure mode. The Bundle now sources `useDefaultCAs: true` ahead of the org `inLine`, so the projected `ca-certificates.crt` is system ∪ org — matching what #353 builds for the repo-server. Since the Bundle now hard-depends on trust-manager's default CA package (trust-manager rejects `useDefaultCAs` Bundles when running without one), `defaultPackage.enabled: true` is pinned explicitly in the chart values rather than relying on the chart default. The template test now asserts both sources.

`golangci-lint run` and `go test -short ./...` are clean. Ready for re-review.

viniciusdc (approved):
Both findings are resolved and I re-verified them on a fresh from-binary deploy: the projected `ca-certificates.crt` is now system + org (144 certs) via `useDefaultCAs` plus the pinned `defaultPackage`, and the scrubbed `nic-config.yaml` carries the inline PEM instead of the machine-local path. trust-manager comes up clean from zero and projects into the namespaces. LGTM.

The nits I raised are non-blocking: the blanket `namespaceSelector: {}` (projects into every namespace, not just the ones that need it) and the fact this really lands "trust-manager + Keycloak" together. The pack-consumption / pip-install half of the epic is tracked separately (see #307 and nebari-dev/data-science-pack#154), not this PR's scope.

PR #348, opened 2026-06-02
feat(gpu): install NVIDIA GPU Operator via foundational software flow

viniciusdc:
Installs the NVIDIA GPU Operator on AWS clusters that declare GPU node groups, so `nvidia.com/gpu` is advertised without every software pack shipping its own operator app. Follows the imperative-Helm-in-`Deploy` pattern of the cluster-autoscaler / LBC / Longhorn addons, gated on a config flag rather than a GitOps ArgoCD app. Decision recorded in ADR-0006 (#361).

## What this ships

- **`pkg/providers/cluster/aws/gpu_operator.go`**: `installGPUOperator` / `upgradeGPUOperator` / `uninstallGPUOperator` over Helm (history-check, then install or upgrade). Install and upgrade use `RunWithContext(ctx, ...)` so cancellation/timeouts propagate (the older sibling installers still use `Run`). Values are the AWS NVIDIA-AMI defaults: `driver` and `toolkit` off (the AL2023 NVIDIA AMI ships them), `devicePlugin` on, and `MOFED_ENABLED=false` (AWS's documented workaround for the device-plugin/EFA `uverbs` conflict). Chart pinned to `v26.3.2` from the NGC registry (`helm.ngc.nvidia.com`, which lists it v-prefixed). No IAM/Pod Identity needed; the operator only touches in-cluster resources.
- **`config.go`**: `HasGPUNodeGroups()` predicate (any node group with `gpu: true`).
- **Wiring** (`provider.go`): `Deploy` installs when `HasGPUNodeGroups()` is true, before the ArgoCD/GitOps stage so `nvidia.com/gpu` is advertised by the time GPU workloads sync. `Destroy` uninstalls before `tofu destroy`, gated on the same flag and best-effort (kubeconfig-unavailable warns and skips; an uninstall failure warns and continues without aborting the destroy, since the operator has no cloud resources to drain).
- Tests: `HasGPUNodeGroups` (empty / mixed / only-GPU) and the Helm values (table-driven).

## Scope / what this deliberately does NOT do

Per ADR-0006, this is the hand-rolled interim shape, and the reconcile/prune loop is deferred to the shared `ConditionalCharts` interface in #349. Concretely, this PR does **not** include:

- live-node instance-type detection,
- uninstall-on-disable during `Deploy` (only a full `Destroy` removes the operator),
- any change to `InfraSettings` or a shared GPU config struct.

Gating is purely the `gpu: true` config flag. If a cluster declares GPU node groups the operator is installed; if not, nothing happens.

## Issue #232

Partially resolved by design: the core "`nvidia.com/gpu` isn't advertised" problem is fixed for the declare-`gpu: true`-and-deploy path. Live detection and prune-on-disable are deferred to #349 per ADR-0006.

Device-management docs behind the AWS defaults: [EKS](https://docs.aws.amazon.com/eks/latest/userguide/device-management-nvidia.html).

### DIFF
--- a/pkg/providers/cluster/aws/config.go
+++ b/pkg/providers/cluster/aws/config.go
@@ -186,6 +186,20 @@ type Taint struct {
 	Effect string `yaml:"effect" json:"effect"` // NO_SCHEDULE, NO_EXECUTE, PREFER_NO_SCHEDULE
 }
 
+// HasGPUNodeGroups reports whether any node group is tagged gpu: true. This is
+// the AWS idiom for "this cluster has GPU hardware": the infra layer already
+// selects the AL2023_x86_64_NVIDIA AMI for these node groups, and the GPU
+// operator install (see gpu_operator.go) keys off it. Exported to match the
+// sibling LonghornEnabled / ClusterAutoscalerEnabled / LoadBalancerControllerEnabled predicates.
+func (c *Config) HasGPUNodeGroups() bool {
+	for _, ng := range c.NodeGroups {
+		if ng.GPU {
+			return true
+		}
+	}
+	return false
+}
+
 // LonghornEnabled returns whether Longhorn distributed block storage should
 // be deployed on this AWS cluster. Defaults to true when the Longhorn block
 // is omitted entirely — Longhorn is the AWS storage default. The shared
--- a/pkg/providers/cluster/aws/gpu_operator.go
+++ b/pkg/providers/cluster/aws/gpu_operator.go
@@ -1,0 +1,257 @@
+package aws
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/storage/driver"
+
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/helm"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/status"
+)
+
+const (
+	gpuOperatorRepoName  = "nvidia"
+	gpuOperatorRepoURL   = "https://helm.ngc.nvidia.com/nvidia"
+	gpuOperatorChartName = "nvidia/gpu-operator"
+	// gpuOperatorChartVersion pins the upstream NVIDIA gpu-operator Helm chart.
+	// The NGC registry lists this chart with a leading "v" (e.g. "v26.3.2"), so
+	// the prefix is intentional and required for LocateChart to resolve it -
+	// unlike the bare-semver aws-load-balancer-controller chart. Check the NGC
+	// catalog (https://catalog.ngc.nvidia.com/orgs/nvidia/helm-charts/gpu-operator)
+	// for the current version when bumping.
+	gpuOperatorChartVersion   = "v26.3.2"
+	gpuOperatorNamespace      = "gpu-operator"
+	gpuOperatorReleaseName    = "gpu-operator"
+	gpuOperatorInstallTimeout = 10 * time.Minute
+
+	// enabledKey is the gpu-operator chart's per-component on/off value key.
+	enabledKey = "enabled"
+)
+
+// gpuOperatorHelmValues builds the Helm values for the gpu-operator chart.
+//
+// These are the AWS defaults: the EKS-optimized AL2023 NVIDIA AMI already ships
+// the NVIDIA driver and the container toolkit, so the operator only needs to
+// layer on the device plugin (what advertises nvidia.com/gpu). MOFED is
+// disabled so the device plugin doesn't claim all /dev/infiniband/uverbs*
+// devices and conflict with the AWS EFA device plugin on EFA-enabled nodes.
+//
+// When the operator grows to other providers these values will need to vary by
+// provider; for now they are AWS-specific.
+func gpuOperatorHelmValues() map[string]any {
+	return map[string]any{
+		"driver":  map[string]any{enabledKey: false},
+		"toolkit": map[string]any{enabledKey: false},
+		"devicePlugin": map[string]any{
+			enabledKey: true,
+			"env": []any{
+				map[string]any{"name": "MOFED_ENABLED", "value": "false"},
+			},
+		},
+	}
+}
+
+// loadGPUOperatorChart locates and loads the gpu-operator Helm chart.
+func loadGPUOperatorChart(chartPathOptions action.ChartPathOptions) (*chart.Chart, error) {
+	chartPath, err := chartPathOptions.LocateChart(gpuOperatorChartName, cli.New())
+	if err != nil {
+		return nil, fmt.Errorf("failed to locate GPU Operator chart: %w", err)
+	}
+
+	loadedChart, err := loader.Load(chartPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load GPU Operator chart: %w", err)
+	}
+
+	return loadedChart, nil
+}
+
+// installGPUOperator installs or upgrades the NVIDIA GPU Operator via Helm.
+// Safe to call on every deploy; it upgrades an existing release rather than
+// failing. The operator needs no cloud IAM - it manages only in-cluster
+// resources - so unlike the LBC and autoscaler there is no Pod Identity here.
+func installGPUOperator(ctx context.Context, kubeconfigBytes []byte) error {
+	tracer := otel.Tracer("nebari-infrastructure-core")
+	ctx, span := tracer.Start(ctx, "aws.installGPUOperator")
+	defer span.End()
+
+	span.SetAttributes(attribute.String("chart_version", gpuOperatorChartVersion))
+
+	kubeconfigPath, cleanup, err := helm.WriteTempKubeconfig(kubeconfigBytes)
+	if err != nil {
+		span.RecordError(err)
+		return err
+	}
+	defer cleanup()
+
+	if err := helm.AddRepo(ctx, gpuOperatorRepoName, gpuOperatorRepoURL); err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("failed to add nvidia Helm repository: %w", err)
+	}
+
+	actionConfig, err := helm.NewActionConfig(kubeconfigPath, gpuOperatorNamespace)
+	if err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("failed to create Helm action config: %w", err)
+	}
+
+	histClient := action.NewHistory(actionConfig)
+	histClient.Max = 1
+	switch _, err := histClient.Run(gpuOperatorReleaseName); {
+	case err == nil:
+		status.Send(ctx, status.NewUpdate(status.LevelInfo, "GPU Operator already installed, upgrading").
+			WithResource("gpu-operator").
+			WithAction("upgrading"))
+		return upgradeGPUOperator(ctx, actionConfig)
+	case errors.Is(err, driver.ErrReleaseNotFound):
+		// No existing release; fall through to fresh install.
+	default:
+		span.RecordError(err)
+		return fmt.Errorf("failed to query Helm release history for %s: %w", gpuOperatorReleaseName, err)
+	}
+
+	status.Send(ctx, status.NewUpdate(status.LevelProgress, "Installing GPU Operator").
+		WithResource("gpu-operator").
+		WithAction("installing").
+		WithMetadata("chart_version", gpuOperatorChartVersion))
+
+	client := action.NewInstall(actionConfig)
+	client.Namespace = gpuOperatorNamespace
+	client.ReleaseName = gpuOperatorReleaseName
+	client.CreateNamespace = true
+	client.Wait = true
+	client.Timeout = gpuOperatorInstallTimeout
+	client.Version = gpuOperatorChartVersion
+
+	loadedChart, err := loadGPUOperatorChart(client.ChartPathOptions)
+	if err != nil {
+		span.RecordError(err)
+		return err
+	}
+
+	// RunWithContext, not Run: Install.Run builds its own context.Background()
+	// and ignores ours, so a cancellation/timeout during the install window
+	// wouldn't propagate to Helm.
+	release, err := client.RunWithContext(ctx, loadedChart, gpuOperatorHelmValues())
+	if err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("failed to install GPU Operator: %w", err)
+	}
+
+	span.SetAttributes(
+		attribute.String("release_status", string(release.Info.Status)),
+		attribute.Int("release_version", release.Version),
+	)
+
+	status.Send(ctx, status.NewUpdate(status.LevelSuccess, "GPU Operator installed").
+		WithResource("gpu-operator").
+		WithAction("installed").
+		WithMetadata("chart_version", gpuOperatorChartVersion))
+
+	return nil
+}
+
+// upgradeGPUOperator upgrades an existing gpu-operator release.
+func upgradeGPUOperator(ctx context.Context, actionConfig *action.Configuration) error {
+	tracer := otel.Tracer("nebari-infrastructure-core")
+	ctx, span := tracer.Start(ctx, "aws.upgradeGPUOperator")
+	defer span.End()
+
+	span.SetAttributes(attribute.String("chart_version", gpuOperatorChartVersion))
+
+	client := action.NewUpgrade(actionConfig)
+	client.Namespace = gpuOperatorNamespace
+	client.Wait = true
+	client.Timeout = gpuOperatorInstallTimeout
+	client.Version = gpuOperatorChartVersion
+
+	loadedChart, err := loadGPUOperatorChart(client.ChartPathOptions)
+	if err != nil {
+		span.RecordError(err)
+		return err
+	}
+
+	// RunWithContext, not Run: Upgrade.Run ignores the caller's context (it
+	// builds its own context.Background()).
+	release, err := client.RunWithContext(ctx, gpuOperatorReleaseName, loadedChart, gpuOperatorHelmValues())
+	if err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("failed to upgrade GPU Operator: %w", err)
+	}
+
+	span.SetAttributes(
+		attribute.String("release_status", string(release.Info.Status)),
+		attribute.Int("release_version", release.Version),
+	)
+
+	status.Send(ctx, status.NewUpdate(status.LevelSuccess, "GPU Operator upgraded").
+		WithResource("gpu-operator").
+		WithAction("upgraded").
+		WithMetadata("chart_version", gpuOperatorChartVersion))
+
+	return nil
+}
+
+// uninstallGPUOperator removes the gpu-operator Helm release if present. A
+// missing release is not an error. Called during cluster destroy, before tofu
+// tears the cluster down. (Uninstall.Run has no RunWithContext in this Helm
+// version, so it stays on Run.)
+func uninstallGPUOperator(ctx context.Context, kubeconfigBytes []byte) error {
+	tracer := otel.Tracer("nebari-infrastructure-core")
+	ctx, span := tracer.Start(ctx, "aws.uninstallGPUOperator")
+	defer span.End()
+
+	kubeconfigPath, cleanup, err := helm.WriteTempKubeconfig(kubeconfigBytes)
+	if err != nil {
+		span.RecordError(err)
+		return err
+	}
+	defer cleanup()
+
+	actionConfig, err := helm.NewActionConfig(kubeconfigPath, gpuOperatorNamespace)
+	if err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("failed to create Helm action config: %w", err)
+	}
+
+	histClient := action.NewHistory(actionConfig)
+	histClient.Max = 1
+	switch _, err := histClient.Run(gpuOperatorReleaseName); {
+	case errors.Is(err, driver.ErrReleaseNotFound):
+		// Nothing installed; nothing to do.
+		return nil
+	case err != nil:
+		span.RecordError(err)
+		return fmt.Errorf("failed to query Helm release history for %s: %w", gpuOperatorReleaseName, err)
+	}
+
+	status.Send(ctx, status.NewUpdate(status.LevelProgress, "Uninstalling GPU Operator").
+		WithResource("gpu-operator").
+		WithAction("uninstalling"))
+
+	// Wait=false on teardown: this only runs during Destroy, right before tofu
+	// deletes the nodes, and the operator has no cloud resources to drain, so
+	// blocking on graceful removal only adds latency and a failure surface.
+	client := action.NewUninstall(actionConfig)
+	client.Wait = false
+
+	if _, err := client.Run(gpuOperatorReleaseName); err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("failed to uninstall GPU Operator: %w", err)
+	}
+
+	status.Send(ctx, status.NewUpdate(status.LevelSuccess, "GPU Operator uninstalled").
+		WithResource("gpu-operator").
+		WithAction("uninstalled"))
+
+	return nil
+}
--- a/pkg/providers/cluster/aws/provider.go
+++ b/pkg/providers/cluster/aws/provider.go
@@ -428,6 +428,23 @@ func (p *Provider) Deploy(ctx context.Context, projectName string, clusterConfig
 		}
 	}
 
+	// Install the NVIDIA GPU Operator when the cluster has GPU node groups, so
+	// nvidia.com/gpu is advertised before the GitOps/ArgoCD stage. Gated on the
+	// gpu: true config flag, matching the LBC and cluster-autoscaler; teardown
+	// happens in Destroy. Decision recorded in ADR-0006 (#361).
+	if awsCfg.HasGPUNodeGroups() {
+		kubeconfigBytes, err := p.GetKubeconfig(ctx, projectName, clusterConfig)
+		if err != nil {
+			span.RecordError(err)
+			return fmt.Errorf("failed to get kubeconfig for GPU Operator install: %w", err)
+		}
+
+		if err := installGPUOperator(ctx, kubeconfigBytes); err != nil {
+			span.RecordError(err)
+			return fmt.Errorf("failed to install GPU Operator: %w", err)
+		}
+	}
+
 	// Create EFS StorageClass if EFS is enabled
 	if awsCfg.EFS != nil && awsCfg.EFS.Enabled {
 		kubeconfigBytes, err := p.GetKubeconfig(ctx, projectName, clusterConfig)
@@ -636,6 +653,25 @@ func (p *Provider) Destroy(ctx context.Context, projectName string, clusterConfi
 				}
 				status.Send(ctx, status.NewUpdate(status.LevelWarning, fmt.Sprintf("Longhorn uninstall failed, continuing with --force: %v", err)).
 					WithResource("longhorn").WithAction("uninstalling"))
+			}
+		}
+	}
+
+	// Uninstall the GPU Operator before tofu destroy, gated on the GPU config
+	// flag to mirror the Longhorn block above. The operator has no cloud
+	// resources that can block teardown, so an uninstall failure is never fatal
+	// to the destroy: we warn and continue even without --force (unlike
+	// Longhorn, whose CSI finalizers can wedge node-group deletion). Best-effort.
+	if awsCfg.HasGPUNodeGroups() {
+		kubeconfigBytes, kErr := p.GetKubeconfig(ctx, projectName, clusterConfig)
+		switch {
+		case kErr != nil:
+			status.Send(ctx, status.NewUpdate(status.LevelWarning, fmt.Sprintf("Skipping GPU Operator uninstall: kubeconfig unavailable: %v", kErr)).
+				WithResource("gpu-operator").WithAction("uninstalling"))
+		default:
+			if err := uninstallGPUOperator(ctx, kubeconfigBytes); err != nil {
+				status.Send(ctx, status.NewUpdate(status.LevelWarning, fmt.Sprintf("GPU Operator uninstall failed, continuing: %v", err)).
+					WithResource("gpu-operator").WithAction("uninstalling"))
 			}
 		}
 	}
--- a/examples/aws-config.yaml
+++ b/examples/aws-config.yaml
@@ -131,6 +131,11 @@ cluster:
     # the deploy waiting for an LB endpoint.
     # load_balancer_scheme: internet-facing  # or: internal
 
+    # When a node group sets gpu: true, nic installs the NVIDIA GPU Operator on
+    # the cluster (device plugin only; the AL2023_x86_64_NVIDIA AMI already
+    # ships the driver and container toolkit) so nvidia.com/gpu becomes
+    # schedulable. No extra config is required. See ADR-0006 for the rationale.
+
     # EFS configuration for shared storage (optional)
     # Creates an EFS file system with mount targets in each private subnet
     efs:
--- a/pkg/providers/cluster/aws/gpu_operator_test.go
+++ b/pkg/providers/cluster/aws/gpu_operator_test.go
@@ -1,0 +1,82 @@
+package aws
+
+import "testing"
+
+func TestHasGPUNodeGroups(t *testing.T) {
+	tests := []struct {
+		name string
+		cfg  Config
+		want bool
+	}{
+		{
+			name: "empty config",
+			cfg:  Config{},
+			want: false,
+		},
+		{
+			name: "no gpu node groups",
+			cfg:  Config{NodeGroups: map[string]NodeGroup{"user": {Instance: "m7i.xlarge"}}},
+			want: false,
+		},
+		{
+			name: "mixed cpu and gpu node groups",
+			cfg: Config{NodeGroups: map[string]NodeGroup{
+				"user": {Instance: "m7i.xlarge"},
+				"gpu":  {Instance: "g4dn.2xlarge", GPU: true},
+			}},
+			want: true,
+		},
+		{
+			name: "only a gpu node group",
+			cfg: Config{NodeGroups: map[string]NodeGroup{
+				"gpu": {Instance: "g4dn.2xlarge", GPU: true},
+			}},
+			want: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := tt.cfg.HasGPUNodeGroups(); got != tt.want {
+				t.Errorf("HasGPUNodeGroups() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestGPUOperatorHelmValues(t *testing.T) {
+	v := gpuOperatorHelmValues()
+
+	// On the EKS NVIDIA AMI the driver and toolkit are preinstalled, so the
+	// operator only layers on the device plugin.
+	componentEnabled := []struct {
+		component string
+		want      bool
+	}{
+		{"driver", false},
+		{"toolkit", false},
+		{"devicePlugin", true},
+	}
+	for _, tt := range componentEnabled {
+		t.Run(tt.component, func(t *testing.T) {
+			m, ok := v[tt.component].(map[string]any)
+			if !ok {
+				t.Fatalf("%s should be a map, got %v", tt.component, v[tt.component])
+			}
+			if m[enabledKey] != tt.want {
+				t.Errorf("%s.enabled = %v, want %v", tt.component, m[enabledKey], tt.want)
+			}
+		})
+	}
+
+	t.Run("MOFED disabled for EFA safety", func(t *testing.T) {
+		dp := v["devicePlugin"].(map[string]any)
+		env, ok := dp["env"].([]any)
+		if !ok || len(env) == 0 {
+			t.Fatalf("devicePlugin.env should set MOFED_ENABLED, got %v", dp["env"])
+		}
+		first, _ := env[0].(map[string]any)
+		if first["name"] != "MOFED_ENABLED" || first["value"] != "false" {
+			t.Errorf("expected MOFED_ENABLED=false, got %v", first)
+		}
+	})
+}


pkg/provider/aws/gpu_operator.go
```diff
@@ -0,0 +1,320 @@
...
+// (compute) families - e.g. g4dn.2xlarge, g5.xlarge, gr6.4xlarge, p4d.24xlarge.
+// No general-purpose family starts with g or p, so a leading g/p plus a digit
+// in the family (which every real GPU family has, including gr6) is a reliable
+// signal while still excluding non-instance strings like "graviton".
+func isGPUInstanceType(instanceType string) bool {
+	family, _, _ := strings.Cut(strings.ToLower(strings.TrimSpace(instanceType)), ".")
+	if family == "" || (family[0] != 'g' && family[0] != 'p') {
+		return false
+	}
+	return strings.ContainsAny(family, "0123456789")
```
dcmcand: **Blocker:** this returns `true` for `g4ad.*`, but g4ad instances carry an AMD Radeon Pro V520, not an NVIDIA GPU. Installing the NVIDIA operator there deploys a device plugin that finds nothing and never advertises `nvidia.com/gpu` - a cluster that looks set up but silently isn't. The `g`+digit rule can't distinguish `g4ad` from the NVIDIA `g4dn`. Please exclude `g4ad` explicitly, and add it to the negative list in `TestIsGPUInstanceType` (which currently omits it, so the gap isn't caught).

pkg/provider/aws/gpu_operator.go
```diff
@@ -0,0 +1,320 @@
...
+	tracer := otel.Tracer("nebari-infrastructure-core")
+	ctx, span := tracer.Start(ctx, "aws.reconcileGPUOperator")
+	defer span.End()
+
+	want := cfg.hasGPUNodeGroups()
+	if !want {
+		live, err := clusterHasGPUNodes(ctx, kubeconfigBytes)
+		if err != nil {
+			span.RecordError(err)
+			return fmt.Errorf("failed to inspect cluster for GPU nodes: %w", err)
```
dcmcand: **Blocker:** this propagates a live-node-check error as a hard `Deploy` failure, and it sits on the CPU-only path that runs on every deploy. A transient Kubernetes API error here aborts an otherwise-successful deploy after `tf.Apply`, whereas `Destroy` treats the same situation as best-effort.

Suggest making the live check advisory: on error, warn via `status.Send` and fall through with `want = false`. `uninstallGPUOperator` is already a safe no-op when the release is absent, so a flaky list call can't fail the deploy for a component that has no cloud resources.

pkg/providers/cluster/aws/gpu_operator.go
```diff
@@ -0,0 +1,320 @@
...
+)
+
+const (
+	gpuOperatorRepoName  = "nvidia"
+	gpuOperatorRepoURL   = "https://helm.ngc.nvidia.com/nvidia"
+	gpuOperatorChartName = "nvidia/gpu-operator"
+	// gpuOperatorChartVersion pins the upstream NVIDIA gpu-operator Helm chart.
+	// v26.3.2 is the latest published version in the nvidia repo (chart version
+	// tracks appVersion). Bump as new releases land.
+	gpuOperatorChartVersion   = "v26.3.2"
```
dcmcand, line 31: Two small things:
- The `v` prefix (`v26.3.2`) differs from the LBC's `defaultLBCChartVersion = "3.2.1"`. Helm chart versions are usually bare semver (the `v` is a git-tag convention). Can you confirm Helm's `LocateChart` against the NGC registry accepts the `v`? If NGC expects `26.3.2`, this fails with an opaque "chart version not found." Matching the LBC format would be safer.
- The comment says the chart version "tracks appVersion" - that isn't generally true for gpu-operator across releases. Pointing at the NGC catalog/releases page would be more durable bump guidance.

pkg/providers/cluster/aws/gpu_operator.go
```diff
@@ -0,0 +1,320 @@
...
+	status.Send(ctx, status.NewUpdate(status.LevelSuccess, "GPU Operator installed").
+		WithResource("gpu-operator").
+		WithAction("installed").
+		WithMetadata("chart_version", gpuOperatorChartVersion))
+
+	return nil
+}
+
+// upgradeGPUOperator upgrades an existing gpu-operator release.
+func upgradeGPUOperator(ctx context.Context, actionConfig *action.Configuration) error {
```
dcmcand, line 164: Question (also applies to `upgradeAWSLoadBalancerController`, so not a regression): the upgrade path doesn't call `helm.AddRepo` before `LocateChart`. Within one deploy it's fine since install runs `AddRepo` first, but on a fresh machine where the operator already exists on-cluster, the upgrade path runs without the NGC repo in the local Helm cache and `LocateChart` can fail. Worth confirming a mitigation or adding `AddRepo` here.

pkg/provider/aws/gpu_operator_test.go
```diff
@@ -0,0 +1,83 @@
+package aws
+
+import "testing"
+
+func TestIsGPUInstanceType(t *testing.T) {
```
dcmcand: Stylistic (repo convention is table-driven tests, per CLAUDE.md): converting this to a `[]struct{name, input, want}` table with `t.Run` would match `TestHasGPUNodeGroups` and make it easy to add the `g4ad` negative case from the blocker above.

dcmcand (changes requested):
Thanks for this - the architecture is clean and consistent with the autoscaler/LBC Helm-in-`Deploy` pattern, the OTel and `status.Send` conventions all hold, and the helm-values test guarding driver/toolkit/MOFED against a future chart bump is a nice touch. Requesting changes for two correctness issues plus the merge state.

Two blockers (inline):
1. `isGPUInstanceType` matches `g4ad`, which is an AMD GPU, not NVIDIA - the operator would install and advertise nothing.
2. `reconcileGPUOperator` fails the whole deploy on a transient live-node-check error, on the CPU-only path that runs on every deploy. `Destroy` already handles this best-effort; `Deploy` should too.

Also, the PR is currently not mergeable: it conflicts with `main` in `provider.go` and `config.go` because #352 (autoscaling) merged and restructured the same `Deploy`/`Destroy` regions. A rebase onto main is needed. The bonus is that `reconcileGPUOperator` can then mirror the merged `installClusterAutoscaler` call site, including a gate so CPU-only clusters skip the extra API round-trips (see the inline question).

A few non-blocking questions/nits are inline (chart-version `v` prefix vs NGC, upgrade-path `AddRepo`, table-driven test). I verified the `gpu_operator` config-block removal is non-breaking (main never carried one) and that `isGPUInstanceType` correctly excludes graviton / inf2 / trn1 (just not g4ad).

Design-of-record ADR-0006 (#361) and the shared interface in #349 noted, no concerns there.

pkg/provider/aws/gpu_operator.go
viniciusdc: Good catch. Fixed — `isGPUInstanceType` now excludes `g4ad` explicitly (it is the only AMD GPU family on EC2), and `g4ad.xlarge` is in the negative list of the now table-driven `TestIsGPUInstanceType`.

pkg/provider/aws/gpu_operator.go
viniciusdc: Fixed — the live-node check is now advisory. On a list error it `status.Send`s a warning and falls through with `want=false` instead of aborting the deploy; install/uninstall errors still surface as real failures. Config `gpu: true` short-circuits the check entirely, so clusters that declare GPU never pay the round-trip, and CPU-only clusters do a single advisory node-list that can no longer fail the deploy. If you would rather drop it to a pure config gate (zero round-trips on CPU-only, at the cost of not catching undeclared GPU nodes), say the word.

pkg/providers/cluster/aws/gpu_operator.go
viniciusdc: Confirmed against the NGC index: gpu-operator publishes its chart as `version: "v26.3.2"`, with the `v` — so the prefix is required for `LocateChart` here, unlike the bare-semver LBC chart. Kept `v26.3.2` and rewrote the comment to say exactly that and point at the NGC catalog for bump guidance, dropping the "tracks appVersion" line (you are right that it does not generally hold).

pkg/providers/cluster/aws/gpu_operator.go
viniciusdc: In our call path it is covered: `reconcileGPUOperator → installGPUOperator` runs `helm.AddRepo` before the history-check that branches to upgrade, so the repo is in the local cache by the time `upgradeGPUOperator → LocateChart` runs. `upgradeGPUOperator` is never called on its own. Left as-is to mirror the LBC structure you flagged; happy to hoist `AddRepo` above the branch if you would prefer it explicit in both.

pkg/provider/aws/gpu_operator_test.go
viniciusdc: Done — converted to a `name/input/want` table with `t.Run`, matching `TestHasGPUNodeGroups`, and added the `g4ad` negative case plus `trn1`/graviton/empty.

viniciusdc:
Thanks for the thorough review @dcmcand. Pushed the changes (rebased onto main, single commit now):

- **g4ad blocker** — `isGPUInstanceType` excludes `g4ad`; added to the table-driven test negatives.
- **Live-node-check blocker** — the check is advisory now (warn + `want=false` on error, never aborts the deploy); install/uninstall errors still surface. `gpu: true` short-circuits it so GPU clusters skip the round-trip.
- **Rebase** — onto `main` past #352; `reconcileGPUOperator` sits right after the autoscaler install in `Deploy`, uninstall before `tofu destroy` in `Destroy`.
- **Chart `v` prefix** — confirmed correct against the NGC index (chart is published v-prefixed); comment rewritten.
- **AddRepo** — covered by the call path (install AddRepos before the history-check branch); details inline.
- **Table-driven test** — done.

Replies on each thread. ADR-0006 (the Helm-vs-GitOps decision) is split out to #361 for its own discussion.

pkg/provider/aws/gpu_operator.go
```diff
@@ -0,0 +1,340 @@
...
+	client.Timeout = gpuOperatorInstallTimeout
+	client.Version = gpuOperatorChartVersion
+
+	loadedChart, err := loadGPUOperatorChart(client.ChartPathOptions)
+	if err != nil {
+		span.RecordError(err)
+		return err
+	}
+
+	release, err := client.Run(loadedChart, gpuOperatorHelmValues())
```
dcmcand: **[Blocker] `client.Run` discards `ctx` - use `RunWithContext`.**

In Helm v3.19.4, `Install.Run` creates `context.Background()` internally and ignores the caller's context. A Ctrl-C or upstream timeout during the ~10-minute install window won't be propagated to Helm: the install keeps running in the background after the CLI has already returned an error or exited.

```go
release, err := client.RunWithContext(ctx, loadedChart, gpuOperatorHelmValues())
```

(The LBC/autoscaler siblings have the same issue but aren't touched by this PR, so just fixing the new code is consistent with project norms.)

pkg/provider/aws/gpu_operator.go
```diff
@@ -0,0 +1,340 @@
...
+	client.Timeout = gpuOperatorInstallTimeout
+	client.Version = gpuOperatorChartVersion
+
+	loadedChart, err := loadGPUOperatorChart(client.ChartPathOptions)
+	if err != nil {
+		span.RecordError(err)
+		return err
+	}
+
+	release, err := client.Run(gpuOperatorReleaseName, loadedChart, gpuOperatorHelmValues())
```
dcmcand: **[Blocker] Same `ctx`-dropping issue on the upgrade path.** `Upgrade.Run` also builds its own `context.Background()` in this Helm version. Switch to `client.RunWithContext(ctx, gpuOperatorReleaseName, loadedChart, gpuOperatorHelmValues())`.

Note: `Uninstall.Run` (line 330) has no `RunWithContext` equivalent in v3.19.4, so leave that one as-is.

pkg/provider/aws/gpu_operator.go
```diff
@@ -0,0 +1,340 @@
...
+	// fallback is advisory: a transient node-list error must not fail an
+	// otherwise-successful deploy, since the operator has no cloud resources and
+	// uninstall is a no-op when absent. On error we warn and treat the cluster
+	// as GPU-free. Install/uninstall errors below still surface as real failures.
+	want := cfg.hasGPUNodeGroups()
+	if !want {
+		live, err := clusterHasGPUNodes(ctx, kubeconfigBytes)
+		if err != nil {
+			status.Send(ctx, status.NewUpdate(status.LevelWarning,
+				fmt.Sprintf("Skipping GPU node check — could not list nodes: %v", err)).
```
dcmcand: **[Blocker] Em dash in user-visible status message.** This project bans em dashes (U+2014) in output. Replace with a colon or hyphen, e.g. `"Skipping GPU node check: could not list nodes: %v"`.

pkg/provider/aws/provider.go
```diff
@@ -632,6 +650,27 @@ func (p *Provider) Destroy(ctx context.Context, projectName string, clusterConfi
...
 	}
 
+	// Uninstall the GPU Operator before tofu destroy. It has no cloud resources
+	// of its own so it can't block teardown, but removing it first keeps the
+	// teardown clean and symmetric with deploy. Best-effort, like Longhorn.
+	{
+		kubeconfigBytes, kErr := p.GetKubeconfig(ctx, projectName, clusterConfig)
+		switch {
+		case kErr != nil:
+			status.Send(ctx, status.NewUpdate(status.LevelWarning, fmt.Sprintf("Skipping GPU Operator uninstall — kubeconfig unavailable: %v", kErr)).
```
dcmcand: **[Blocker] Em dash in user-visible status message.** Same as the gpu_operator.go message - replace `—` with a colon or hyphen: `"Skipping GPU Operator uninstall: kubeconfig unavailable: %v"`. (The two existing Longhorn messages just above have the same pre-existing issue, but the new message should be clean.)

pkg/provider/aws/provider.go
```diff
@@ -420,6 +420,24 @@ func (p *Provider) Deploy(ctx context.Context, projectName string, clusterConfig
...
+	// the live-node check inside is advisory and never fails the deploy (see
+	// reconcileGPUOperator). Decision recorded in ADR-0006 (#361).
+	{
+		kubeconfigBytes, err := p.GetKubeconfig(ctx, projectName, clusterConfig)
+		if err != nil {
+			span.RecordError(err)
+			return fmt.Errorf("failed to get kubeconfig for GPU Operator reconcile: %w", err)
+		}
+
+		if err := reconcileGPUOperator(ctx, kubeconfigBytes, awsCfg); err != nil {
```
dcmcand: **[Question] Unconditional reconcile on every deploy.** This block runs for every deploy, including clusters with no GPU. For those, it fetches a kubeconfig (cached, cheap), makes a real `clusterHasGPUNodes` node-list API call, then a no-op uninstall. LBC and cluster-autoscaler are each guarded by a config flag instead.

Was the always-on reconcile intentional - i.e. a true desired-state loop that removes the operator when GPU nodes disappear? If the node-list call on non-GPU clusters is a concern, a `if cfg.hasGPUNodeGroups()` short-circuit before the live check would skip it, but that would miss undeclared GPU nodes. Just confirming intent so we know whether the live-node path is permanent or transitional.

pkg/providers/cluster/aws/gpu_operator.go
```diff
@@ -0,0 +1,340 @@
...
+		return err
+	}
+
+	release, err := client.Run(gpuOperatorReleaseName, loadedChart, gpuOperatorHelmValues())
+	if err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("failed to upgrade GPU Operator: %w", err)
+	}
+
+	span.SetAttributes(
```
dcmcand, line 191: **[Stylistic] Upgrade span is missing `chart_version`.** `installGPUOperator` sets `attribute.String("chart_version", gpuOperatorChartVersion)` on its span (line 179) but the upgrade path doesn't. Upgrade is the hot path on subsequent deploys for a GPU cluster, so anyone debugging "why didn't the chart version change" won't see it in upgrade traces. Add the same attribute here for parity.

dcmcand (commented):
Code review (Go). Strong implementation overall - GPU family detection, the union-of-signals reconcile logic, Destroy/Longhorn symmetry, and OTel instrumentation are all well done and faithfully follow the LBC/Longhorn pattern. Inline comments below cover 2 blockers, 1 question, and a few stylistic notes. (Architecture/ADR findings are being handled separately.)

pkg/providers/cluster/aws/gpu_operator.go
```diff
@@ -0,0 +1,340 @@
...
+			"env": []any{
+				map[string]any{"name": "MOFED_ENABLED", "value": "false"},
+			},
+		},
+	}
+}
+
+// loadGPUOperatorChart locates and loads the gpu-operator Helm chart.
+func loadGPUOperatorChart(chartPathOptions action.ChartPathOptions) (*chart.Chart, error) {
+	chartPath, err := chartPathOptions.LocateChart(gpuOperatorChartName, cli.New())
```
dcmcand, line 65: **[Stylistic / tracking only] `cli.New()` reads `HELM_*` env vars** (`HELM_NAMESPACE`, `HELM_REPOSITORY_CACHE`, etc.), so the effective Helm settings depend on the operator's shell environment in a way that isn't obvious from the signature. Consistent with the LBC/autoscaler siblings, so not a regression - just worth tracking as a source of subtle behavior differences when the #349 shared Helm infrastructure lands.

pkg/provider/aws/gpu_operator_test.go
```diff
@@ -0,0 +1,94 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := isGPUInstanceType(tt.input); got != tt.want {
+				t.Errorf("isGPUInstanceType(%q) = %v, want %v", tt.input, got, tt.want)
+			}
+		})
+	}
+}
+
+func TestHasGPUNodeGroups(t *testing.T) {
```
dcmcand: **[Stylistic] Add edge cases to `TestHasGPUNodeGroups`.** Ranging over a nil/empty map is safe and returns false, but an explicit `"empty node groups"` case (`Config{}`) documents that nil-safe behavior as intentional and guards against a future refactor to a length check. A case with only a `GPU: true` group (no non-GPU group) would also round out coverage.

pkg/provider/aws/gpu_operator_test.go
```diff
@@ -0,0 +1,94 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := tt.cfg.hasGPUNodeGroups(); got != tt.want {
+				t.Errorf("hasGPUNodeGroups() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestGPUOperatorHelmValues(t *testing.T) {
```
dcmcand: **[Stylistic] `TestGPUOperatorHelmValues` reads as a flat sequential test; project convention is table-driven.** The checked values (`driver.enabled`, `toolkit.enabled`, `devicePlugin.enabled`, `MOFED_ENABLED`) map cleanly to a table of `{path, expected}` rows, which also makes adding future coverage (`gfd.enabled`, `dcgm.enabled`) easier without restructuring.

pkg/provider/aws/gpu_operator.go
viniciusdc: Fixed - install now uses `client.RunWithContext(ctx, loadedChart, ...)` so cancellation/timeouts propagate.

pkg/provider/aws/gpu_operator.go
viniciusdc: Fixed - upgrade uses `client.RunWithContext(ctx, gpuOperatorReleaseName, loadedChart, ...)`. Left `Uninstall.Run` as-is per your note (no RunWithContext in this Helm version), with a comment to that effect.

pkg/provider/aws/gpu_operator.go
viniciusdc: Resolved by removal - that message lived in the live-node reconcile path, which is gone now (we gated on config instead, see the provider.go thread).

pkg/provider/aws/provider.go
viniciusdc: Fixed - colon instead of the em dash: `"Skipping GPU Operator uninstall: kubeconfig unavailable: %v"`.

pkg/provider/aws/provider.go
viniciusdc: Good call - switched to gating on `cfg.hasGPUNodeGroups()` in Deploy, matching the LBC and cluster-autoscaler. The always-on live-node loop is gone: I removed `clusterHasGPUNodes` and `isGPUInstanceType` entirely, so CPU-only clusters now do zero GPU round-trips. Tradeoff accepted: config `gpu: true` is the sole signal, so an undeclared GPU node group wouldn't trigger install, which is consistent with the other conditional installs. (This also dissolved the g4ad concern and the em-dash message above.)

pkg/providers/cluster/aws/gpu_operator.go
viniciusdc: Added - `span.SetAttributes(attribute.String("chart_version", gpuOperatorChartVersion))` on the upgrade span for parity with install.

pkg/providers/cluster/aws/gpu_operator.go
viniciusdc: Noted - tracking-only and consistent with the LBC/autoscaler siblings, so leaving as-is; the shared Helm infrastructure in #349 is the right place to make the Helm settings explicit.

pkg/provider/aws/gpu_operator_test.go
viniciusdc: Added - `TestHasGPUNodeGroups` now has an `empty config` case and an `only a gpu node group` case alongside the mixed one.

pkg/provider/aws/gpu_operator_test.go
viniciusdc: Converted - `TestGPUOperatorHelmValues` is now table-driven (a row per component for the enabled flags, plus a MOFED subtest). `TestIsGPUInstanceType` is gone with the live-node detection.

viniciusdc:
Thanks again @dcmcand. Pushed (rebased onto current main past the pkg/providers/cluster restructure, single commit):

- **ctx blockers** - install and upgrade now use `RunWithContext(ctx, ...)`; `Uninstall.Run` left as-is per your note.
- **em dash blockers** - the gpu_operator.go one is gone with the reconcile path; the provider.go uninstall message now uses a colon.
- **Unconditional reconcile (question)** - gated on `hasGPUNodeGroups()` to match LBC/autoscaler. Removed the live-node detection (`clusterHasGPUNodes`/`isGPUInstanceType`) entirely, so CPU-only clusters do zero GPU round-trips. Config `gpu: true` is now the sole signal; that also dissolved the g4ad concern.
- **Nits** - `chart_version` added to the upgrade span; `TestHasGPUNodeGroups` gained empty + only-GPU cases; `TestGPUOperatorHelmValues` is table-driven. `cli.New()` HELM_* left as tracking-only for #349.

Replies on each thread. ADR-0006 (the Helm-vs-GitOps decision) is in #361.

pkg/providers/cluster/aws/gpu_operator.go
```diff
@@ -0,0 +1,255 @@
...
+// When the operator grows to other providers these values will need to vary by
+// provider; for now they are AWS-specific.
+func gpuOperatorHelmValues() map[string]any {
+	return map[string]any{
+		"driver":  map[string]any{enabledKey: false},
+		"toolkit": map[string]any{enabledKey: false},
+		"devicePlugin": map[string]any{
+			enabledKey: true,
+			"env": []any{
+				map[string]any{"name": "MOFED_ENABLED", "value": "false"},
```
dcmcand, line 57: **Kudos.** Verified these against AWS's EKS device-management docs. `driver`/`toolkit` off matches the AL2023 NVIDIA AMI (which ships them), `devicePlugin` on is what advertises `nvidia.com/gpu`, and `MOFED_ENABLED=false` is exactly AWS's documented workaround for the device-plugin/EFA `uverbs` conflict, passed as the string `"false"` (correct for a k8s env value). Well researched.

pkg/providers/cluster/aws/gpu_operator.go
```diff
@@ -0,0 +1,255 @@
...
+	client.Timeout = gpuOperatorInstallTimeout
+	client.Version = gpuOperatorChartVersion
+
+	loadedChart, err := loadGPUOperatorChart(client.ChartPathOptions)
+	if err != nil {
+		span.RecordError(err)
+		return err
+	}
+
+	// RunWithContext, not Run: Install.Run builds its own context.Background()
```
dcmcand, line 141: **Kudos.** This is a real correctness improvement over the sibling installers. I confirmed in vendored helm v3.19.4 that `Install.Run` and `Upgrade.Run` build their own `context.Background()` and discard the caller's ctx, so a cancellation or timeout mid-install would not reach Helm. Using `RunWithContext` here (install and upgrade) propagates it correctly; `cluster_autoscaler.go` and `aws_load_balancer_controller.go` still use `Run`. The note on `uninstallGPUOperator` that `Uninstall` has no `RunWithContext` is also accurate.

pkg/providers/cluster/aws/gpu_operator.go
```diff
@@ -0,0 +1,255 @@
...
+		span.RecordError(err)
+		return fmt.Errorf("failed to query Helm release history for %s: %w", gpuOperatorReleaseName, err)
+	}
+
+	status.Send(ctx, status.NewUpdate(status.LevelProgress, "Uninstalling GPU Operator").
+		WithResource("gpu-operator").
+		WithAction("uninstalling"))
+
+	client := action.NewUninstall(actionConfig)
+	client.Wait = true
```
dcmcand: **Question.** `uninstallGPUOperator` sets `Wait=true` with a 10-minute timeout, and in `Destroy` a failure aborts `tofu destroy` unless `--force` is set. Since the operator has no cloud resources that can block teardown (as the `Destroy` comment itself notes), is the blocking graceful uninstall intended? I can see it both ways and do not want to assume: waiting buys clean webhook/finalizer removal before the nodes go, but it also adds teardown latency and a failure surface right before tofu deletes the cluster anyway. Would `Wait=false` (or at least not aborting destroy on an uninstall failure even without `--force`) be acceptable here?

pkg/providers/cluster/aws/config.go
```diff
@@ -186,6 +186,19 @@ type Taint struct {
 	Effect string `yaml:"effect" json:"effect"` // NO_SCHEDULE, NO_EXECUTE, PREFER_NO_SCHEDULE
 }
 
+// hasGPUNodeGroups reports whether any node group is tagged gpu: true. This is
+// the AWS idiom for "this cluster has GPU hardware" — the infra layer already
+// selects the AL2023_x86_64_NVIDIA AMI for these node groups, and the GPU
+// operator install (see gpu_operator.go) keys off it.
+func (c *Config) hasGPUNodeGroups() bool {
```
dcmcand: **Stylistic.** `hasGPUNodeGroups` is unexported, while the sibling predicates on `Config` in this file (`LonghornEnabled`, `ClusterAutoscalerEnabled`, `LoadBalancerControllerEnabled`) are all exported. It compiles fine since it is only used in-package today, but for consistency with that family of "does this cluster want X" methods, consider exporting it as `HasGPUNodeGroups` (or a brief note if it is intentionally private).

pkg/providers/cluster/aws/provider.go
```diff
@@ -640,6 +657,27 @@ func (p *Provider) Destroy(ctx context.Context, projectName string, clusterConfi
 		}
 	}
 
+	// Uninstall the GPU Operator before tofu destroy. It has no cloud resources
```
dcmcand: **Nice:** this `Destroy` block faithfully mirrors the Longhorn best-effort pattern (kubeconfig-unavailable warns and skips, uninstall failure honors `--force`), and running before `tofu destroy` is consistent with the established teardown order.

**Stylistic:** one small difference from the Longhorn block it mirrors is that Longhorn is gated behind `if awsCfg.LonghornEnabled()`, whereas this runs unconditionally on every destroy (kubeconfig fetch + Helm history lookup even for clusters that never had GPU nodes). Harmless, since `uninstallGPUOperator` no-ops on `ErrReleaseNotFound`, and leaving it unconditional actually doubles as the only path that removes the operator after GPU node groups are dropped from config. If that is deliberate, a one-line comment would help; otherwise gating on `hasGPUNodeGroups()` would match Longhorn.

pkg/providers/cluster/aws/gpu_operator.go
```diff
@@ -0,0 +1,255 @@
...
+	gpuOperatorRepoName  = "nvidia"
+	gpuOperatorRepoURL   = "https://helm.ngc.nvidia.com/nvidia"
+	gpuOperatorChartName = "nvidia/gpu-operator"
+	// gpuOperatorChartVersion pins the upstream NVIDIA gpu-operator Helm chart.
+	// The NGC registry lists this chart with a leading "v" (e.g. "v26.3.2"), so
+	// the prefix is intentional and required for LocateChart to resolve it -
+	// unlike the bare-semver aws-load-balancer-controller chart. Check the NGC
+	// catalog (https://catalog.ngc.nvidia.com/orgs/nvidia/helm-charts/gpu-operator)
+	// for the current version when bumping.
+	gpuOperatorChartVersion   = "v26.3.2"
```
dcmcand, line 31: **Stylistic / heads-up.** `v26.3.2` is valid and current enough; note `v26.3.3` is now the latest published `gpu-operator` chart on NGC, one patch ahead. No need to bump in this PR, just a note for the next version bump.

dcmcand (changes requested):
Thanks for this. The GPU operator implementation itself is clean, well-researched, and a faithful match to the established AWS addon pattern (cluster-autoscaler / LBC / Longhorn). Build, vet, lint, and the new table-driven tests all pass, and I verified the AWS-specific Helm values (driver/toolkit off, devicePlugin on, `MOFED_ENABLED=false`) against AWS's EKS device-management docs. They are correct, including MOFED off to avoid the device-plugin/EFA `uverbs` conflict.

I'm requesting changes for one reason, and it is not the code:

### Blocker: the PR description does not match what this PR ships

The description says this PR implements `reconcileGPUOperator` ("installs when the cluster has GPU nodes, uninstalls otherwise"), `isGPUInstanceType` (a live-node instance-type check), and removes `pkg/provider/gpu.go`, `InfraSettings.GPUOperator`, `ResolveGPUOperator`, and the ArgoCD template. None of that is in the diff. It is +393/-0 across 5 files, and none of those symbols or deletions exist at this commit.

What actually ships is narrower and reasonable: a config-flag-gated install (`hasGPUNodeGroups()`, i.e. node groups with `gpu: true`), no live-node detection, and no uninstall-on-disable during `Deploy` (only a full `Destroy` removes it). That narrower scope is consistent with ADR-0006, which explicitly defers the reconcile/prune-on-disable loop to the shared `ConditionalCharts` interface in #349 and names this PR as a hand-rolled precursor.

Please reconcile the two before merge, either:
- (a) trim the description to match the shipped behavior (config-gated install only, no live detection, no auto-uninstall-on-disable), or
- (b) implement the described `reconcileGPUOperator` / `isGPUInstanceType` behavior.

Either is fine; (a) is the lighter path and stays within ADR-0006's sanctioned interim shape. Why it matters beyond tidiness: #349 plans to fold this PR in as a reference instance, so a description claiming live-detection and disable-time uninstall already exist would mislead that follow-up.

### Verified, no action needed
- The install runs before the ArgoCD/GitOps stage (`pkg/nic/deploy.go`).
- The extra `GetKubeconfig` calls hit the existing per-(project, region) cache, so they do not add load against #241.
- The chart source and pin (`nvidia/gpu-operator` `v26.3.2` from `helm.ngc.nvidia.com`) is legitimate, and scope is tight (no `go.mod` changes, no `init()`, no new exec/network surface).

Nice touch using `RunWithContext` where the two older sibling installers still use `Run` and silently drop the caller's context.

### Issue #232
Partially resolved by design. The core "`nvidia.com/gpu` isn't advertised" problem is fixed for the declare-`gpu: true`-and-deploy path; live detection and prune-on-disable are deferred to #349 per ADR-0006.

pkg/providers/cluster/aws/gpu_operator.go
viniciusdc: Agreed - changed the teardown uninstall to `Wait=false`, and in `Destroy` an uninstall failure now warns and continues even without `--force` (it can't block teardown, and the nodes are about to go anyway). Longhorn keeps the `--force` gate because its CSI finalizers genuinely can wedge node-group deletion; the GPU operator has no such resources.

pkg/providers/cluster/aws/config.go
viniciusdc: Done - exported as `HasGPUNodeGroups` to match `LonghornEnabled` / `ClusterAutoscalerEnabled` / `LoadBalancerControllerEnabled`.

pkg/providers/cluster/aws/provider.go
viniciusdc: Gated it on `HasGPUNodeGroups()` to match the Longhorn block, rather than the unconditional version. On a full `Destroy` the "removal after config-drop" path is moot (the cluster is being deleted anyway), so the round-trip on non-GPU clusters wasn't buying anything. Also made the failure non-fatal (see the Wait thread).

pkg/providers/cluster/aws/gpu_operator.go
viniciusdc: Thanks - leaving it at `v26.3.2` for this PR as you suggested; noted `v26.3.3` for the next bump.

viniciusdc:
Thanks @dcmcand - you're right, the description had drifted from what actually ships. That was my miss: I updated the code to the config-gated install last round but never re-synced the description, so it still described the old `reconcileGPUOperator` / `isGPUInstanceType` / live-detection version and claimed deletions that aren't in this (`+393/-0`) diff.

Took option (a): rewrote the description to match the shipped behavior - config-flag-gated install (`HasGPUNodeGroups()`), no live detection, no uninstall-on-disable during Deploy (only Destroy removes it), and an explicit "what this deliberately does NOT do" section pointing the reconcile/prune loop at #349 per ADR-0006. Issue #232 marked partially resolved by design.

Also folded in the inline items: `Wait=false` + non-fatal teardown uninstall, exported `HasGPUNodeGroups`, and gated the Destroy uninstall on it to mirror Longhorn. Left the chart at `v26.3.2` (noted `v26.3.3` for next bump). Build/vet/lint/tests green.

pkg/providers/cluster/aws/gpu_operator.go
```diff
@@ -0,0 +1,257 @@
...
+	loadedChart, err := loadGPUOperatorChart(client.ChartPathOptions)
+	if err != nil {
+		span.RecordError(err)
+		return err
+	}
+
+	// RunWithContext, not Run: Install.Run builds its own context.Background()
+	// and ignores ours, so a cancellation/timeout during the install window
+	// wouldn't propagate to Helm.
+	release, err := client.RunWithContext(ctx, loadedChart, gpuOperatorHelmValues())
```
dcmcand, line 144: Nice catch using `RunWithContext` here. The older LBC and autoscaler installers use plain `Run`, which silently drops the caller's context, so this genuinely propagates cancellation and timeouts where they do not. The comment above it explaining the Helm-internal `context.Background()` is a helpful touch.

pkg/providers/cluster/aws/gpu_operator.go
```diff
@@ -0,0 +1,257 @@
...
+	gpuOperatorRepoName  = "nvidia"
+	gpuOperatorRepoURL   = "https://helm.ngc.nvidia.com/nvidia"
+	gpuOperatorChartName = "nvidia/gpu-operator"
+	// gpuOperatorChartVersion pins the upstream NVIDIA gpu-operator Helm chart.
+	// The NGC registry lists this chart with a leading "v" (e.g. "v26.3.2"), so
+	// the prefix is intentional and required for LocateChart to resolve it -
+	// unlike the bare-semver aws-load-balancer-controller chart. Check the NGC
+	// catalog (https://catalog.ngc.nvidia.com/orgs/nvidia/helm-charts/gpu-operator)
+	// for the current version when bumping.
+	gpuOperatorChartVersion   = "v26.3.2"
```
dcmcand, line 31: The `v` prefix is easy to get wrong and the comment documents exactly why it is needed. Confirmed on NGC that `gpu-operator-v26.3.2.tgz` resolves while the bare version 404s, so this pin is correct.

pkg/providers/cluster/aws/gpu_operator.go
```diff
@@ -0,0 +1,257 @@
...
+// When the operator grows to other providers these values will need to vary by
+// provider; for now they are AWS-specific.
+func gpuOperatorHelmValues() map[string]any {
+	return map[string]any{
+		"driver":  map[string]any{enabledKey: false},
+		"toolkit": map[string]any{enabledKey: false},
+		"devicePlugin": map[string]any{
+			enabledKey: true,
+			"env": []any{
+				map[string]any{"name": "MOFED_ENABLED", "value": "false"},
```
dcmcand, line 57: `MOFED_ENABLED=false` is the right default for EFA-enabled nodes and the comment ties it directly to the device-plugin/uverbs conflict. Good to see the rationale captured in-line rather than left as a magic value.

pkg/providers/cluster/aws/provider.go
```diff
@@ -640,6 +657,25 @@ func (p *Provider) Destroy(ctx context.Context, projectName string, clusterConfi
 		}
 	}
 
+	// Uninstall the GPU Operator before tofu destroy, gated on the GPU config
+	// flag to mirror the Longhorn block above. The operator has no cloud
+	// resources that can block teardown, so an uninstall failure is never fatal
+	// to the destroy: we warn and continue even without --force (unlike
+	// Longhorn, whose CSI finalizers can wedge node-group deletion). Best-effort.
+	if awsCfg.HasGPUNodeGroups() {
```
dcmcand, line 665: The best-effort teardown is well-reasoned: warning and continuing (even without `--force`) is correct here since the operator has no cloud resources to drain, and the comment contrasting it against Longhorn's CSI finalizers makes the divergence a conscious choice rather than an oversight.

dcmcand (approved):
Approving. This is a clean, well-reasoned piece of work and it was a pleasure to read.

The things that stood out:

- **It respects the architecture instead of fighting it.** `gpu_operator.go` is a faithful sibling of `cluster_autoscaler.go` and `aws_load_balancer_controller.go` (values builder, chart loader, install-with-history-check, upgrade, gating `Config` predicate), all GPU logic stays inside the AWS provider package, and you correctly did NOT add an `InfraSettings` field since nothing outside the provider branches on this. ADR-0006 does the load-bearing work of justifying imperative Helm over GitOps here, and the code matches the ADR exactly.
- **`RunWithContext` on install/upgrade is a real improvement over the older siblings, and the comment explaining why uninstall stays on `Run` checks out against the Helm version in the tree.**
- **The details are right and were clearly verified, not guessed:** the `v`-prefixed chart pin resolves on NGC (the bare version 404s), `CreateNamespace=true` is the correct divergence for a dedicated namespace, `MOFED_ENABLED=false` implements the exact EFA/uverbs mitigation from the issue, DryRun is properly bypassed, and the extra `GetKubeconfig` call is a cache hit rather than a redundant EKS describe.
- **The scope discipline is excellent.** The "what this deliberately does NOT do" section matches ADR-0006 and #349, so the deferrals are honest and traceable rather than silent gaps.

Non-blocking follow-ups for whenever you get to them (none of these hold up the merge):

1. Have you validated the 10-minute install timeout against a cold GPU node group coming up? `Wait=true` blocks on the device-plugin rollout and NVIDIA image pulls can be slow. Fail-closed is the right behavior, I just cannot tell the right ceiling without your test data.
2. Rollout sequencing with packs that currently ship their own GPU operator (per #232): worth confirming a `gpu: true` cluster never ends up running two device plugins during the transition.
3. Consider pulling a simple `enabled: false` escape hatch forward into this PR rather than #349. Right now a `gpu: true` cluster cannot opt out or override the chart version without editing Go, which is the one deferred item with a real near-term cost during the pack migration. Small change, mirrors the sibling config structs.

Great work.

PR #350, opened 2026-06-02
feat(keycloak): trust org CA bundle for outbound TLS

tylerpotts:
## Summary
Implements https://github.com/nebari-dev/nebari-infrastructure-core/issues/311 (child of https://github.com/nebari-dev/nebari-infrastructure-core/issues/307).

When a top-level `trust_bundle` is configured, Keycloak now trusts the org CA for all outbound TLS (IdP federation, token introspection):
- Mounts the trust-manager-projected `nebari-trust-bundle` ConfigMap (key `ca-certificates.crt`) into the Keycloak pod at `/etc/nebari/truststore`.
- Sets `KC_TRUSTSTORE_PATHS=/etc/nebari/truststore` so Keycloak loads it into its native truststore.

The issue suggested `X509_CA_BUNDLE`, but that is a WildFly-image-only feature; the chart here (`keycloakx` 7.1.6) ships Keycloak 26.5.0 (Quarkus), which ignores it. The native `--truststore-paths` / `KC_TRUSTSTORE_PATHS` mechanism is used instead — the fallback the issue explicitly anticipated.

All wiring is gated on `TrustManagerEnabled`, so deploys without a `trust_bundle` are unchanged.

> Note: stacked on `feat/trust-manager-309` (open PR #346), which adds the trust-manager Bundle/ConfigMap this depends on and is not yet in `main`.

## Test plan
- [x] `go test ./pkg/argocd/` — added `TestKeycloakTemplate_TrustBundle` (asserts mount/env present when enabled, absent when disabled; validates rendered Application + inner Helm values parse as YAML).
- [x] `golangci-lint run ./pkg/argocd/` — clean.
- [x] End-to-end on a local kind cluster:
  - trust-manager projects `nebari-trust-bundle` into the `keycloak` namespace.
  - Keycloak StatefulSet carries `KC_TRUSTSTORE_PATHS` + the volume mount; the CA file (`CN = Test Org Root CA`) is present in-pod.
  - Startup log: `TruststoreBuilder ... Found the following truststore files under .../etc/nebari/truststore`.
  - Outbound TLS to an in-cluster HTTPS endpoint whose cert chains to the org CA → import succeeds (HTTP 200).
  - Control: outbound TLS to an untrusted self-signed endpoint → fails with `PKIX path building failed`, confirming the truststore is doing real work.

### DIFF
--- a/pkg/argocd/templates/apps/keycloak.yaml
+++ b/pkg/argocd/templates/apps/keycloak.yaml
@@ -51,6 +51,14 @@ spec:
               value: "https://keycloak.{{ .Domain }}{{ .KeycloakBasePath }}"
             - name: KC_FEATURES
               value: "token-exchange,admin-fine-grained-authz:v1"
+{{- if .TrustManagerEnabled }}
+            # Keycloak 26 (Quarkus) ignores X509_CA_BUNDLE; the org CA is loaded
+            # via the native truststore. KC_TRUSTSTORE_PATHS points at the
+            # trust-manager-projected bundle mounted below and feeds every
+            # outbound TLS call (IdP federation, token introspection).
+            - name: KC_TRUSTSTORE_PATHS
+              value: /etc/nebari/truststore
+{{- end }}
           proxy:
             mode: xforwarded
           livenessProbe: |
@@ -91,6 +99,19 @@ spec:
             limits:
               cpu: 2000m
               memory: 2Gi
+{{- if .TrustManagerEnabled }}
+          extraVolumes: |
+            - name: nebari-trust-bundle
+              configMap:
+                name: nebari-trust-bundle
+                items:
+                  - key: ca-certificates.crt
+                    path: ca-certificates.crt
+          extraVolumeMounts: |
+            - name: nebari-trust-bundle
+              mountPath: /etc/nebari/truststore
+              readOnly: true
+{{- end }}
     # Source 2: Realm setup job (PostSync hook)
     - repoURL: {{ .GitRepoURL }}
       targetRevision: {{ .GitBranch }}
--- a/pkg/argocd/writer_test.go
+++ b/pkg/argocd/writer_test.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"testing"
 
+	"gopkg.in/yaml.v3"
+
 	"github.com/nebari-dev/nebari-infrastructure-core/pkg/config"
 	"github.com/nebari-dev/nebari-infrastructure-core/pkg/provider"
 )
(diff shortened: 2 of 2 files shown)


viniciusdc:
#336 landed in `main`. Note this one is stacked on #346 (`feat/trust-manager-309`) rather than on #336 directly, so the retarget to `main` here follows once #346 merges

pkg/argocd/templates/apps/keycloak.yaml
```diff
@@ -91,6 +99,19 @@ spec:
             limits:
               cpu: 2000m
               memory: 2Gi
+{{- if .TrustManagerEnabled }}
+          extraVolumes: |
```
dcmcand, line 103: Stylistic / non-blocking: this configMap volume is non-optional, and the Bundle that produces `nebari-trust-bundle` (`trust-bundle.yaml`) is sync-wave 4 - the same wave as this Keycloak app (line 10). On a first sync the projected ConfigMap is not guaranteed to exist in the namespace before the Keycloak pod is scheduled, so the pod can sit in `ContainerCreating` briefly.

It self-heals, which is why it is not a blocker: trust-manager is wave 3 and healthy before wave 4 starts, it projects the bundle independently of Keycloak's pod state, kubelet retries the failed mount, and ArgoCD waits for the whole wave to go healthy before advancing. Worst case is a one-time first-install delay; on steady state the ConfigMap already exists.

Keeping the volume non-optional is the right fail-closed choice for a trust control (better than starting Keycloak silently without the org CA). If you want to close the timing window, bump this app to sync-wave 5 so it runs after the Bundle, or mark the volume `optional`. Your call - either is fine.

dcmcand (approved):
Approving. Clean, well-scoped change and the mechanism is right.

Verified the cross-PR contract against the base branch (`feat/trust-manager-309`): the trust-manager Bundle is named `nebari-trust-bundle` with target key `ca-certificates.crt` and `namespaceSelector: {}`, so the projected ConfigMap and key this mounts will actually exist in the keycloak namespace. The gating is tight - `TrustManagerEnabled` is true iff a `trust_bundle` is configured, so the mount can never reference a missing ConfigMap and no-bundle deploys render unchanged.

Good call using the native `KC_TRUSTSTORE_PATHS` over `X509_CA_BUNDLE` (WildFly-only; Keycloak 26 / Quarkus ignores it), and the test parsing both the Application and the inner keycloakx Helm values as YAML is exactly the right guard against block-scalar indentation regressions.

One non-blocking note inline about sync-wave ordering vs the non-optional volume. Like #353, this is stacked on #346; retarget to `main` once that merges.

PR #352, opened 2026-06-03
Support autoscaling in AWS provider

marcelovilla:
## Closes

Closes #347

## Description

This PR adds autoscaling support in the AWS provider by installing the [cluster autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/charts/cluster-autoscaler) helm chart by default. Users can opt-out by disabling it in the config.

It also adds a new value in the longhorn helm chart so autoscaling is recognized by the `instance-manager` pod and scaling down works by marking it safe to evict.

## How to Test

To test this PR, deploy a cluster to AWS with a similar config to this:

```yaml
cluster:
  aws:
    region: us-west-2
    kubernetes_version: "1.35"
    availability_zones:
      - us-west-2a
      - us-west-2b
    vpc_cidr_block: "10.10.0.0/16"
    endpoint_private_access: true
    endpoint_public_access: true

    node_groups:
      user:
        instance: m7i.2xlarge
        min_nodes: 1
        max_nodes: 5

      workload:
        instance: c5.xlarge
        min_nodes: 0
        max_nodes: 10
        labels:
          workload-class: batch
        taints:
          - key: workload-class
            value: batch
            effect: NO_SCHEDULE
```

Note that the `workload` node group has labels, taints, and will initially have zero nodes. The idea is to test that scale from zero works, and users can use arbitrary labels/tolerations in their manifests.

Once the cluster is up and running, apply the following manifest:

```bash
kubectl apply -f autoscale-test.yaml
```

```yaml
# autoscale-test.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: autoscale-test
  namespace: default
  labels:
    app: autoscale-test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: autoscale-test
  template:
    metadata:
      labels:
        app: autoscale-test
    spec:
      nodeSelector:
        workload-class: batch
      tolerations:
        - key: workload-class
          operator: Equal
          value: batch
          effect: NoSchedule
      terminationGracePeriodSeconds: 0
      containers:
        - name: pause
          image: registry.k8s.io/pause:3.9
          resources:
            requests:
              cpu: 100m
              memory: 64Mi
            limits:
              cpu: 100m
              memory: 64Mi
```

After one or two minutes, you should see a new node coming up to accommodate the `autoscale-test` pod. Make sure the new node matches the node group with the label/taint from the config.

Once that's verified, you can remove the deployment (as in the k8s deployment, not the Nebari/cluster deployment) by running:
```bash
kubectl delete -f autoscale-test.yaml
```

After 10-15 minutes, the node will be scaled down as it doesn't have any workload on it.

If you also want to test that nodes with longhorn running on them scale down properly, remove the taint from the nodegroup, re-deploy, and repeat. You can also create another node group without a toleration and modify the deployment manifest (i.e. autoscale-test.yaml) to use another node selector.


pkg/provider/aws/templates/main.tf
```diff
@@ -1,6 +1,10 @@
 module "eks_cluster" {
-  source  = "nebari-dev/eks-cluster/aws"
-  version = "0.5.0"
+  # Intentionally pinned to the feat/cluster-autoscaler branch rather than the
+  # released registry version: the branch carries changes we depend on that are
+  # not yet in a tagged release of nebari-dev/eks-cluster/aws.
+  #   source  = "nebari-dev/eks-cluster/aws"
+  #   version = "0.5.0"
+  source = "git::https://github.com/nebari-dev/terraform-aws-eks-cluster.git?ref=feat/cluster-autoscaler"
```
marcelovilla: Need to update this to point to a new version of the TF module once https://github.com/nebari-dev/terraform-aws-eks-cluster/pull/34 has been merged and a release has been cut.

.golangci.yml
```diff
@@ -28,6 +28,10 @@ linters:
       # Default is 3, which flags many genuinely-incidental repeats.
       # Bump to 4 so only repeated strings that clearly warrant a constant get flagged.
       min-occurrences: 4
+      # Don't count occurrences in _test.go toward the threshold. Test files are
+      # already excluded from reporting below, but without this they still
+      # inflate the ocurrence count for production code.
+      ignore-tests: true
```
marcelovilla, line 34: Added this because `goconst` was complaining about too many occurrences for the `"clusterName"` string. That's a map key and I'd rather not extract is a const. Makes it less readable in my opinion.

pkg/storage/longhorn/install.go
```diff
@@ -204,6 +204,12 @@ func buildHelmValues(cfg *Config) map[string]any {
 		"replicaAutoBalance":          "best-effort",
 	}
 
+	// Only render the autoscaler setting when a provider explicitly sets it.
+	// Leaving it unset keeps Longhorn's default.
+	if cfg != nil && cfg.ClusterAutoscalerEnabled != nil {
+		settings["kubernetesClusterAutoscalerEnabled"] = *cfg.ClusterAutoscalerEnabled
```
marcelovilla, line 210: I added this because without it, the longhorn `instance-manager` pod was causing nodes without any workloads to not be scaled down. I think the reason is that every other pod that we run on other nodes than the main one are daemonsets and this value adds a safe to evict annotation.

pkg/provider/aws/aws_load_balancer_controller.go
```diff
@@ -42,6 +42,9 @@ func awsLoadBalancerControllerHelmValues(cfg *Config, clusterName, vpcID string)
 			"create": true,
 			"name":   lbcServiceAccount,
 		},
+		// Reuse the existing webhook TLS secret across upgrades instead of
+		// regenerating it.
+		"keepTLSSecret": true,
```
marcelovilla, line 47: I was running into the following error when running `nic deploy` after including the changes to install the helm chart:
```
Error: deploy infrastructure: failed to install Cluster Autoscaler: failed to install Cluster Autoscaler: 1 error occurred:
        * Internal error occurred: failed calling webhook "mservice.elbv2.k8s.aws": failed to call webhook: Post "https://aws-load-balancer-webhook-service.kube-system.svc:443/mutate-v1-service?timeout=10s": tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "aws-load-balancer-controller-ca")
```

I think the problem was that the cluster autoscaler helm chart deploys a k8s service, and it seems every service is intercepted by the LB controller's mutating hook. Every helm upgrade we run on the LB controller's helm chart (i.e. every `nic deploy`), regenerates its self-signed webhook cert and this leads to drift, which is what made the cluster autoscaler deployment to fail on a non-fresh deployment. Passing `keepTLSSecret: true` in the LB controller Helm Chart makes use of the existing one instead of regenerating it.

oren-openteams:
## Field test feedback after running this branch

Spent some time today with this branch in a real cluster. The cluster autoscaler install itself works as intended — IRSA wiring is correct, the `cluster-autoscaler-aws-cluster-autoscaler` deployment comes up cleanly, scale-up triggers on Pending pods and scale-down on idle nodes. Passing `kubernetesClusterAutoscalerEnabled` through to the Longhorn chart (bb3e4ae) is the right call — without it, Longhorn would block all scale-down by refusing replica evictions.

Two interaction surfaces worth surfacing before this merges, because they bit in non-obvious ways:

### 1. Longhorn replica destruction on idle scale-down

Filed at #354 with [a follow-up comment](https://github.com/nebari-dev/nebari-infrastructure-core/issues/354#issuecomment-4616573613) covering this exact case. Short version: when a nodegroup has `min_nodes: 0` (the default for `user`-style groups) and autoscaler scales it all the way to 0, every Longhorn replica on those terminating nodes dies with them. With `replica_count: 2` (Longhorn default) on a small cluster, every scale-down event has a near-certain chance of destroying singleuser PVCs.

This isn't a bug in *this* PR — it's a Longhorn lifecycle issue that #354 already tracks. But this PR meaningfully changes its severity: scale-downs happen continuously now rather than only at `nic deploy` time, turning a deploy-time hazard into a steady-state one. When this lands, the changelog should call out the `min_nodes` requirement for any nodegroup hosting Longhorn-backed PVCs.

### 2. GPU node + device plugin race on cold-start scale-up

When the autoscaler brings a GPU nodegroup up from 0, there's a ~60-90s window where the node is `Ready` but the NVIDIA device plugin hasn't yet advertised `nvidia.com/gpu` capacity. During that window the Pending pod that triggered the scale-up still can't schedule. With a short `scale-down-unneeded-time`, the autoscaler can reap the node before the pod ever lands → loop.

In our case it eventually worked once we also pinned `min_nodes: 1` on the GPU nodegroup, but the failure shape is real. Two mitigations to consider for sane defaults:

- Default `min_nodes: 1` on any nodegroup with `gpu: true` (or document the cold-start race loudly)
- Bump the cluster autoscaler's `scale-down-unneeded-time` from its 10m default so device-plugin installs have more grace

### Suggestion

The PR description (rightly) focuses on the wiring. Before merge, a short "Operational expectations" section in the changelog / README would help users avoid these footguns:

- `min_nodes` should be ≥ Longhorn `replica_count` on any nodegroup hosting persistent storage
- GPU nodegroups should default to `min_nodes: 1` unless cold-start races are acceptable
- `scale-down-unneeded-time` and `scale-down-delay-after-add` considerations for device-plugin-dependent workloads

Happy to draft that prose if useful. Otherwise this is great — the actual wiring is clean and the Longhorn flag flow-through is exactly right.

dcmcand (approved):
Approving. The autoscaler wiring is clean and well-tested, and the one blocker from review is resolved.

Transparency: I pushed two commits to this branch:
- `59081f0` regenerates the OpenTofu lockfiles to clear the `drift` CI failure (aws `6.47.0 -> 6.49.0` from your `.tf` changes; azurerm `4.75.0 -> 4.76.0` came along because the drift job regenerates and checks every template lockfile in one run).
- `388434a` fixes the blocker: the Longhorn-install path was aliasing and mutating the caller-owned `awsCfg.Longhorn`. It now goes through a nil-safe `longhorn.Config.WithClusterAutoscalerEnabled` that returns a copy, with a unit test covering the no-mutation and nil-receiver paths.

Verified while reviewing:
- Scale-from-zero with label/taint node groups works without manual `node-template` ASG tags: these are EKS managed node groups, cluster-autoscaler auto-discovers their labels/taints via `eks:DescribeNodegroup`, and the module's pod-identity policy (eks-pod-identity 2.7.0) grants it. Doc caveat: CA only calls `DescribeNodegroup` for a group created at 0 that never had nodes.
- The `clusterName` guard sits before the release-history check, so both install and upgrade paths are covered.

Two non-blocking follow-ups, your call:
1. The autoscaler installs by default (`ClusterAutoscalerEnabled()` returns true when unset, per `c086d42`). Deliberate and consistent with Longhorn, but existing clusters get a surprise Helm install on the next `nic deploy`. Worth a changelog/migration note.
2. The Helm install uses `Wait=true` with a 5-minute timeout. If the Pod Identity association has not propagated yet, the first deploy could time out. `nic deploy` is idempotent so a retry recovers, but a readiness/retry guard would smooth the first run.

Merge-order heads-up with #356: that PR switches the taint-effect validator from `NoSchedule` to the EKS enum `NO_SCHEDULE`. Whichever merges second needs a small conflict resolution in `provider.go`'s taint validation.

Issue #138, opened 2026-03-05
Check helm release status before skipping upgrade


dcmcand:
## Context

In #85, the ArgoCD install logic was updated to skip `helm upgrade` when the deployed chart version matches the target version. The skip check compares `releases[0].Chart.Metadata.Version == config.Version`.

However, `histClient.Max = 1` returns only the most recent release, which could be in a failed state (e.g., a previous upgrade attempt that failed). In that case, `releases[0].Chart.Metadata.Version` reflects the *target* version of the failed attempt, not what's actually running. The skip logic would see a version match and skip the upgrade, leaving ArgoCD in a broken state.

## Proposed fix

Check `releases[0].Info.Status` before comparing versions. If the release is in a failed or pending state, always proceed with the upgrade regardless of version match.

```go
if current.Info.Status == release.StatusDeployed &&
    current.Chart != nil && current.Chart.Metadata != nil &&
    current.Chart.Metadata.Version == config.Version {
    // safe to skip
}
```

## References

- PR #85 review comment: https://github.com/nebari-dev/nebari-infrastructure-core/pull/85#pullrequestreview-3894945519
- Relevant code: `pkg/argocd/helm.go` line 86-99

PR #353, opened 2026-06-03
feat(argocd): trust org CA bundle in the repo-server

viniciusdc:
## Summary

Closes #310. The in-pod-trust slice of epic #307 for the ArgoCD **repo-server** — the component that clones git repos and pulls Helm/OCI charts through a TLS-inspecting egress proxy.

When a top-level `trust_bundle` is configured, this:
- creates an install-time `argocd-org-ca` ConfigMap in the `argocd` namespace from the resolved PEM, and
- bakes repo-server Helm values that mount it, run an init container combining the image's **system** CA bundle with the **org** CA into a shared `emptyDir`, and set `SSL_CERT_FILE` / `GIT_SSL_CAINFO` / `CURL_CA_BUNDLE` at the combined file (all three are needed — git ignores `SSL_CERT_FILE` alone).

Gated on `trust_bundle` being set; otherwise the repo-server is unchanged. The wiring lives behind a single `configureRepoServerCATrust` seam (no-op when unset), with the ConfigMap built by a `createOrgCAConfigMap` helper beside the other foundational create helpers — so `Install` doesn't accrete per-feature blocks.

## Why this diverges from the issue's proposed-change text

The issue suggested mounting **trust-manager's projected ConfigMap** and populating `argocd-tls-certs-cm`. Both are deliberately avoided (endorsed by the maintainer comments on #310):

1. **Install-time ConfigMap, not the projected one — bootstrap ordering.** The repo-server is what *pulls trust-manager's chart through the proxy*. trust-manager's projected `nebari-trust-bundle` ConfigMap doesn't exist until the repo-server has already synced trust-manager, which it can't do without CA trust. So the repo-server must get its CA at install time. (Keycloak — #350 — can use the projected CM because it syncs *after* trust-manager; the repo-server is upstream of it.)
2. **`argocd-tls-certs-cm` left empty on purpose.** Entries there make Argo pass `--ca-file` to git/helm, which *replaces* (not augments) the system trust pool and breaks cross-host redirects. The combined-bundle env approach is additive (system ∪ org) and is the field-proven path.

## Scope

- **repo-server only.** The application-controller's egress is the in-cluster API (validated against the service-account `ca.crt`, unaffected) and gRPC to the repo-server; it doesn't clone/fetch, so it doesn't need the org CA. If a controller-side egress case ever surfaces it can get the identical mount — noted as a follow-up, not done here.

## Notes for reviewers

- The `argocd-org-ca` ConfigMap is **upserted** (`createOrUpdateConfigMap`), not create-only like `createSecret`, so a rotated `trust_bundle` actually propagates; managed labels are reconciled on update, and a non-`NotFound` Get error is surfaced rather than masked as "absent".
- The init container runs with a hardened `securityContext` (`runAsNonRoot`, `readOnlyRootFilesystem`, drop `ALL`, `seccompProfile: RuntimeDefault`); it only writes to the shared `emptyDir`.
- The init image is the chart's own repo-server image, resolved via the chart's `tpl` pass over `repoServer.initContainers` (no hardcoded tag) — see the `helm template` caveat in the test plan.
- An `org_ca_configured` span attribute records whether trust was wired, so the skip path is observable.

## ⚠️ Known limitation — applies to fresh installs; existing clusters need a separate change

This wires the CA into the ArgoCD **Helm values**. `shouldSkipUpgrade` (added in #85) compares **chart version only**, so on a cluster where ArgoCD is already `Deployed` at the pinned version, a values-only change is **not applied** until `defaultChartVersion` is bumped. This PR intentionally does **not** touch `shouldSkipUpgrade` — making the skip logic values-aware is a cross-cutting change to the upgrade engine that affects every `nic deploy` and deserves its own issue/PR. Until that lands, this takes effect on fresh installs (or after a version bump). Tracked in #364.

## Stacking

Based on `feat/trust-manager-309` (#346), which adds the top-level `trust_bundle` config and `ResolvePEM()` this depends on. **Retarget to `main` once #346 merges.**

## Test plan

- [x] `go build`, `go vet`, `go test ./pkg/argocd/... ./pkg/nic/...` — pass.
- [x] `TestAddOrgCAMount` — injected volumes, init container (command, chart-resolved image, securityContext) and the three env vars asserted against literal paths; `TestAddOrgCAMountPreservesAndCoexists` — preserves existing `repoServer` keys and coexists with the local-gitops mount.
- [x] `TestConfigureRepoServerCATrust` — gating (empty bundle → no ConfigMap, values untouched), the foundational `part-of`/`managed-by` labels on the created ConfigMap, and that the repo-server values are wired.
- [x] `TestCreateOrUpdateConfigMap` — create / **update-on-rotation** (upsert, unlike create-only `createSecret`) / no-op-when-unchanged, the last asserting **no write action** is issued so the early-return is proven.
- [ ] **e2e (maintainer step, needs a TLS-inspected cluster):** confirm `git clone` of an HTTPS repo terminated by the proxy succeeds from the repo-server, and that `/etc/ssl/certs/ca-certificates.crt` exists in the pinned ArgoCD image. The unit tests prove the values map is built correctly; they do **not** prove the chart honors it (the `tpl`-based init image in particular) or that git trusts the file — a `helm template` render assertion is a candidate to narrow that gap.

Refs #307.

### DIFF
--- a/pkg/argocd/foundational.go
+++ b/pkg/argocd/foundational.go
@@ -32,6 +32,9 @@ const (
 	// NebariLandingRedisSecretName is the name of the Kubernetes secret containing Redis password for nebari-landing.
 	NebariLandingRedisSecretName = "nebari-landing-redis" //nolint:gosec // This is a secret name reference, not a credential
 
+	// PartOfLabel is the app.kubernetes.io/part-of label key.
+	PartOfLabel = "app.kubernetes.io/part-of"
+
 	// NebariFoundationalPartOf is the value of the app.kubernetes.io/part-of label for foundational resources.
 	NebariFoundationalPartOf = "nebari-foundational"
 
@@ -287,8 +290,8 @@ func createOrgCAConfigMap(ctx context.Context, client kubernetes.Interface, name
 			Name:      orgCAConfigMapName,
 			Namespace: namespace,
 			Labels: map[string]string{
-				"app.kubernetes.io/part-of":    NebariFoundationalPartOf,
-				"app.kubernetes.io/managed-by": "nebari-infrastructure-core",
+				PartOfLabel:    NebariFoundationalPartOf,
+				ManagedByLabel: NebariManagedByValue,
 			},
 		},
 		Data: map[string]string{orgCAConfigMapKey: orgCABundlePEM},
@@ -371,8 +374,8 @@ func createKeycloakSecrets(ctx context.Context, client kubernetes.Interface, key
 				Name:      "nebari-realm-admin-credentials",
 				Namespace: namespace,
 				Labels: map[string]string{
-					"app.kubernetes.io/part-of": NebariFoundationalPartOf,
-					ManagedByLabel:              NebariManagedByValue,
+					PartOfLabel:    NebariFoundationalPartOf,
+					ManagedByLabel: NebariManagedByValue,
 				},
 			},
 			Type: corev1.SecretTypeOpaque,
@@ -392,8 +395,8 @@ func createKeycloakSecrets(ctx context.Context, client kubernetes.Interface, key
 				Name:      "argocd-oidc-client-secret",
 				Namespace: namespace,
 				Labels: map[string]string{
-					"app.kubernetes.io/part-of": NebariFoundationalPartOf,
-					ManagedByLabel:              NebariManagedByValue,
+					PartOfLabel:    NebariFoundationalPartOf,
+					ManagedByLabel: NebariManagedByValue,
 				},
 			},
 			Type: corev1.SecretTypeOpaque,
@@ -419,8 +422,8 @@ func createLandingPageSecrets(ctx context.Context, client kubernetes.Interface, 
 			Name:      NebariLandingRedisSecretName,
 			Namespace: namespace,
 			Labels: map[string]string{
-				"app.kubernetes.io/part-of": NebariFoundationalPartOf,
-				ManagedByLabel:              NebariManagedByValue,
+				PartOfLabel:    NebariFoundationalPartOf,
+				ManagedByLabel: NebariManagedByValue,
 			},
 		},
 		Type: corev1.SecretTypeOpaque,
--- a/pkg/argocd/foundational.go
+++ b/pkg/argocd/foundational.go
@@ -8,6 +8,7 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
@@ -221,15 +222,20 @@ func createNamespace(ctx context.Context, client kubernetes.Interface, namespace
 	return nil
 }
 
-// createOrUpdateConfigMap creates the ConfigMap, or updates its data when it
-// already exists with different contents. Unlike createSecret (create-only, for
-// generated one-time credentials), an org CA bundle is operator-supplied and
-// rotates, so this upserts to make a changed trust_bundle actually propagate.
+// createOrUpdateConfigMap creates the ConfigMap, or reconciles its data and
+// labels when it already exists with different contents. Unlike createSecret
+// (create-only by design, for generated one-time credentials), an org CA bundle
+// is operator-supplied and rotates, so this upserts to make a changed
+// trust_bundle actually propagate.
 func createOrUpdateConfigMap(ctx context.Context, client kubernetes.Interface, cm *corev1.ConfigMap) error {
 	namespace := cm.Namespace
 	existing, err := client.CoreV1().ConfigMaps(namespace).Get(ctx, cm.Name, metav1.GetOptions{})
 	if err != nil {
-		// Doesn't exist (or unreadable) — create it.
+		if !apierrors.IsNotFound(err) {
+			// A transient/permission error must not be mistaken for "absent" —
+			// otherwise we'd blindly Create and mask the real failure.
+			return fmt.Errorf("failed to get configmap %s: %w", cm.Name, err)
+		}
 		if _, err := client.CoreV1().ConfigMaps(namespace).Create(ctx, cm, metav1.CreateOptions{}); err != nil {
 			return fmt.Errorf("failed to create configmap %s: %w", cm.Name, err)
 		}
@@ -240,12 +246,17 @@ func createOrUpdateConfigMap(ctx context.Context, client kubernetes.Interface, c
 		return nil
 	}
 
-	if reflect.DeepEqual(existing.Data, cm.Data) {
-		// Already up to date — nothing to do.
+	// Reconcile both data and our managed labels; no-op when already in sync.
+	if reflect.DeepEqual(existing.Data, cm.Data) && labelsContain(existing.Labels, cm.Labels) {
 		return nil
 	}
-
 	existing.Data = cm.Data
+	if existing.Labels == nil {
+		existing.Labels = map[string]string{}
+	}
+	for k, v := range cm.Labels {
+		existing.Labels[k] = v
+	}
 	if _, err := client.CoreV1().ConfigMaps(namespace).Update(ctx, existing, metav1.UpdateOptions{}); err != nil {
 		return fmt.Errorf("failed to update configmap %s: %w", cm.Name, err)
 	}
@@ -256,7 +267,38 @@ func createOrUpdateConfigMap(ctx context.Context, client kubernetes.Interface, c
 	return nil
 }
 
-// createSecret creates a Kubernetes secret if it doesn't already exist
+// labelsContain reports whether have already includes every key/value in want.
+func labelsContain(have, want map[string]string) bool {
+	for k, v := range want {
+		if have[k] != v {
+			return false
+		}
+	}
+	return true
+}
+
+// createOrgCAConfigMap upserts the install-time argocd-org-ca ConfigMap holding
+// the operator's org CA, labeled as a foundational resource. It is the org-CA
+// analogue of createKeycloakSecrets — keeping the resource construction beside
+// the other foundational create helpers rather than inline in Install.
+func createOrgCAConfigMap(ctx context.Context, client kubernetes.Interface, namespace, orgCABundlePEM string) error {
+	return createOrUpdateConfigMap(ctx, client, &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      orgCAConfigMapName,
+			Namespace: namespace,
+			Labels: map[string]string{
+				"app.kubernetes.io/part-of":    NebariFoundationalPartOf,
+				"app.kubernetes.io/managed-by": "nebari-infrastructure-core",
+			},
+		},
+		Data: map[string]string{orgCAConfigMapKey: orgCABundlePEM},
+	})
+}
+
+// createSecret creates a Kubernetes secret if it doesn't already exist.
+// Create-only by design: these are generated one-time credentials that must
+// never be overwritten on re-deploy. For operator-supplied data that can rotate
+// (e.g. a CA bundle), use createOrUpdateConfigMap / createOrgCAConfigMap instead.
 func createSecret(ctx context.Context, client kubernetes.Interface, secret *corev1.Secret) error {
 	namespace := secret.Namespace
 	_, err := client.CoreV1().Secrets(namespace).Get(ctx, secret.Name, metav1.GetOptions{})
--- a/pkg/argocd/foundational.go
+++ b/pkg/argocd/foundational.go
@@ -3,6 +3,7 @@ package argocd
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
@@ -217,6 +218,41 @@ func createNamespace(ctx context.Context, client kubernetes.Interface, namespace
 		WithAction("created").
 		WithMetadata("namespace", namespace))
 
+	return nil
+}
+
+// createOrUpdateConfigMap creates the ConfigMap, or updates its data when it
+// already exists with different contents. Unlike createSecret (create-only, for
+// generated one-time credentials), an org CA bundle is operator-supplied and
+// rotates, so this upserts to make a changed trust_bundle actually propagate.
+func createOrUpdateConfigMap(ctx context.Context, client kubernetes.Interface, cm *corev1.ConfigMap) error {
+	namespace := cm.Namespace
+	existing, err := client.CoreV1().ConfigMaps(namespace).Get(ctx, cm.Name, metav1.GetOptions{})
+	if err != nil {
+		// Doesn't exist (or unreadable) — create it.
+		if _, err := client.CoreV1().ConfigMaps(namespace).Create(ctx, cm, metav1.CreateOptions{}); err != nil {
+			return fmt.Errorf("failed to create configmap %s: %w", cm.Name, err)
+		}
+		status.Send(ctx, status.NewUpdate(status.LevelInfo, fmt.Sprintf("Created ConfigMap %s", cm.Name)).
+			WithResource("configmap").
+			WithAction("created").
+			WithMetadata("configmap_name", cm.Name))
+		return nil
+	}
+
+	if reflect.DeepEqual(existing.Data, cm.Data) {
+		// Already up to date — nothing to do.
+		return nil
+	}
+
+	existing.Data = cm.Data
+	if _, err := client.CoreV1().ConfigMaps(namespace).Update(ctx, existing, metav1.UpdateOptions{}); err != nil {
+		return fmt.Errorf("failed to update configmap %s: %w", cm.Name, err)
+	}
+	status.Send(ctx, status.NewUpdate(status.LevelInfo, fmt.Sprintf("Updated ConfigMap %s", cm.Name)).
+		WithResource("configmap").
+		WithAction("updated").
+		WithMetadata("configmap_name", cm.Name))
 	return nil
 }
 
--- a/pkg/argocd/install.go
+++ b/pkg/argocd/install.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"strconv"
 	"time"
 
 	"go.opentelemetry.io/otel"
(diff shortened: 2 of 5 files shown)


pkg/argocd/install.go
```diff
@@ -21,13 +24,39 @@ import (
...
+	combinedCAPath       = "/etc/ssl/certs-combined/ca-bundle.crt"
+	systemCAPath         = "/etc/ssl/certs/ca-certificates.crt"
+	combineCAInitName    = "combine-ca-bundle"
+
+	// repoServerImageTpl resolves, via the chart's own tpl pass over
+	// repoServer.initContainers, to the same image the repo-server runs — so the
+	// init container's system CA bundle matches the runtime. This mirrors how the
+	// chart's built-in copyutil init container picks its image, and avoids pinning
+	// a tag that would drift from defaultChartVersion.
+	repoServerImageTpl = `{{ default .Values.global.image.repository .Values.repoServer.image.repository }}:{{ default (include "argo-cd.defaultTag" .) .Values.repoServer.image.tag }}`
```
dcmcand, line 59: Nice: this mirrors the chart's own copyutil image expression instead of pinning a tag, so the init image tracks `defaultChartVersion`. I confirmed the argo-cd chart renders `repoServer.initContainers` through `tpl`, so this resolves correctly at render time.

pkg/argocd/install.go
```diff
@@ -367,6 +409,113 @@ func addLocalGitopsMount(ctx context.Context, values map[string]any, localPath s
...
+	}
+	if err := createOrgCAConfigMap(ctx, client, namespace, orgCABundlePEM); err != nil {
+		return err
+	}
+	addOrgCAMount(ctx, values)
+	// Log a fingerprint, never the PEM body, so rotation is observable.
+	status.Send(ctx, status.NewUpdate(status.LevelInfo, "Configured repo-server to trust org CA bundle").
+		WithResource("argocd").
+		WithAction("ca-configured").
+		WithMetadata("ca_fingerprint", sha256Fingerprint(orgCABundlePEM)).
```
dcmcand: Good call logging a SHA-256 fingerprint rather than the PEM body. Rotation stays observable without leaking the bundle into logs or spans.

pkg/argocd/install.go
```diff
@@ -367,6 +409,113 @@ func addLocalGitopsMount(ctx context.Context, values map[string]any, localPath s
...
+		"volumeMounts": []any{
+			map[string]any{"name": orgCAVolumeName, "mountPath": orgCAMountDir, "readOnly": true},
+			map[string]any{"name": combinedCAVolumeName, "mountPath": combinedCADir},
+		},
+		"securityContext": map[string]any{
+			"runAsNonRoot":             true,
+			"allowPrivilegeEscalation": false,
+			"readOnlyRootFilesystem":   true,
+			"capabilities":             map[string]any{"drop": []any{"ALL"}},
+			"seccompProfile":           map[string]any{"type": "RuntimeDefault"},
```
dcmcand: Solid hardening on the init container: `runAsNonRoot`, `readOnlyRootFilesystem`, drop `ALL`, `seccompProfile: RuntimeDefault`, and it only writes the shared `emptyDir`.

pkg/argocd/install.go
```diff
@@ -367,6 +409,113 @@ func addLocalGitopsMount(ctx context.Context, values map[string]any, localPath s
...
+	repoServer["volumeMounts"] = appendToSlice(repoServer["volumeMounts"], map[string]any{
+		"name":      combinedCAVolumeName,
+		"mountPath": combinedCADir,
+		"readOnly":  true,
+	})
+
+	// Init container builds system-CA ∪ org-CA. The `cat` tolerates a missing
+	// system file (slim base image) rather than crash-looping the pod; the org CA
+	// is our own mounted ConfigMap, so its absence is a real error.
+	combineScript := fmt.Sprintf(
```
dcmcand, line 517: Stylistic, non-blocking: the positional-verb `Sprintf` (`%[1]s` etc.) is the one spot here that rewards a careful read. It is already commented and correct, so fine as-is; just flagging it.

pkg/argocd/foundational.go
```diff
@@ -214,7 +216,83 @@ func createNamespace(ctx context.Context, client kubernetes.Interface, namespace
...
+// labels when it already exists with different contents. Unlike createSecret
+// (create-only by design, for generated one-time credentials), an org CA bundle
+// is operator-supplied and rotates, so this upserts to make a changed
+// trust_bundle actually propagate.
+func createOrUpdateConfigMap(ctx context.Context, client kubernetes.Interface, cm *corev1.ConfigMap) error {
+	namespace := cm.Namespace
+	existing, err := client.CoreV1().ConfigMaps(namespace).Get(ctx, cm.Name, metav1.GetOptions{})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			// A transient/permission error must not be mistaken for "absent" —
```
dcmcand, line 238: The right instinct: a non-`NotFound` Get error is surfaced rather than mistaken for "absent" and blindly Created. Avoids masking a transient or RBAC failure as a create.

pkg/argocd/orgca_test.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+	})
+
+	t.Run("no-op when data unchanged issues no update", func(t *testing.T) {
+		existing := newCM("CA-A")
+		client := fake.NewSimpleClientset(existing) //nolint:staticcheck // SA1019: NewSimpleClientset is deprecated but still functional for tests
+		if err := createOrUpdateConfigMap(context.Background(), client, newCM("CA-A")); err != nil {
+			t.Fatalf("createOrUpdateConfigMap: %v", err)
+		}
+		// Prove the DeepEqual early-return fired: a write would defeat the no-op.
```
dcmcand, line 177: Good test: asserting that no create/update action fires actually proves the `DeepEqual` early-return, instead of just trusting it happened.

pkg/nic/deploy.go
```diff
@@ -185,7 +185,7 @@ func (c *Client) Deploy(ctx context.Context, cfg *config.NebariConfig, opts Depl
 		// Build ArgoCD config with Keycloak OIDC SSO
 		argoCDConfig := argocd.ConfigWithOIDC(cfg.Domain, infraSettings.KeycloakBasePath, argoCDClientSecret)
 
-		if err := argocd.Install(ctx, cfg, clusterProvider, gitConfig, argoCDConfig); err != nil {
+		if err := argocd.Install(ctx, cfg, clusterProvider, gitConfig, trustPEM, argoCDConfig); err != nil {
```
dcmcand, line 197: Question (non-blocking, and pre-existing): an `Install` error here is logged as a warning and the deploy continues. With `trust_bundle` set, a `configureRepoServerCATrust` failure means the repo-server cannot pull through the proxy and nothing syncs, yet the deploy still reports completion. Is warn-and-continue still the right policy when trust is required, or should a `trust_bundle` run treat this as fatal? Fine to defer, since the warn-and-continue predates this PR.

dcmcand (dismissed):
Approving. Clean, well-reasoned implementation and the tests are genuinely thorough.

The divergence from #310's proposed approach (install-time ConfigMap instead of the projected one; combined-bundle env vars instead of `argocd-tls-certs-cm`) is the right call and well-documented: the repo-server is upstream of trust-manager in the bootstrap, and `--ca-file` would replace rather than augment the system trust pool. The single `configureRepoServerCATrust` seam keeps `Install` from accreting per-feature blocks, and the upsert-vs-create-only distinction (a CA rotates, generated secrets do not) is handled and documented on both helpers.

Two non-blocking questions, also left inline:
1. `deploy.go` soft-fails an `Install` error (pre-existing). Worth deciding whether a `trust_bundle` deploy should treat that as fatal, since nothing syncs without the repo-server trusting the proxy.
2. The documented `shouldSkipUpgrade` limitation (values-only changes are not applied to existing clusters at the pinned chart version): is there a tracking issue for the upgrade-engine change? Existing-cluster operators are otherwise silently stuck until `defaultChartVersion` bumps.

One thing outstanding by design: #310's DoD of a real proxied `git clone` is the e2e step you flagged for a TLS-inspected cluster. I confirmed the argo-cd chart renders `repoServer.initContainers` through `tpl` (so the templated init image resolves), but against the chart's `main` branch, not the exact pinned `defaultChartVersion`. A `helm template` render assertion against the pinned version would cheaply guard init-image resolution without needing a cluster.

Issue #310: the values-rendering DoD is met with strong coverage; the proxied-clone DoD is deferred to the documented e2e step.

viniciusdc:
Thanks for the thorough review @dcmcand. Dispositions on the three points:

1. **Soft-fail policy (`deploy.go`)** — this already has a home in #302 (deploy partial-failure semantics), so I don't want to change the warn-and-continue piecemeal here and pre-empt that decision. I've added the `trust_bundle` angle to #302: with a bundle set, an `Install` failure means nothing syncs, which argues for fatal-on-required-step. Deferring the actual change to #302.

2. **`shouldSkipUpgrade` values-awareness** — opened #364 and linked it in the Known-limitation section. It's the values-aware skip (skip only when version *and* rendered values match), with the normalization caveat so we don't re-introduce the per-deploy churn #85 removed. Distinct from #138, which asked for the failed/pending status check that's already in place.

3. **e2e / `tpl` init image** — agreed; a `helm template` render assertion against the pinned `defaultChartVersion` is the cheap guard for init-image resolution, and I'll bundle it with the proxied-clone e2e step on a TLS-inspected cluster.

Holding the merge anyway until #346 lands and this retargets to `main`.

viniciusdc:
Self-review pass before merge. One real gap found and fixed in this branch (d3776b7):

**Rotation was silently ineffective for a running repo-server.** The combined bundle is built by the init container into an emptyDir, so it only gets rebuilt on pod restart. `createOrUpdateConfigMap` correctly upserts `argocd-org-ca` on a rotated `trust_bundle`, but nothing rolled the pod, so it kept serving the stale combined bundle. The sha256 fingerprint we already compute for the status log is now also stamped on the pod template as `repoServer.podAnnotations["nebari.dev/org-ca-checksum"]` inside `addOrgCAMount` (merged with any existing podAnnotations, not clobbered). A rotated bundle changes the annotation, which changes the pod template, which rolls the pod and re-runs the combine init container. Existing `TestAddOrgCAMount*` updated for the new signature plus a merge assertion and an end-to-end checksum check.

Residual caveat, same as the PR body: while #364 stands, an upgrade against an already-Deployed release at the same chart version is skipped entirely, so the annotation only takes effect once the values actually apply (fresh install, chart bump, or manual `rollout restart`). It makes rotation correct whenever an upgrade lands rather than a no-op forever; it does not bypass the skip logic.

Still to add as a follow-up: a client-side `helm template` render assertion that the init container's `tpl` image resolves on the pinned chart (9.4.1), not just `main`. A silent tpl non-resolution would ship a literal `{{ ... }}` image string and the repo-server would never schedule; this turns the "helm template assertion" bullet from the test plan into a real test. Will push it on this branch shortly.

viniciusdc:
Adding some detail on how this actually works, since the CA plumbing is easy to lose track of.

## The problem, in plain terms

The repo-server's job is to clone git repos and pull Helm charts over HTTPS. In a corporate network every outbound connection gets intercepted by a TLS-inspecting proxy, which hands back each site's certificate re-signed by the *company's own* CA. The repo-server has never heard of that CA, so it rejects everything with `x509: certificate signed by unknown authority`, and the cluster never finishes bootstrapping because it can't pull a single chart.

The exchange looks roughly like this:

```mermaid
sequenceDiagram
    participant R as ArgoCD repo-server
    participant P as TLS-inspecting proxy
    participant G as github.com / chart repos
    R->>P: I need to reach github.com
    P->>G: opens its own TLS session upstream
    P-->>R: here's github.com, cert signed by "Org CA"
    Note over R: before this PR, trust = public CAs only
    R--xP: nope, unknown authority (x509)
    Note over R: after this PR, trust = system CAs + Org CA
    R->>P: Org CA is on my list, carry on
    P-->>R: clone / pull succeeds
```

## The fix

Give the repo-server a trust list that is *system CAs plus the org CA*, without throwing the public ones away:

```mermaid
flowchart LR
    A["nic writes the org CA into<br/>the argocd-org-ca ConfigMap"] --> B
    subgraph pod["repo-server pod startup"]
      B["init container<br/>combine-ca-bundle"] -->|"cat system bundle,<br/>then append org CA"| C["combined bundle<br/>(emptyDir)"]
    end
    C --> D["repo-server reads it via<br/>SSL_CERT_FILE / GIT_SSL_CAINFO / CURL_CA_BUNDLE"]
    D --> E["git clone + helm pull<br/>through the proxy, Synced"]
```

Three things worth calling out on why it is shaped this way:

- The org CA lands in a ConfigMap that *nic itself* writes at install time, not the trust-manager Bundle. That is deliberate. The repo-server is the thing that has to pull trust-manager's own image, so it needs to trust the proxy *before* trust-manager exists. Chicken and egg, solved by not depending on it.
- The union happens in an init container running the repo-server's own image, so it picks up whatever system bundle that image ships (the cert path differs across Debian/RHEL/Alpine). The combined file lands in an emptyDir the main container mounts read-only.
- All three of `SSL_CERT_FILE`, `GIT_SSL_CAINFO`, `CURL_CA_BUNDLE` get set, because the consumers are different: the repo-server's own Go client, the `git` CLI subprocess, and the `helm` subprocess. Setting only one leaves the others on the default store.

## Rotation (d3776b7)

The init container only re-runs when the pod restarts, so rotating the org CA would leave a running repo-server on stale trust until someone bounced it by hand. Added a `nebari.dev/org-ca-checksum` pod annotation set to the CA's fingerprint, so changing the bundle changes the annotation, which rolls the pod, which re-runs the init container. Rotation lands on the next apply with no manual step.

## Proof it holds end to end

On k3d I put mitmproxy in front as the egress proxy and gave it our org CA as its signing CA, so it re-signs github.com with the org CA exactly like a corporate proxy would. With the combined bundle the repo-server's `git ls-remote` succeeds. Point `GIT_SSL_CAINFO` at the system-only bundle instead and it fails with `certificate signer not trusted`. So the clone works *only* because the injected org CA is there, which is the whole point.

One honest follow-up: I wanted a tpl-render unit test asserting the combined bundle is system + org, but the way I sketched it it's network-dependent. Leaving it as a follow-up rather than pretending it's covered.

pkg/argocd/foundational.go
```diff
@@ -220,7 +225,83 @@ func createNamespace(ctx context.Context, client kubernetes.Interface, namespace
 	return nil
 }
 
-// createSecret creates a Kubernetes secret if it doesn't already exist
+// createOrUpdateConfigMap creates the ConfigMap, or reconciles its data and
+// labels when it already exists with different contents. Unlike createSecret
+// (create-only by design, for generated one-time credentials), an org CA bundle
+// is operator-supplied and rotates, so this upserts to make a changed
+// trust_bundle actually propagate.
+func createOrUpdateConfigMap(ctx context.Context, client kubernetes.Interface, cm *corev1.ConfigMap) error {
```
tylerpotts, line 233: This helper gets the details right: a non-NotFound Get error surfaces instead of being masked as "absent", labels are merged rather than clobbered on update, and the no-op path is proven in the test by asserting no write action was issued against the fake clientset. That is the strongest form of that test.

pkg/argocd/install.go
```diff
@@ -373,21 +428,140 @@ func addLocalGitopsMount(ctx context.Context, values map[string]any, localPath s
...
+	repoServer["volumeMounts"] = appendToSlice(repoServer["volumeMounts"], map[string]any{
+		"name":       combinedCAVolumeName,
+		keyMountPath: combinedCADir,
+		"readOnly":   true,
+	})
+
+	// Init container builds system-CA ∪ org-CA. The `cat` tolerates a missing
+	// system file (slim base image) rather than crash-looping the pod; the org CA
+	// is our own mounted ConfigMap, so its absence is a real error.
+	combineScript := fmt.Sprintf(
```
tylerpotts, line 517: Nice that the script is built entirely from package constants, so no user input ever reaches the shell line, and the PEM only travels as ConfigMap data. The hardened init `securityContext` below is also correct as-is (`runAsNonRoot` works because the argocd image declares numeric user 999).

pkg/argocd/install.go
```diff
@@ -373,21 +428,140 @@ func addLocalGitopsMount(ctx context.Context, values map[string]any, localPath s
...
+	if err := createOrgCAConfigMap(ctx, client, namespace, orgCABundlePEM); err != nil {
+		return err
+	}
+	caFingerprint := sha256Fingerprint(orgCABundlePEM)
+	addOrgCAMount(ctx, values, caFingerprint)
+	// Log a fingerprint, never the PEM body, so rotation is observable.
+	status.Send(ctx, status.NewUpdate(status.LevelInfo, "Configured repo-server to trust org CA bundle").
+		WithResource("argocd").
+		WithAction("ca-configured").
+		WithMetadata("ca_fingerprint", caFingerprint).
```
tylerpotts, line 465: **Suggestion (non-blocking):** on an existing cluster this success message is followed by `shouldSkipUpgrade` dropping the values that were just wired, so the operator sees "Configured repo-server to trust org CA bundle" then "Argo CD already up to date" while the security feature they enabled silently no-ops. The fix rightly belongs in https://github.com/nebari-dev/nebari-infrastructure-core/issues/364, but *detection* is cheap now: in `InstallHelm`, the skipped release's applied values are already in hand (helm.go:96), so the old `nebari.dev/org-ca-checksum` annotation can be compared against the new fingerprint and a `status.LevelWarning` emitted when they differ. Fine to fold into #364 instead if you prefer.

pkg/argocd/install.go
```diff
@@ -373,21 +428,140 @@ func addLocalGitopsMount(ctx context.Context, values map[string]any, localPath s
...
+			"runAsNonRoot":             true,
+			"allowPrivilegeEscalation": false,
+			"readOnlyRootFilesystem":   true,
+			"capabilities":             map[string]any{"drop": []any{"ALL"}},
+			"seccompProfile":           map[string]any{keyType: "RuntimeDefault"},
+		},
+	})
+
+	// Point the repo-server's TLS clients at the combined bundle. git uses
+	// GIT_SSL_CAINFO, curl/helm use CURL_CA_BUNDLE, Go uses SSL_CERT_FILE.
```
tylerpotts, line 538: Comment nit: helm is Go, so it honors `SSL_CERT_FILE` and does not read `CURL_CA_BUNDLE` (that one covers curl proper). The extra env var is harmless belt-and-braces; just adjust the wording so future readers do not conclude helm needs it.

pkg/argocd/secrets.go
```diff
@@ -41,9 +41,9 @@ func ConfigureGitRepoAccess(ctx context.Context, client kubernetes.Interface, gi
 
 	// Create repository secret data
 	secretData := map[string]string{
-		"name": "gitops-repo",
-		"type": gitRepoType,
-		"url":  gitConfig.URL,
+		"name":  "gitops-repo",
+		keyType: gitRepoType,
```
tylerpotts, line 45: Stylistic: the goconst-appeasement constants (`keyType`, `keyMountPath`) leaking into this unrelated file trade self-evident map-key literals for indirection. A `//nolint:goconst` at the offending sites or a linter config tweak would read better, and avoids setting a precedent of shared constants for generic map keys.

pkg/argocd/orgca_test.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	// Existing podAnnotations are preserved and the checksum is merged in.
+	pa := repoServer["podAnnotations"].(map[string]any)
+	if pa["existing/annotation"] != "keep" || pa["nebari.dev/org-ca-checksum"] != "fp" {
+		t.Errorf("podAnnotations not merged: %v", pa)
+	}
+}
+
+func TestCreateOrUpdateConfigMap(t *testing.T) {
```
tylerpotts, line 136: Minor test gaps, both two-line table additions: the label-drift branch (data equal, managed labels missing, so update fires via `labelsContain`) and the non-NotFound Get error path are untested.

tylerpotts (approved):
Solid work. The two headline divergences from https://github.com/nebari-dev/nebari-infrastructure-core/issues/310 hold up on their merits: the install-time ConfigMap is a genuine bootstrap-ordering requirement (the repo-server is what pulls trust-manager's chart, so it cannot consume trust-manager's projected bundle), and leaving `argocd-tls-certs-cm` empty is well supported by the field reports on the issue showing the per-hostname `--ca-file` approach is counter-productive.

**Independent verification done as part of this review:**
- `go test ./pkg/argocd/... ./pkg/nic/...` passes; `golangci-lint run` is clean.
- Rendered argo-cd **9.4.1** (`helm template`) with the exact values `addOrgCAMount` produces: the chart's `tpl` pass resolves the init image template to a concrete `quay.io/argoproj/argocd:v3.3.0`, and the checksum annotation, all three env vars, both volumes, and the mounts land correctly in the repo-server Deployment. That closes the "does the chart honor these values" half of the test-plan gap; the remaining e2e item (git actually trusting the file behind a real TLS-inspecting proxy) still needs a cluster.

**Design notes (non-blocking):**
- The upsert-vs-create-only split (`createOrUpdateConfigMap` vs `createSecret`) aligns well with the repo's reconciliation philosophy, and documenting the rationale at both seams is appreciated. Future foundational resources should follow this vocabulary.
- `configureRepoServerCATrust` as a single no-op-when-unset seam is a good template for future per-feature wiring, better than the inline block pattern it sits beside.
- The known https://github.com/nebari-dev/nebari-infrastructure-core/issues/364 limitation is honestly flagged. One suggestion inline about surfacing detection of the skip-path case now, even though the fix itself rightly belongs in that issue.
- On existing clusters the feature half-applies (ConfigMap upserted, values dropped by `shouldSkipUpgrade`), which is documented and tracked, and fresh installs, the primary path for a new feature, work correctly. Not gating.
- `orgCABundlePEM` as a positional param is consistent with the `WriteAllToGit(..., trustBundlePEM)` precedent, but `Install` is at six params now; the next resolved input should consolidate into `argocd.Config` or an options struct.
- `repoServerImageTpl` couples NIC to a chart-internal helper (`argo-cd.defaultTag`), bounded by the version pin. The render assertion suggested in the PR body should be treated as required follow-up before the next `defaultChartVersion` bump, since a bump is exactly the event that will finally apply these values.

PR #356, opened 2026-06-04
fix(longhorn): tolerate storage-node taints for system-managed components

dcmcand:
## What

When `dedicated_nodes: true`, NIC now sets `defaultSettings.taintToleration` and `defaultSettings.systemManagedComponentsNodeSelector` so Longhorn's system-managed components (instance-manager, engine-image, CSI plugin) can schedule onto tainted dedicated storage nodes. The embedded iSCSI prerequisite DaemonSet now tolerates all taints. The change lives in the shared `pkg/storage/longhorn` package, so every provider that installs Longhorn with dedicated nodes benefits.

Also adds a production-recommended dedicated Longhorn config (tainted storage node group, one node per AZ, small instance with a large gp3 disk, `longhorn.dedicated_nodes: true`) to `examples/aws-config.yaml`.

## Why

Previously only `longhornManager`/`longhornDriver` received the storage-node toleration. The system-managed components that serve replicas tolerate taints only via the `taint-toleration` setting, so tainting the dedicated node group left Longhorn unable to run storage engines there.

## Testing

- New table-driven tests in `pkg/storage/longhorn` for the two `defaultSettings` keys, the `formatNodeSelector` helper, and the iSCSI DaemonSet toleration.
- `go test ./...`, `go vet ./...`, `gofmt -l`, `go build ./...` all pass; `nic validate -f examples/aws-config.yaml` passes.

## Caveat

`taintToleration` is pinned to the default taint `node.longhorn.io/storage=true:NoSchedule`. Overriding `node_selector` together with a non-default taint value won't line up; that can be a follow-up if anyone needs it.

Closes #355

## Test coverage

#355 criterion 1 (Longhorn functional on tainted dedicated nodes) is covered by unit tests and `nic validate`, not an end-to-end integration test against a live cluster.

### DIFF
--- a/docs/adr/0002-longhorn-distributed-block-storage-for-aws.md
+++ b/docs/adr/0002-longhorn-distributed-block-storage-for-aws.md
@@ -172,32 +172,47 @@ When Longhorn is disabled (`longhorn.enabled: false`), falls back to `gp2`.
 
 Default (no longhorn section needed):
 ```yaml
-amazon_web_services:
-  region: us-west-2
-  kubernetes_version: "1.34"
-  node_groups:
-    user:
-      instance: m7i.xlarge
-      min_nodes: 0
-      max_nodes: 5
+cluster:
+  aws:
+    region: us-west-2
+    kubernetes_version: "1.34"
+    node_groups:
+      user:
+        instance: m7i.xlarge
+        min_nodes: 0
+        max_nodes: 5
 ```
 
-Explicit configuration:
+Explicit configuration with dedicated, tainted storage nodes:
 ```yaml
-amazon_web_services:
-  region: us-west-2
-  longhorn:
-    replica_count: 3
-    dedicated_nodes: true
-    node_selector:
-      node.longhorn.io/storage: "true"
+cluster:
+  aws:
+    region: us-west-2
+    node_groups:
+      storage:
+        instance: m7g.large
+        min_nodes: 2
+        max_nodes: 2
+        disk_size: 500
+        labels:
+          node.longhorn.io/storage: "true"
+        taints:
+          - key: node.longhorn.io/storage
+            value: "true"
+            effect: NO_SCHEDULE
+    longhorn:
+      replica_count: 3
+      dedicated_nodes: true
+      node_selector:
+        node.longhorn.io/storage: "true"
 ```
 
 Opt out:
 ```yaml
-amazon_web_services:
-  longhorn:
-    enabled: false
+cluster:
+  aws:
+    longhorn:
+      enabled: false
 ```
 
 ### Helm Values
@@ -218,6 +233,11 @@ Dedicated nodes deployment adds:
 ```yaml
 defaultSettings:
   createDefaultDiskLabeledNodes: true
+  # System-managed components (instance-manager, engine-image, CSI plugin) are
+  # not covered by the per-component tolerations below; they honor node taints
+  # and pinning only through these two settings.
+  taintToleration: node.longhorn.io/storage=true:NoSchedule
+  systemManagedComponentsNodeSelector: node.longhorn.io/storage:true
 
 longhornManager:
   nodeSelector:
--- a/examples/aws-config.yaml
+++ b/examples/aws-config.yaml
@@ -44,6 +44,26 @@ cluster:
         min_nodes: 1 # minimum of 1 node required
         max_nodes: 5
 
+      # Dedicated Longhorn storage nodes (production-recommended).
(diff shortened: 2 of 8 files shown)


viniciusdc (dismissed):
LGTM

viniciusdc:
Ran a full maintainability pass plus an architecture review over this. Verdict up front: architecturally sound, no blockers. The fix sits in the right place (`pkg/storage/longhorn` stays provider-agnostic, only Longhorn-domain labels, no cloud leakage), and the cross-cutting conventions all hold (OTel spans, `%w` wrapping, `status.Send`). Tests are genuinely thorough. A few things worth your call before merge.

**The hardcoded toleration (the caveat you already flagged).** This is the one item that matters. `taintToleration` is pinned to `node.longhorn.io/storage=true:NoSchedule` at `install.go:33`, but `systemManagedComponentsNodeSelector` follows the user's `node_selector`. So a user who customizes the selector but not the taint gets system-managed components stuck `Pending` with no obvious signal, which is the #355 failure mode coming back on the custom path. IMO it's fine to defer parameterizing the taint since that's a bigger change, as long as two things land here: (1) document the constraint right at the `NodeSelector` field in `config.go`, and (2) open a tracking follow-up. With those, I'm happy to leave it.

**Self-referential PR number in the example.** `examples/aws-config.yaml:96` references `#356` as a requirement, which reads oddly once this merges. Reword to a released NIC version or drop it.

**Dead validator nit.** The example's `effect: NO_SCHEDULE` is correct (it's the EKS API enum, EKS maps it to the k8s `NoSchedule` taint the toleration matches). But `Provider.Validate` in `pkg/provider/aws/provider.go:215-220` checks against k8s-style `NoSchedule`/etc, which is inconsistent. It rejects nothing today because `Provider.Validate` has no production caller, but if anyone wires it in it'll reject valid configs. Worth a cleanup pass, not a blocker.

**ADR-0002 doc drift (pre-existing).** The config examples there still use the legacy top-level `amazon_web_services:` key instead of the `cluster.aws.longhorn` nesting this PR's example uses, and the "Helm Values → Dedicated nodes" block doesn't list the two new `defaultSettings` keys. Not introduced here, but the dedicated-nodes work is a good moment to fix it.

One more: criterion #1 from #355 (Longhorn actually functional on tainted dedicated nodes) is unit-tested + `nic validate` only, no integration test. Worth saying that plainly in the PR description so it's clear what's covered.

dcmcand:
Addressed in 868efcb plus a follow-up issue.

- **Hardcoded toleration:** documented the constraint directly on `longhorn.Config.NodeSelector`, and opened #363 to parameterize the taint. Left it pinned for now, per your call.
- **Self-referential PR number:** dropped from the example; the comment now describes the required taint directly.
- **Dead validator:** `Provider.Validate` now checks taint effects against the EKS enum (`NO_SCHEDULE`/`NO_EXECUTE`/`PREFER_NO_SCHEDULE`), with a regression test. As you noted, it would have rejected this PR's own example if it were ever wired into a caller.
- **ADR-0002:** updated to the `cluster.aws` nesting and added the two new `defaultSettings` keys (`taintToleration`, `systemManagedComponentsNodeSelector`) to the dedicated-nodes Helm values block.
- **Test coverage:** called out in the PR description that #355 criterion 1 is covered by unit tests and `nic validate`, not an end-to-end integration test.

dcmcand:
@viniciusdc all your review points are addressed, latest is 5176b22:

- Documented the pinned-taint constraint on `longhorn.Config.NodeSelector` and opened #363 to parameterize it (left it pinned for now per your call).
- Dropped the self-referential PR number from the example.
- Fixed the taint-effect validator to the EKS enum (`NO_SCHEDULE`/`NO_EXECUTE`/`PREFER_NO_SCHEDULE`); it now lives in a `validateTaints` helper with direct test coverage.
- Updated ADR-0002 to the `cluster.aws` nesting plus the two new `defaultSettings` keys.
- Added a test-coverage note to the PR description (unit + `nic validate`, no integration test).

Ready for another look when you have a moment.

viniciusdc:
Went through all three follow-up commits, this covers everything. The `NodeSelector` doc constraint + #363 is exactly what I had in mind for the pinned-taint caveat. Nice that you fixed the empty `node_selector: {}` case outright (`len() > 0` + the fallback test) rather than just documenting it. The `validateTaints` extraction to the EKS enum is the right call, and dropping the IMDS timeout from the valid-case tests is a good bonus.

LGTM, thanks @dcmcand.

viniciusdc (approved):

PR #361, opened 2026-06-05
docs(adr): ADR-0006 — conditional foundational software via provider-driven Helm

viniciusdc:
Splitting this out of #348 so the decision gets its own discussion thread and footprint, per our convention that ADRs land on their own PRs.

## What this records

We decided in an internal sync that **provider/cluster-conditional foundational software installs via provider-driven Helm in `Deploy()`/`Destroy()`, not as ArgoCD apps.** Unconditional foundational software (cert-manager, Keycloak, Envoy Gateway, PostgreSQL, the Nebari operator) stays in GitOps. ADR-0006 writes that down and refines the blanket "GitOps for software" rule in ADR-0001 / AGENTS.md #5 for the conditional case.

The motivation: the install/remove decision and the values for conditional software depend on provider + live-cluster state the provider already computes (GPU nodes present, existing-vs-cloud, region, VPC ID, node-group-derived toggles). Gating that through the GitOps render layer leaks provider knowledge into `pkg/argocd` and `InfraSettings` and doesn't scale; several components also need ordered teardown relative to `tofu destroy` (Longhorn's CSI finalizers). The Helm logic is already hand-rolled and duplicated across `cluster_autoscaler.go`, `aws_load_balancer_controller.go`, and `gpu_operator.go`.

## Target design (implementation tracked in #349)

A shared `helmInstaller` (kubeconfig + chart + values + lifecycle) plus a per-provider `ConditionalCharts` lister, with a reconcile loop that installs the listed set and uninstalls managed releases no longer listed. The GPU operator (#348) and the cluster-autoscaler (#352) are the first hand-rolled instances to fold in; MetalLB migrates off the writer-skip mechanism.

This PR is the decision + design only (one markdown file + the ADR index row). The interface implementation is #349. Discussion welcome here.


dcmcand (dismissed):
Looks Good

dcmcand (approved):

PR #367, opened 2026-06-05
fix(longhorn): make dedicated_nodes serve volumes cluster-wide (#366, #369)

viniciusdc:
resolves #354 

## What

Refactors NIC's Longhorn `dedicated_nodes` to the intended topology: **replica data only on dedicated storage nodes, but the volume-serving system pods on every node.** Fully fixes #369 and #366. Supersedes this branch's original cosmetic-warning fix for #365 — that warning is moot now (the nodeSelector it concerned is removed, not corrected).

## The bugs (live-cluster confirmed)

- **#366 (P1):** NIC pinned `longhorn-csi-plugin` + `longhorn-manager` + instance-managers to storage nodes → pods on `user`/`general`/`gpu` nodes failed to mount (`CSINode … does not contain driver driver.longhorn.io`).
- **#369 (P1):** storage nodes never got `node.longhorn.io/create-default-disk=true` → zero disks → every volume `detached/faulted` with `ReplicaSchedulingFailure`.

## Changes

- **`pkg/storage/longhorn/install.go`** — `buildHelmValues` for `dedicated_nodes` no longer sets `systemManagedComponentsNodeSelector` or a `nodeSelector` on `longhornManager`/`longhornDriver`; their tolerations become tolerate-all (`{operator: Exists}`), same rationale as the iSCSI prerequisite DaemonSet. `taintToleration` is set to `":"` so the system-managed components (csi-plugin, instance-managers, engine-image, share-manager) tolerate every taint and run on tainted workload pools too; `createDefaultDiskLabeledNodes` stays. Replica confinement now comes from disks existing only on storage nodes. Removed the dead `formatNodeSelector`.
- **`pkg/storage/longhorn/storage_nodes.go`** (new) — `StorageSelector(cfg)` / `StorageNodeLabels(cfg)` own the storage-node label policy in the shared package, and `warnIfMissingStorageDiskLabel` warns (post install/upgrade) when `dedicated_nodes` is set but no node carries the disk label — turning the silent non-AWS fault into an actionable signal.
- **`pkg/provider/aws/tofu.go`** — injects `node.longhorn.io/create-default-disk=true` onto node groups matching the storage selector (via the shared helpers), applied by kubelet at node registration (autoscaler-safe). `pkg/storage/longhorn` stays provider-agnostic.
- **`pkg/storage/longhorn/config.go`** — export `NodeStorageLabel` + `CreateDefaultDiskLabel`, document the disk-label contract + exact-value-match semantics; `NodeSelector` now identifies the storage pool (no longer pins components).
- **`examples/aws-config.yaml`** — show the `create-default-disk` label; fix the stale "pins Longhorn" comments.
- **`docs/adr/0002`** — update the dedicated-nodes Helm-values block + a 2026-06 note recording the topology change and its limitations.
- Tests: full dedicated-nodes contract assertion, shared-helper tables, the disk-label warn guard, and AWS edge cases (multiple storage groups, idempotency, no-mutation, literal wire value).

## Scope of the issue fixes

**`Closes #369` — fully (AWS).** The disk label lands on storage nodes via the AWS injection, autoscaler-safe, tested. Non-AWS providers must label their storage pool themselves; the new guard warns when they haven't. Follow-up tracked for cross-provider parity.

**`Closes #366` — fully.** Mounts work on every workload node, untainted or tainted. The system-managed `longhorn-csi-plugin` tolerates every taint via `taintToleration: ":"` (merged from #372), so a `nvidia.com/gpu:NoSchedule` pool runs csi-plugin and PVC pods mount there. Replica data stays on the storage nodes because only they carry a Longhorn disk.

One #366 DoD item is intentionally not implemented as written: the controller that reconciles `allowScheduling:false` on non-storage Node CRs. Disk-confinement replaces it (non-storage nodes have no disk, so replicas cannot land there), which is simpler and autoscaler-safe. The trade-off is that confinement is an emergent property (no disk) rather than an explicit guard.

## Note on the tolerate-all change

The tolerate-all settings (`{operator: Exists}` on manager/driver, and `taintToleration: ":"` for the system-managed components) put privileged Longhorn pods on control-plane, GPU, and any tainted nodes. Intended for managed control planes (EKS) and consistent with the iSCSI DaemonSet precedent; on bring-your-own clusters with reachable masters it lands storage infra there.

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...` (incl. `-race`), `gofmt -l` all clean.
- Verified end-to-end on a live EKS cluster (storage x2 tainted+labeled, an untainted general pool, and a tainted non-storage pool): csi-plugin and instance-manager run on every node, disks exist only on the storage nodes, all replicas are confined to the storage nodes, and PVC pods mount on both an untainted and a tainted node with data readable. Volumes that had faulted before the disk fix self-healed once the disks appeared.

Closes #369
Closes #366

### DIFF
--- a/docs/adr/0002-longhorn-distributed-block-storage-for-aws.md
+++ b/docs/adr/0002-longhorn-distributed-block-storage-for-aws.md
@@ -322,3 +322,7 @@ Known limitations (tracked separately): csi-plugin tolerations come from the
 need their taint added there before mounts work on them (relates to #363/#368);
 and the disk-label injection is AWS-only — Hetzner/existing must label their
 storage pool themselves (the installer now warns when no node carries the label).
+
+Switching an existing cluster between `dedicated_nodes` modes is a manual
+replica migration (Longhorn does not move existing replicas on a flag change).
+See [Migrating Longhorn dedicated_nodes modes](../operations/longhorn-dedicated-nodes-migration.md).
--- a/docs/adr/0002-longhorn-distributed-block-storage-for-aws.md
+++ b/docs/adr/0002-longhorn-distributed-block-storage-for-aws.md
@@ -229,32 +229,34 @@ defaultSettings:
   replicaAutoBalance: best-effort
 ```
 
-Dedicated nodes deployment adds:
+Dedicated nodes deployment adds (current behavior — see "Update (2026-06)" below
+for the topology change away from component pinning):
 ```yaml
 defaultSettings:
   createDefaultDiskLabeledNodes: true
-  # System-managed components (instance-manager, engine-image, CSI plugin) are
-  # not covered by the per-component tolerations below; they honor node taints
-  # and pinning only through these two settings.
+  # Lets the replica-role system-managed components (instance-manager,
+  # engine-image, share-manager) tolerate the storage-node taint so they can run
+  # on the dedicated storage nodes. NOTE: the csi-plugin DaemonSet is also
+  # system-managed and inherits ONLY this toleration, so a tainted workload pool
+  # (e.g. GPU) would need its taint added here for mounts to work there.
   taintToleration: node.longhorn.io/storage=true:NoSchedule
-  systemManagedComponentsNodeSelector: node.longhorn.io/storage:true
 
+# longhorn-manager (DaemonSet) and the driver run on EVERY node so volumes can be
+# served to workloads anywhere — tolerate-all, NO nodeSelector. Replica DATA is
+# confined to storage nodes by createDefaultDiskLabeledNodes (only labeled
+# storage nodes get a disk), not by pinning these components.
 longhornManager:
-  nodeSelector:
-    node.longhorn.io/storage: "true"
   tolerations:
-    - key: node.longhorn.io/storage
-      operator: Exists
-      effect: NoSchedule
-
+    - operator: Exists
 longhornDriver:
-  nodeSelector:
-    node.longhorn.io/storage: "true"
   tolerations:
-    - key: node.longhorn.io/storage
-      operator: Exists
-      effect: NoSchedule
+    - operator: Exists
 ```
+
+The storage node group must carry `node.longhorn.io/create-default-disk=true`
+(the AWS provider injects it automatically; other providers must label their
+storage pool). Without it, `createDefaultDiskLabeledNodes` provisions no disks
+and every volume faults with `ReplicaSchedulingFailure`.
 
 ### Testing Strategy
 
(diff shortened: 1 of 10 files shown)


oren-openteams:
## Review: cosmetic fix is correct, but it locks in the architecture #366 says is wrong

Verified empirically on a live cluster with `dedicated_nodes: true`: the warning IS cosmetic. The live `longhorn-manager` DaemonSet had `nodeSelector: {"node.longhorn.io/storage":"true"}` applied despite the warning — Helm's "destination value kept" semantics worked as you described in the "Note on the scheduling impact" section. So the answer to the open question in that note is: **yes, the pods were genuinely pinned, the warning was cosmetic.**

```bash
kubectl -n longhorn-system get daemonset longhorn-manager \
    -o jsonpath='{.spec.template.spec.nodeSelector}{"\n"}'
# {"node.longhorn.io/storage":"true"}    ← pinned, despite the warning
```

So as a fix for #365 in isolation, this PR is correct: removes the warning, makes the type explicit, adds a regression test. 👍

**Architectural concern worth flagging:** this PR makes `longhornManager.nodeSelector` and `longhornDriver.nodeSelector` apply *more reliably*, but #366 argues those nodeSelectors shouldn't be set at all. On a live cluster with the current `install.go`, `longhorn-manager` pinned to storage nodes means `instance-manager` pods (which it spawns) also only land on storage nodes — and pods on workload/GPU/user nodes then fail to attach Longhorn volumes with:

```
rpc error: code = Aborted desc = volume pvc-... is not ready for workloads
```

The architecturally correct fix (per #366) is to drop the nodeSelector on these deployer-managed pods entirely and control replica placement via `allowScheduling: false` on Longhorn `Node` CRs for non-storage nodes. That's the opposite direction from this PR.

Two possible paths:

1. **Merge this as-is** (it fixes #365 cleanly) and treat #366 as the follow-up that supersedes the nodeSelector lines this PR is making more correct. Acceptable if #366 has a near-term owner.
2. **Combine into a single fix** that drops the nodeSelector lines AND adds the `allowScheduling: false` reconciliation in one PR. Cleaner end state, larger scope.

Either is fine — flagging so the choice is made deliberately rather than discovering later that #366 has to revert the lines this PR just made canonical.

dcmcand:
Heads-up: #372 takes a different architectural direction that overlaps with this PR, so I wanted to flag it here.

This PR correctly fixes the coalescing warning from #365 (passing `nodeSelector` as `map[string]any` so Helm merges it cleanly), and the diagnosis that the warning is cosmetic looks right. The catch is that it makes the `longhornManager`/`longhornDriver` nodeSelector apply *correctly*, which locks in the pinning that #366 identifies as the actual bug: pinning Longhorn's system pods (CSI plugin, manager, engine instance-managers) to the storage nodes breaks PVC mounts for pods on every other node.

#372 instead drops those nodeSelectors (and the `system-managed-components-node-selector` setting) entirely, runs the system pods cluster-wide, and restricts replica placement to storage nodes via `allowScheduling` on the Longhorn `Node` CRs. Removing those keys also clears the empty-map warning #365 reported, so #372 closes #365 as a side effect.

If #372's approach is accepted, this PR would be superseded and can be closed. Flagging rather than closing, since that's the maintainers' call.

dcmcand (approved):
Approving. This is the complete fix for the Longhorn `dedicated_nodes` topology, and I verified it end-to-end on a live EKS cluster.

On the running cluster:
- csi-plugin and instance-manager run on every node, including a tainted non-storage pool
- disks exist only on the storage nodes
- all replicas are confined to the storage nodes
- PVC pods mount on both an untainted and a tainted node, with data readable
- volumes that had faulted before the disk fix self-healed once the disks appeared

The design holds up: the shared `StorageSelector`/`StorageNodeLabels` policy keeps `pkg/storage/longhorn` provider-agnostic, the disk-label injection at kubelet registration is autoscaler-safe, and disk-confinement is a clean replacement for an explicit `allowScheduling` reconcile. The migration runbook and the data-loss warning on toggling `dedicated_nodes` are a real strength.

One non-blocking follow-up: #373 (`warnIfMissingStorageDiskLabel` aborts the deploy on a transient node-list API error, where a genuinely-missing label only warns).

PR #370, opened 2026-06-06
feat(aws): auto-apply nvidia.com/gpu taint to GPU node groups

dcmcand:
## Summary

AWS node groups with `gpu: true` now get the taint `nvidia.com/gpu=true:NO_SCHEDULE` applied automatically, so ordinary pods stay off GPU nodes without anyone hand-writing taint config.

This corrects a common misconception: the NVIDIA GPU Operator does not taint GPU nodes itself. It only ships a toleration for `nvidia.com/gpu` on its own operands, so applying the taint is the platform's job. See [NVIDIA/gpu-operator#26](https://github.com/NVIDIA/gpu-operator/issues/26).

Closes #368.

> [!WARNING]
> **Blocked.** On EKS nothing auto-injects the `nvidia.com/gpu` toleration (the GPU Operator only tolerates the taint on its own operands, and `ExtendedResourceToleration` is not available on EKS). So once this taint lands, GPU workloads in our software packs will stop scheduling onto GPU nodes until each pack tolerates `nvidia.com/gpu`. This PR must not merge until all three are resolved:
>
> - [ ] nebari-dev/nebari-data-science-pack#117 - GPU JupyterHub profiles tolerate the taint
> - [x] nebari-dev/nebari-llm-serving-pack#88 - GPU serving workloads tolerate the taint
> - [x] nebari-dev/nebari-rayserve-pack#14 - Ray Serve GPU workloads tolerate the taint

## What changed

- `pkg/provider/aws/tofu.go`: new `applyGPUTaint` helper and `gpuTaintKey` const. Renamed `resolveNodeGroupAMIs` to `resolveNodeGroupDefaults` since it now resolves both the AMI type and the taint.
- An existing `nvidia.com/gpu` taint (any value or effect) is left untouched; unrelated taints are preserved alongside the GPU taint. The source config is never mutated (the `Taints` slice is copied before appending).
- Scope is AWS only. GKE already auto-taints GPU pools with `nvidia.com/gpu=present:NoSchedule`, so doing the same on GCP would double-taint. Azure is out of scope.
- Docs: `examples/aws-config.yaml` and `docs/design-doc/appendix/16-configuration-reference.md` updated. The reference doc previously showed `effect: NoSchedule`, which is the wrong enum form for the EKS path; that block is removed in favor of the automatic taint.

## Automated checks

- `go test ./...` passes
- `golangci-lint run` (v2): 0 issues
- `go vet` and `gofmt`: clean
- New table-driven tests in `tofu_test.go`: no-taints, existing-gpu-taint-untouched, unrelated-taint-kept, non-gpu, and a no-mutation check.

## Manual test plan

The plan has two goals: (1) confirm the taint is applied and behaves correctly, and (2) confirm nothing else regresses, including existing clusters on upgrade.

### Part 0 - Variable generation (local, no cloud)

1. `make build`.
2. Run a deploy/plan against a config containing a `gpu: true` node group and inspect the generated Terraform variables (or run `tofu plan` in the working dir). Confirm the GPU node group's `taints` includes `{key: nvidia.com/gpu, value: "true", effect: NO_SCHEDULE}` and that non-GPU groups have no such taint.

### Part 1 - Taint applied as intended (AWS, real or staging EKS)

0. **End-to-end JupyterHub GPU profile (primary use case):** with a `gpu: true` node group deployed and tainted, log into JupyterHub and launch a GPU profile that uses the data science software pack. Confirm the server pod schedules onto the GPU node and starts. This only works if the GPU profile carries a toleration for `nvidia.com/gpu`, so as part of this check verify the profile's KubeSpawner config includes that toleration (and a matching nodeSelector). Inside the running server, confirm the GPU is visible (for example `nvidia-smi` returns the device, and a framework such as PyTorch/TensorFlow reports CUDA available). This is the headline scenario: the taint must keep non-GPU pods off the node while still letting the GPU notebook server land on it.
1. **Fresh GPU group, no taints set:** deploy a config with a `gpu: true` node group and no `taints:` block. After the node joins, run `kubectl describe node <gpu-node>` and confirm the taint `nvidia.com/gpu=true:NoSchedule` is present.
2. **Pod without toleration is repelled:** schedule a plain pod with a GPU nodeSelector but no toleration. Confirm it stays `Pending` and does not land on the GPU node.
3. **Pod with toleration schedules:** schedule the same pod with a toleration for `nvidia.com/gpu`. Confirm it schedules onto the GPU node.
4. **GPU Operator still works:** if the NVIDIA GPU Operator is installed, confirm its operands (driver daemonset, device plugin, DCGM) still schedule on the tainted node, since they tolerate the key out of the box. Confirm `nvidia.com/gpu` shows up as allocatable on the node.
5. **Override is respected:** set an explicit `nvidia.com/gpu` taint with a different value/effect (for example `value: present`, `effect: NO_EXECUTE`) on a `gpu: true` group. Confirm NIC leaves it untouched, with no duplicate `nvidia.com/gpu` taint.
6. **Unrelated taint preserved:** set an unrelated taint (for example `dedicated=ml:NO_SCHEDULE`) on a `gpu: true` group. Confirm both that taint and the auto GPU taint are present on the node.

### Part 2 - Does not break anything else (regression)

1. **Non-GPU node groups unaffected:** deploy a config with only standard node groups. Confirm nodes get the standard AMI, no `nvidia.com/gpu` taint, and ordinary pods schedule normally.
2. **Existing taints on non-GPU groups unchanged:** with a Longhorn storage node group (`node.longhorn.io/storage=true:NO_SCHEDULE`), confirm that taint is unchanged and no GPU taint is added.
3. **Idempotency / no churn:** run the deploy twice against an unchanged config. Confirm the second `tofu plan` shows no changes for the GPU node group (no taint add/remove churn).
4. **Upgrade of a cluster with pre-existing untainted GPU nodes (IMPORTANT migration check):** take a cluster that already runs `gpu: true` nodes provisioned before this change (so they currently have no taint). Re-run deploy. Expect the taint to be newly added to the GPU node group. Verify what happens to GPU workloads already running there that lack a matching toleration: with `NoSchedule` they keep running but new untolerated pods will not schedule; confirm no unexpected eviction. The most likely affected workload is the JupyterHub GPU profile: a profile that previously scheduled onto an untainted GPU node with no toleration will stop scheduling new servers once the taint is in place. Confirm the GPU profile tolerates `nvidia.com/gpu` and document this in upgrade notes so operators add the toleration before upgrading.
5. **Full suite:** `go test ./...`, `golangci-lint run`, `go vet ./...` all green.

### Rollback

Revert the commit; existing node groups keep whatever taints are in their config. Nodes provisioned with the auto taint retain it until the node group is replaced or the taint is removed in config.

### DIFF
--- a/docs/design-doc/appendix/16-configuration-reference.md
+++ b/docs/design-doc/appendix/16-configuration-reference.md
@@ -181,17 +181,19 @@ amazon_web_services:
       max_nodes: 10
       taints: []
 
-    # GPU node group example
+    # GPU node group example.
+    # With gpu: true, NIC selects the AL2023_x86_64_NVIDIA AMI and automatically
+    # applies the taint nvidia.com/gpu=true:NO_SCHEDULE so only pods that
+    # tolerate it land on GPU nodes. The NVIDIA GPU Operator does not taint
+    # nodes itself; it only tolerates this taint on its own operands. To use a
+    # different value or effect, set an explicit nvidia.com/gpu taint below and
+    # NIC leaves it untouched.
     gpu:
       instance: g5.2xlarge
       min_nodes: 0
       max_nodes: 5
       gpu: true
       spot: false
-      taints:
-        - key: nvidia.com/gpu
-          value: "true"
-          effect: NoSchedule
 
     # Spot instance node group example
     spot-workers:
--- a/examples/aws-config.yaml
+++ b/examples/aws-config.yaml
@@ -75,7 +75,10 @@ cluster:
       #   instance: g5.xlarge
       #   min_nodes: 0
       #   max_nodes: 3
-      #   gpu: true # Uses AL2023_x86_64_NVIDIA AMI
+      #   gpu: true # Uses AL2023_x86_64_NVIDIA AMI and auto-applies the
+      #             # nvidia.com/gpu=true:NO_SCHEDULE taint (set your own
+      #             # nvidia.com/gpu taint to override). GPU workloads must
+      #             # tolerate it; the NVIDIA GPU Operator tolerates it already.
       #
       # # Example: ARM64 Graviton node group
       # graviton:
--- a/pkg/provider/aws/tofu.go
+++ b/pkg/provider/aws/tofu.go
@@ -14,6 +14,11 @@ const (
 	longhornWebhookConversionKey = "longhorn_webhook_conversion"
 )
 
+// gpuTaintKey is the taint key applied to GPU node groups. It matches the key
+// the NVIDIA GPU Operator's operands tolerate out of the box, so operator
+// components keep scheduling while ordinary pods are kept off GPU nodes.
+const gpuTaintKey = "nvidia.com/gpu"
+
 type TFVars struct {
 	Region                        string               `json:"region"`
 	ProjectName                   string               `json:"project_name,omitempty"`
@@ -49,7 +54,10 @@ type TFVars struct {
 	EnableIRSA                         *bool `json:"enable_irsa,omitempty"`
 }
 
-func resolveNodeGroupAMIs(nodeGroups map[string]NodeGroup) map[string]NodeGroup {
+// resolveNodeGroupDefaults derives per-node-group defaults from the parsed
+// config: the EKS AMI type (NVIDIA for GPU groups, standard otherwise) and the
+// GPU taint. It returns a new map and never mutates the caller's node groups.
+func resolveNodeGroupDefaults(nodeGroups map[string]NodeGroup) map[string]NodeGroup {
 	result := make(map[string]NodeGroup, len(nodeGroups))
 	for name, group := range nodeGroups {
 		if group.AMIType == nil {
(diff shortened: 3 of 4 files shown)


marcelovilla (approved):
@dcmcand thanks for this PR!

I just tested this on a live cluster using the following config:

```yaml
cluster:
  aws:
    region: us-west-2
    kubernetes_version: "1.35"
    availability_zones:
      - us-west-2a
      - us-west-2b
    vpc_cidr_block: "10.10.0.0/16"
    endpoint_private_access: true
    endpoint_public_access: true

    node_groups:
      general:
        instance: m5.2xlarge
        min_nodes: 1
        max_nodes: 1

      gpu:
        instance: g6e.xlarge
        min_nodes: 1
        max_nodes: 5
        gpu: true
        disk_size: 100
```

I can confirm the taint is being correctly applied to the GPU node groups:
```console
kubectl get nodes -o 'custom-columns=NODE:.metadata.name,INSTANCE:.metadata.labels.node\.kubernetes\.io/instance-type,TAINT:.spec.taints[*].key,VALUE:.spec.taints[*].value,EFFECT:.spec.taints[*].effect'

NODE                                          INSTANCE     TAINT            VALUE    EFFECT
ip-10-10-150-6.us-west-2.compute.internal     m5.2xlarge   <none>           <none>   <none>
ip-10-10-158-240.us-west-2.compute.internal   g6e.xlarge   nvidia.com/gpu   true     NoSchedule
```

PR #376, opened 2026-06-09
Remove duplicate CLI error logging and usage output on failures

marcelovilla:
## Closes

Closes #326 

## Description

This PR does two things:

- Pass `SilenceErrors: true` and `SilenceUsage:  true` to Cobra so errors are not duplicated and usage is not shown on failure. A wrong call to the CLI (e.g., wrong flags being passed) will still raise a usage message, which is the expected behavior.
- Remove `slog.Error` calls across different commands, as those are already being logged from main and it only resulted in the triplication of errors.

## How to Test

1. Create a malformed configuration file that will make the CLI raise an error when ran:
    ```bash
    echo 'project_name: "unterminated' > /tmp/bad.yaml
    ```

2. Before building off this branch, run the following to verify the old behavior:
    ```bash
    ./nic validate -f /tmp/bad.yaml
    ```
   
   The output will look something like this:
    ```
    {"time":"2026-06-09T18:37:05.507513+02:00","level":"ERROR","msg":"Configuration validation failed","error":"config file /tmp/bad.yaml: failed to parse YAML: [1:15] could not find end character of double-quoted text\n>  1 | project_name: \"unterminated\n                     ^\n","file":"/tmp/bad.yaml"}
    Error: config file /tmp/bad.yaml: failed to parse YAML: [1:15] could not find end character of double-quoted text
    >  1 | project_name: "unterminated
                         ^
    
    Usage:
      nic validate [flags]
    
    Flags:
      -f, --file string   Path to nebari-config.yaml file (auto-discovered if omitted)
      -h, --help          help for validate
    
    {"time":"2026-06-09T18:37:05.50755+02:00","level":"ERROR","msg":"Command execution failed","error":"config file /tmp/bad.yaml: failed to parse YAML: [1:15] could not find end character of double-quoted text\n>  1 | project_name: \"unterminated\n                     ^\n"}
    ```

    As you can see, the same error is being logged three times. One is the original error, another is the duplicate `slog.Error` inside `validate.go` and the third one comes from cobra surfacing the original error as plain text. Furthermore, usage is printed, even though it is completely irrelevant here.

3. Build off this branch, and re-run the same command:
    
    ```bash
    ./nic validate -f /tmp/bad.yaml
    ```

    The output will now look as it is supposed to, showing only one error and no usage:
    ```
    {"time":"2026-06-09T18:52:21.353657+02:00","level":"ERROR","msg":"Command execution failed","error":"config file /tmp/bad.yaml: failed to parse YAML: [1:15] could not find end character of double-quoted text\n>  1 | project_name: \"unterminated\n                     ^\n"}
    ```

### DIFF
--- a/cmd/nic/main.go
+++ b/cmd/nic/main.go
@@ -18,6 +18,12 @@ var rootCmd = &cobra.Command{
 	Short: "Nebari Infrastructure Core - Cloud infrastructure management for Nebari",
 	Long: `Nebari Infrastructure Core (NIC) is a standalone CLI tool that manages
 cloud infrastructure for Nebari using native cloud SDKs with declarative semantics.`,
+	// Runtime failures are reported once, by main() via slog. Silence cobra's
+	// own error print and usage block so a runtime error is neither duplicated
+	// nor mistaken for misuse of the command. Cobra still prints usage for
+	// genuine argument/flag errors, which it detects before RunE runs.
+	SilenceErrors: true,
+	SilenceUsage:  true,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		logger := slog.New(slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
 			Level: slog.LevelInfo,
--- a/cmd/nic/main.go
+++ b/cmd/nic/main.go
@@ -13,6 +13,13 @@ import (
 	"github.com/nebari-dev/nebari-infrastructure-core/pkg/telemetry"
 )
 
+// reachedRunE reports whether cobra parsed flags and validated args
+// successfully, i.e. PersistentPreRun ran and a command's RunE is about to (or
+// did) execute. main() uses it to distinguish runtime failures (which we log)
+// from usage-class errors (bad flag, unknown command, wrong number of args),
+// which surface before PersistentPreRun and are already printed by cobra.
+var reachedRunE bool
+
 var rootCmd = &cobra.Command{
 	Use:   "nic",
 	Short: "Nebari Infrastructure Core - Cloud infrastructure management for Nebari",
@@ -32,6 +39,7 @@ cloud infrastructure for Nebari using native cloud SDKs with declarative semanti
 		// the error and usage block for those.
 		cmd.SilenceErrors = true
 		cmd.SilenceUsage = true
+		reachedRunE = true
 	},
 }
 
@@ -72,17 +80,15 @@ func main() {
 		}
 	}()
 
-	cmd, err := rootCmd.ExecuteContextC(ctx)
-	if err != nil {
+	if err := rootCmd.ExecuteContext(ctx); err != nil {
 		if ctx.Err() == context.Canceled {
 			slog.Info("Shutdown complete")
 			os.Exit(130)
 		}
-		// SilenceUsage is set in PersistentPreRun, so it is true only once RunE
-		// is reached. We log just those runtime failures and leave usage-class
-		// errors (bad flag, unknown command, bad args) to cobra, which already
-		// printed them.
-		if cmd != nil && cmd.SilenceUsage {
+		// Log only runtime failures (those that occur once RunE is reached) and
+		// leave usage-class errors (bad flag, unknown command, bad args) to
+		// cobra, which already printed them.
+		if reachedRunE {
 			slog.Error("Command execution failed", "error", err)
 		}
 		os.Exit(1)
--- a/cmd/nic/main.go
+++ b/cmd/nic/main.go
@@ -18,17 +18,20 @@ var rootCmd = &cobra.Command{
 	Short: "Nebari Infrastructure Core - Cloud infrastructure management for Nebari",
 	Long: `Nebari Infrastructure Core (NIC) is a standalone CLI tool that manages
 cloud infrastructure for Nebari using native cloud SDKs with declarative semantics.`,
-	// Runtime failures are reported once, by main() via slog. Silence cobra's
-	// own error print and usage block so a runtime error is neither duplicated
-	// nor mistaken for misuse of the command. Cobra still prints usage for
-	// genuine argument/flag errors, which it detects before RunE runs.
-	SilenceErrors: true,
-	SilenceUsage:  true,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		logger := slog.New(slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
 			Level: slog.LevelInfo,
 		}))
 		slog.SetDefault(logger)
+
+		// PersistentPreRun runs only after cobra has parsed flags and validated
+		// args. Any failure from here on is a runtime error and not a misuse,
+		// so silence cobra's own error/usage output and let main() report it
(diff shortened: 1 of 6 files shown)


viniciusdc:
Thanks @marcelovilla, this cleanly fixes #326. I built it and confirmed the triplication collapses to a single `Command execution failed` line with no usage block. The dedup is applied uniformly across all the `cmd/nic` commands (`version` already conformed), and nothing is swallowed: every dropped `slog.Error` keeps its `return` and `span.RecordError`. The `kubeconfig.go` os.Write paths going from `slog`+bare-return to `fmt.Errorf("...: %w", ...)` is a nice bonus, it actually carries more context than before.

One thing worth settling before this merges, and it's about the comment/description more than the code.

The comment on `SilenceUsage` (and the PR description) says "Cobra still prints usage for genuine argument/flag errors." That's not actually what happens. `SilenceUsage: true` on the root suppresses the usage block for flag-parse and unknown-subcommand errors too, not just runtime errors. There's no custom `FlagErrorFunc` to bring it back, so:

```
$ ./nic validate --bogus-flag
ERROR Command execution failed error="unknown flag: --bogus-flag"   # exit 1, no usage
$ ./nic boguscmd
ERROR Command execution failed error="unknown command \"boguscmd\" for \"nic\""   # exit 1, no usage
```

A couple of follow-ups, not blocking:

- A tiny regression test would be worth it here, since this is silently reintroducible. The next person who adds a command with a `slog.Error(...)` before `return` brings #326 right back with no signal. A bad `--timeout=nonsense` hits a RunE error with no cloud call, and there's already a `captureStdout` helper in `deploy_dns_test.go` to assert cobra prints no error/usage.
- AGENTS.md's Logging Convention still literally says "Use `slog.Error()` in `cmd/nic/` commands," which is what invites this bug back. I think it's worth a line codifying the rule this PR establishes: within a RunE, record on the span and return the wrapped error; don't `slog.Error` something you also return, since `main()` logs it once.

marcelovilla:
@viniciusdc thanks for catching that. I somehow managed to test against a stale binary and thought things were working as expected. It was definitely not my intention to lose cobra's usage and error report when trying to run a subcommand that does not exist or using wrong flags.

I moved `SilenceErrors` and `SilenceUsage` inside `PersistentPreRun` to allow cobra to check the command usage first, and added a small guard around `slog.Error` to have it only print application errors rather than usage errors, which should be reported by cobra itself. Check 2d8fc0806020f105d23eca797ac609e51218b998

I also updated `AGENTS.md` with a small note about not logging errors that are returned inside the commands. See 0fb2c628c0dab84a165647549a8f3fc004ca63c3

Would you be able to test this again?

cmd/nic/main.go
```diff
@@ -63,12 +72,19 @@ func main() {
...
 			slog.Info("Shutdown complete")
 			os.Exit(130)
 		}
-		slog.Error("Command execution failed", "error", err)
+		// SilenceUsage is set in PersistentPreRun, so it is true only once RunE
+		// is reached. We log just those runtime failures and leave usage-class
+		// errors (bad flag, unknown command, bad args) to cobra, which already
+		// printed them.
+		if cmd != nil && cmd.SilenceUsage {
+			slog.Error("Command execution failed", "error", err)
```
dcmcand, line 92: Clever and it works today, but using `cmd.SilenceUsage` as the proxy for "we reached `RunE`, so log it" couples cobra's *output config* to the app's *logging decision*. If someone later sets `SilenceUsage: true` directly on a command (a common cobra pattern in the command struct), usage-class errors for that command would start flowing through this branch and getting logged. A dedicated signal (e.g. a flag set in `PersistentPreRun`) would express the intent without the side-coupling. Not blocking - the comment documents the assumption well.

cmd/nic/main.go
```diff
@@ -23,6 +23,15 @@ cloud infrastructure for Nebari using native cloud SDKs with declarative semanti
 			Level: slog.LevelInfo,
 		}))
 		slog.SetDefault(logger)
+
+		// PersistentPreRun runs only after cobra has parsed flags and validated
+		// args. Any failure from here on is a runtime error and not a misuse,
+		// so silence cobra's own error/usage output and let main() report it
+		// once via slog. Usage-class errors (bad flag, unknown command, wrong
+		// number of args) surface before this hook runs, so cobra still prints
```
dcmcand, line 38: Really appreciate this comment - it captures the one subtlety (PersistentPreRun runs after parse/validate, so usage-class errors surface before it) that makes the whole approach correct.

dcmcand (dismissed):
Clean, well-scoped fix that goes after the root cause of #326 rather than the symptom. I built the branch and verified all three paths: a runtime error now produces a single structured log with no usage block, while bad flags and unknown commands still get cobra's error + usage output. `go build`/`vet`/`test`, `gofmt`, and `golangci-lint` are all clean.

A few things I really liked:

- **Root cause, not symptom.** The triplication came from three sources (cobra's plain-text print, the redundant in-`RunE` `slog.Error`, and main's `slog.Error`). You cut exactly the right two, leaving one structured log at the application layer - in line with the repo's "log only at `cmd/nic`" convention.
- **OTel preserved.** Every `span.RecordError(err)` stays; only the duplicate `slog.Error` lines are gone. That's the part that mattered.
- **The inline comments in `main.go` are genuinely excellent.** The cobra wiring here is non-obvious, and explaining *why* `PersistentPreRun` is the right seam (it runs only after flag/arg parsing, so anything past it is a runtime error, not a misuse) will save the next person real time.
- **Docs updated in the same PR** (`AGENTS.md`), with a back-reference to #326. The convention now matches the code.
- **Nice opportunistic cleanup** in `kubeconfig.go`: two bare `return err` upgraded to `fmt.Errorf(...: %w)` with operational context.

Approving. Two non-blocking thoughts left as inline comments.

cmd/nic/main.go
marcelovilla: That's a fair point. Addressed in 6ca162036010026c4ced6cd6d084fb8becdf9427

dcmcand (approved):
**Approve** :white_check_mark:

Clean, well-reasoned fix for the triple-logging in #326. A few things I really liked:

- Setting `SilenceErrors`/`SilenceUsage` inside `PersistentPreRun` rather than statically is exactly the right cobra idiom - it keeps the helpful usage output for flag/arg mistakes while suppressing it for runtime failures. Nicely targeted rather than a blunt global silence.
- The `reachedRunE` comment block clearly explains the invariant and the cobra parse-before-PreRun ordering it depends on. I verified that ordering and that no subcommand defines its own `PersistentPreRun` to shadow the root hook - the gate is sound for every command.
- Codifying the "don't log an error you also return" rule in AGENTS.md (with the #326 reference) puts the fix where it prevents recurrence.
- The `fmt.Errorf("write kubeconfig file %q: %w", ...)` wrapping in `kubeconfig.go` is a genuine improvement independent of the logging fix - the filename now travels with the error.

Non-blocking follow-ups for whenever: `version.go`'s `runVersion` is the one `RunE` still returning a `NewClient` error without `span.RecordError` (pre-existing, untouched here), and the bare `ParseConfig`/`NewClient` returns could be `%w`-wrapped for symmetry with the kubeconfig paths. Neither gates this.

Issue #71, opened 2026-02-09
Wire DNS provider cert-manager config into foundational services deploy


dcmcand:
## Summary

When deploying foundational services, cert-manager needs DNS provider configuration (Cloudflare zone info, API token secret reference) to perform DNS-01 challenges for Let's Encrypt TLS certificates.

The `DNSProvider.GetCertManagerConfig()` method exists in the interface but is not wired into the deploy flow. It needs to be called **before** foundational services install so the cert-manager ClusterIssuer gets the correct DNS-01 solver config.

## Deploy flow ordering

```
1. Deploy infrastructure (OpenTofu)
2. Get cert-manager config from DNS provider  ← GetCertManagerConfig()
3. Install ArgoCD
4. Install foundational services (passing cert-manager config)
5. Wait for LB endpoint
6. Provision DNS records                       ← ProvisionRecords() (issue #70)
```

## What needs to happen

- Call `dnsProvider.GetCertManagerConfig(ctx, cfg)` before `argocd.InstallFoundationalServices()`
- Pass the returned config into `FoundationalConfig` (may need a new field)
- Generate a cert-manager ClusterIssuer with Cloudflare DNS-01 solver
- Create a Kubernetes secret with the Cloudflare API token for cert-manager to reference

## Dependencies

- #70 — Implements the Cloudflare DNS provider with `ProvisionRecords`/`DestroyRecords`
- cert-manager must be deployed as a foundational service

## Related issues

- #67 — Let's Encrypt certificate fails: wildcard + explicit subdomains rejected by ACME
- #68 — cert-manager missing Gateway API config for HTTP-01 solver

PR #377, opened 2026-06-09
feat(certificate): add dns01 ACME challenge for private networks

tylerpotts:
## Summary

- Adds `certificate.acme.challenge` config field (`http01` default, `dns01`) so Let's Encrypt certificates can be issued for clusters the ACME server cannot reach — private IPs, VPN-only deployments. DNS-01 validates ownership via DNS TXT records instead of inbound HTTP.
- When `dns01` is selected, the `letsencrypt-issuer` ClusterIssuer renders a Cloudflare `dns01` solver (`apiTokenSecretRef` → `cloudflare-api-token`/`api-token`), and the foundational bootstrap creates that Secret in the `cert-manager` namespace from `CLOUDFLARE_API_TOKEN` (the same credential the DNS record provisioner uses) before the root App-of-Apps syncs the issuer. The token is never written to the GitOps repository.
- Config validation requires a `dns:` block with the `cloudflare` provider when `dns01` is chosen, and rejects unknown challenge values. The default `http01` path renders byte-identical to before.

Cloudflare is the only solver wired up since it is the only DNS provider in tree; additional hosts (route53, azure-dns) can be added behind the same `challenge` field later.

Refs https://github.com/nebari-dev/nebari-infrastructure-core/issues/71

## Test plan

- [x] `go test -short ./...` passes
- [x] Unit tests: challenge validation (valid/invalid/missing dns block), `NeedsDNS01Solver`, ClusterIssuer template rendering for both solvers, template-data plumbing, solver Secret creation (fake clientset, including missing-env error)
- [x] `golangci-lint run` — 0 issues
- [ ] End-to-end: deploy with `challenge: dns01` on a private-network cluster and confirm cert issuance (planned on the nrec test rig)

### DIFF
--- a/examples/aws-config-with-dns.yaml
+++ b/examples/aws-config-with-dns.yaml
@@ -6,6 +6,13 @@ certificate:
   type: letsencrypt
   acme:
     email: admin@example.com
+    # challenge selects how Let's Encrypt validates domain ownership:
+    #   http01 (default) - ACME server connects to the cluster's gateway;
+    #                      requires the cluster to be reachable from the internet
+    #   dns01            - validation via DNS TXT records; works for private/VPN
+    #                      clusters and requires the dns block below (cert-manager
+    #                      uses the same CLOUDFLARE_API_TOKEN credential)
+    # challenge: dns01
 
 # GitOps repository configuration (required for foundational services)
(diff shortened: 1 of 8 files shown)


tylerpotts:
closed as we don't need this right now

PR #404, opened 2026-06-18
feat(certificate): support user-supplied TLS cert for the gateway

tylerpotts:
## Summary

Closes https://github.com/nebari-dev/nebari-infrastructure-core/issues/333

Allow operators to supply their own TLS certificate for the Nebari gateway instead of having cert-manager mint one via the self-signed/Let's Encrypt `ClusterIssuer`.

Adds `certificate.type: existing` with three mutually-exclusive sources:

| Source | Behavior |
|--------|----------|
| `existing_secret` | References a `kubernetes.io/tls` secret the user manages. When it lives outside `envoy-gateway-system`, NIC sets the `certificateRefs` namespace **and** renders a Gateway-API `ReferenceGrant`. |
| `files` | NIC reads PEM cert/key from disk and creates the secret directly in `envoy-gateway-system`. |
| `env` | NIC reads raw (non-base64) PEM from env vars and creates the secret directly. |

For `files`/`env` the material is written **only** to the cluster secret — never to the GitOps repo. Cert/key bytes are never logged or recorded in OpenTelemetry span attributes.

## Behavior details

- **Validation** rejects: missing source, more than one source, partial `files`/`env` pairs, `existing_secret` without a name, `secret_name` combined with `existing_secret`, and `acme` combined with `type=existing`.
- **cert-manager skipped**: when `type=existing`, both `gateway-certificate.yaml` and its Argo CD `certificates` Application are skipped (the app would otherwise sync an empty directory under `allowEmpty: false`).
- **SAN check**: NIC parses the resolved cert and **warns (does not fail)** when `<domain>`, `keycloak.<domain>`, or `argocd.<domain>` is missing from the SANs. Wildcard + apex is honored.
- **Fail-fast pre-flight**: `files`/`env` material is validated (readable + valid keypair) at deploy start, before any cloud provisioning, so a bad path or mismatched keypair fails fast instead of producing a silently broken gateway.

## Testing

- New table-driven unit tests in `pkg/config` and `pkg/argocd` cover parsing, every validation rejection, same-ns/cross-ns cases, file source, env source, idempotency, the cert-manager/app skip, and the SAN warning (including wildcard coverage).
- `make check` passes (fmt, vet, lint, race + coverage tests).
- Verified end-to-end through the real CLI: the example config validates, and ambiguous configs are rejected.

## Docs & example

- `docs/custom-tls-certificate.md` — SAN requirements, the three sources, `kubectl` secret creation, config snippets, the cross-namespace + `ReferenceGrant` case (with the namespace prerequisite), and a troubleshooting table.
- `examples/custom-tls-config.yaml` — annotated working config.

## Not covered here

The issue's manual on-cluster verification (deploy each source, hit `argocd.<domain>`/`keycloak.<domain>` in a browser, confirm the served cert) requires a live cluster.

### DIFF
--- a/pkg/argocd/certificate.go
+++ b/pkg/argocd/certificate.go
@@ -149,8 +149,8 @@ func createGatewayTLSSecret(ctx context.Context, client kubernetes.Interface, cf
 			Name:      name,
 			Namespace: namespace,
 			Labels: map[string]string{
-				"app.kubernetes.io/name":       "nebari-gateway",
-				"app.kubernetes.io/managed-by": "nebari-infrastructure-core",
+				"app.kubernetes.io/name": "nebari-gateway",
+				ManagedByLabel:           NebariManagedByValue,
 			},
 		},
 		Type: corev1.SecretTypeTLS,
--- a/pkg/argocd/certificate.go
+++ b/pkg/argocd/certificate.go
@@ -1,0 +1,255 @@
+package argocd
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"os"
+	"strings"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/config"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/status"
+)
+
+// PreflightGatewayTLS validates user-supplied certificate material for the
+// files and env sources before any cluster mutation or GitOps commit, so a bad
+// path, unset env var, or mismatched keypair fails the deploy fast instead of
+// silently rendering a gateway that can never terminate TLS.
+//
+// It is a no-op for non-existing certificate types and for the existing_secret
+// source (which is validated against the live cluster during ConfigureGatewayTLS).
+// No certificate or key material is logged.
+func PreflightGatewayTLS(cfg *config.NebariConfig) error {
+	cert := cfg.Certificate
+	if cert == nil || cert.Type != config.CertificateTypeExisting {
+		return nil
+	}
+
+	var certPEM, keyPEM []byte
+	var err error
+	switch {
+	case cert.Files != nil:
+		certPEM, keyPEM, err = readCertFiles(cert.Files)
+	case cert.Env != nil:
+		certPEM, keyPEM, err = readCertEnv(cert.Env)
+	default:
+		// existing_secret is validated against the cluster later.
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	if _, err := tls.X509KeyPair(certPEM, keyPEM); err != nil {
+		return fmt.Errorf("invalid TLS certificate/key pair: %w", err)
+	}
+	return nil
+}
+
+// ConfigureGatewayTLS handles user-supplied TLS certificates
+// (certificate.type=existing). For the files and env sources it reads the PEM
+// material and creates a kubernetes.io/tls secret directly in
+// envoy-gateway-system. For the existing_secret source it fetches the
+// already-present secret. For every source it parses the resolved certificate
+// and warns (never fails) when a recommended SAN is missing.
+//
+// Certificate and key material is never logged or recorded in span attributes.
+// For non-existing certificate types this is a no-op.
+func ConfigureGatewayTLS(ctx context.Context, client kubernetes.Interface, cfg *config.NebariConfig) error {
+	cert := cfg.Certificate
+	if cert == nil || cert.Type != config.CertificateTypeExisting {
+		return nil
+	}
+
+	tracer := otel.Tracer("nebari-infrastructure-core")
+	ctx, span := tracer.Start(ctx, "argocd.ConfigureGatewayTLS")
+	defer span.End()
+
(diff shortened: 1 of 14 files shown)


pkg/nic/deploy.go
```diff
@@ -94,6 +94,15 @@ func (c *Client) Deploy(ctx context.Context, cfg *config.NebariConfig, opts Depl
 		WithMetadata("provider", cfg.Cluster.ProviderName()).
 		WithMetadata("project_name", cfg.ProjectName))
 
+	// For user-supplied certificates sourced from files/env, validate the
+	// material is readable and a valid keypair before provisioning anything.
+	// This turns a local config error into a fast failure instead of a silently
+	// broken gateway discovered after the cluster is up.
+	if err := argocd.PreflightGatewayTLS(cfg); err != nil {
```
dcmcand, line 101: Great placement. Validating files/env material before any cloud provisioning turns a silently-broken gateway into a fast, local config error. Exactly the right place for this.

pkg/argocd/writer.go
```diff
@@ -339,6 +359,33 @@ func isMetalLBPath(relPath string) bool {
...
+// cross-namespace existing secret.
+//
+// The certificates Application is skipped alongside the Certificate because the
+// gateway cert is the only resource in manifests/security/certificates. Leaving
+// the Application would point it at an empty directory (allowEmpty: false) and
+// it would report as failed.
+func skipCertificateTemplate(relPath string, data TemplateData) bool {
+	switch relPath {
+	case gatewayCertificatePath, certificatesAppPath:
+		return data.UseExistingCertificate
```
dcmcand, line 381: This is the subtle-but-important bit done right: skipping both the Certificate and the certificates Argo CD app for `type=existing` keeps cert-manager from fighting over the secret and dodges the `allowEmpty: false` empty-dir failure. The explanatory comment above makes the intent obvious for the next reader.

pkg/argocd/certificate.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("parse certificate: %w", err)
+	}
+
+	required := []string{domain, "keycloak." + domain, "argocd." + domain}
+	var missing []string
+	for _, host := range required {
+		if cert.VerifyHostname(host) != nil {
```
dcmcand, line 255: Using `x509.VerifyHostname` here means wildcard SANs are honored for free, and warning (rather than failing) keeps legitimate certs with extra hostnames from being rejected. Matches the issue's intent precisely.

pkg/argocd/templates/manifests/networking/gateway-tls-referencegrant.yaml
```diff
@@ -0,0 +1,20 @@
...
+    app.kubernetes.io/managed-by: nebari-infrastructure-core
+spec:
+  from:
+    - group: gateway.networking.k8s.io
+      kind: Gateway
+      namespace: envoy-gateway-system
+  to:
+    - group: ""
+      kind: Secret
+      name: {{ .GatewayTLSSecretName }}
```
dcmcand, line 20: Scoping the grant to the specific secret name rather than all secrets in the namespace is the least-privilege choice. Nice.

pkg/argocd/install.go
```diff
@@ -147,6 +147,21 @@ func Install(ctx context.Context, cfg *config.NebariConfig, prov provider.Provid
 		}
 	}
 
+	// Configure a user-supplied gateway TLS certificate (certificate.type=existing).
+	// No-op for cert-manager-issued certs. For files/env sources this creates the
```
dcmcand: Non-blocking question: a `ConfigureGatewayTLS` failure is downgraded to a warning for all sources. For files/env this is the only step that places the cert in the cluster, so a non-transient failure here (RBAC/quota denial) would let the deploy report success while the gateway has no usable cert. Since the user explicitly supplied a cert and preflight already validated the material, would it be worth making secret-creation failures fatal for files/env specifically, while keeping the advisory behavior for existing_secret? Happy either way - just flagging.

dcmcand (dismissed):
Nice work, Tyler. This is a clean, well-tested implementation of #333 and it reads like it fits the codebase rather than fighting it.

A few things I want to call out specifically:

- **Fail-fast preflight before any provisioning** is exactly the right call - a bad path or mismatched keypair becomes an instant local error instead of a gateway that silently can't terminate TLS after the cluster is already up.
- **cert-manager is cleanly stepped aside** for `type=existing`: skipping both the Certificate and its Argo CD app avoids cert-manager fighting over the secret *and* the `allowEmpty: false` empty-directory failure. The rationale being in the code comment is appreciated.
- **Security is handled carefully** - cert/key bytes never get logged and never land in span attributes (only the source enum does), and nothing is written to the GitOps repo.
- **Thorough table-driven tests** across `pkg/config` and `pkg/argocd` covering the validation rejections, same-ns/cross-ns, file/env sources, idempotency, and the SAN warning.

Approving. One non-blocking thought left inline on `install.go` about warning-vs-fatal for the files/env secret-creation path - worth a look but I don't think it should hold up merge.

pkg/argocd/install.go
tylerpotts: Good call - made it fatal. The call site now returns the error instead of warning (4c18d13). This naturally scopes to files/env: `ConfigureGatewayTLS` only ever returns a non-nil error from those sources (where secret creation is the only step that places the cert), while the `existing_secret` path is advisory-only and always returns nil, so it keeps its non-fatal behavior without the call site branching on source. Documented the invariant on the function and the call site.

dcmcand (approved):

PR #405, opened 2026-06-19
Move provider and dnsprovider into pkg/providers

marcelovilla:
## Closes

Closes #381

## Description

This PR moves everything under `pkg/provider` and `pkg/dnsprovider` to a shared providers folder in `pkg/providers`.

This is more consistent than the current separation and will allow a smoother addition of new provider types (i.e., `git` and `certificate`) in the future. Also, it paves the way to a cleaner [out-of-tree provider architecture](https://github.com/nebari-dev/nebari-infrastructure-core/pull/256), where the new umbrella `pkg/providers/` folder would host the interfaces for each different provider type and their concrete implementations would live in external repos.

## How to Test

This PR does not introduce any new feature nor does change any of the existing behavior. Testing should be already covered by CI itself. If you want to double-check, feel free to build off this branch and run a `nic deploy -f examples/<any-example>.yaml --dry-run`. That should confirm all the wiring is still in place.


dcmcand (dismissed):
Clean, well-executed restructuring. `go build ./...`, `go vet ./...`, and the provider test suites all pass on the PR head. The abstraction boundaries are intact: `pkg/nic/registry.go` remains the sole concrete-provider import site, no cross-provider imports were introduced, and zero stale `pkg/provider`/`pkg/dnsprovider` Go imports remain. The new `cluster.Provider`/`cluster.InfraSettings` and `dns.Provider` names read more naturally than the old ones. Nice work.

Approving. Two non-blocking items:

### Question: Azure provider version bump

The description says this "does not change any of the existing behavior," but the Azure lockfile was regenerated rather than relocated: `azurerm` moved from **4.76.0 to 4.78.0** (and the `azapi` hashes changed) in the "Update tofu lockfiles" commit. The AWS lockfile is a clean rename with identical content, so only Azure drifted.

Was this an intentional bump, or a side effect of running `tofu init` in the relocated `templates/` directory? It is low-risk, but it is a real dependency change inside an otherwise behavior-neutral PR. Either pinning it back to 4.76.0 or noting the bump in the description would keep the "no behavior change" claim honest.

### Follow-up: stale paths under `docs/design-doc/`

The doc sweep updated the root-level docs but stopped before `docs/design-doc/`, which still points at the old layout. Could you open a follow-up PR (or issue) to refresh these?

- `docs/design-doc/operations/12-testing-strategy.md` - contains `go test ./pkg/provider/aws`, which now fails if a contributor runs it (highest-value fix)
- `docs/design-doc/implementation/09-dns-provider-architecture.md` (marked "Status: Implemented") - `pkg/dnsprovider/...` paths
- `docs/design-doc/appendix/16-configuration-reference.md` - `pkg/dnsprovider/cloudflare/config.go`, `pkg/provider/hetzner/config.go`
- `docs/design-doc/status-updates.md` - several `pkg/provider/*` references

The ADRs (`0002`, `0004`) also reference old paths but are historical decision records, so leave those as-is.

marcelovilla:
Thanks for the review @dcmcand. The Azure lockfile update (and AWS too) were intentional because the OpenTofu lockfile drift workflow was complaining: https://github.com/nebari-dev/nebari-infrastructure-core/actions/runs/27826744646/job/82353006984

We need to rethink that workflow because it's forcing us to update when there are clearly no relevant changes but alas, here we are. I don't think it's bad to include the updates here even if they're not fully consistent with the intention of this PR.

Regarding the stale docs, I addressed it in f54f5022abee303b1a02590193a54aa01926bd13

dcmcand (approved):
Re-approving. The Azure lockfile bump is confirmed intentional (forced by the lockfile-drift workflow) and the stale docs are resolved in f54f502 - including the `12-testing-strategy.md` command that was broken. Agree the drift workflow is worth a separate look, but that's out of scope here. Thanks for the quick turnaround.

PR #432, opened 2026-06-26
fix(keycloak): realm-setup job works on OpenShift (arbitrary UID)

brandonrc:
## Problem
On OpenShift, the `keycloak-realm-setup` job never completes, so the `nebari` realm + OIDC clients are never created and **Nebari login returns "Page not found"** (the realm 404s).

Two root causes, both from OpenShift running the pod as an **arbitrary high UID with no `/etc/passwd` entry**:
1. `$HOME` resolves to `/` (read-only), so `kcadm config credentials` can't write `~/.keycloak/kcadm.config` → fails. The readiness loop swallowed the error (`2>/dev/null`) and looped "Keycloak not ready" until giving up, then silently no-op'd realm creation.
2. Even with `HOME=/tmp`, `kcadm` is a Java tool and resolves its config dir from `user.home` (the OS passwd entry), **not** `$HOME` — so for an arbitrary UID the saved session still isn't found and every call after `config credentials` fails `No server specified`.

## Fix
- Wrap `kcadm` to pass **`--config /tmp/kcadm.config`** on every call — an explicit, world-writable, UID-independent path for the saved session. (No-op on vanilla k8s.)
- Set `HOME=/tmp` as well.
- Make the readiness loop **fail loudly** with the real error instead of silently continuing.

## Testing
Verified end-to-end on a ROSA HCP cluster: realm-setup completes, the `nebari` realm + landing/argocd OIDC clients are created, and Nebari login works. Confirmed a clean from-scratch deploy reaches `realm_ready` with a working OIDC authorize flow.

> Draft for review — surfaced while building an OpenShift cluster provider (separate PR), but this fix is independent and benefits any OpenShift deployment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/argocd/templates/manifests/keycloak/realm-setup-job.yaml
+++ b/pkg/argocd/templates/manifests/keycloak/realm-setup-job.yaml
@@ -49,12 +49,20 @@ spec:
             - -c
             - |
               set -e
-              KCADM="/opt/keycloak/bin/kcadm.sh"
+              # kcadm stores its authenticated session in a config file. On
+              # OpenShift this pod runs as an arbitrary UID with no /etc/passwd
+              # entry, so kcadm's default ~/.keycloak/kcadm.config resolves to a
+              # non-writable home: `config credentials` cannot persist the token
+              # and every later call fails with "No server specified". Pin the
+              # config file to an explicit, world-writable path via --config on
+              # every invocation (UID-independent; a no-op on vanilla k8s).
+              KCCFG=/tmp/kcadm.config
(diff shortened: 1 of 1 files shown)


brandonrc:
Closing — this change only matters on OpenShift (it's a no-op on vanilla Kubernetes), so it doesn't stand alone as a general fix. It's already included in the OpenShift support PR #433, which is the coherent home for it. The fork branch `fix/keycloak-openshift-realm-setup` can be deleted.

PR #433, opened 2026-06-26
feat: OpenShift (ROSA HCP) cluster provider

brandonrc:
## Summary
Adds a first-class **`openshift`** cluster provider so NIC can deploy Nebari onto Red Hat OpenShift, targeting ROSA HCP first. Built and validated end-to-end on a live ROSA HCP cluster, including a full **destroy → rebuild from scratch** with working login and a software pack deployed via GitOps.

## What's here
- **Dual-mode provider** (`pkg/providers/cluster/openshift/`): `existing` (deploy onto a running cluster) and `provision` (ROSA HCP via OpenTofu — `terraform-redhat/rosa-hcp` + `rhcs`, validated against the module schema). Registered in `pkg/nic/registry.go`.
- **Automatic SCC bootstrap** — grants the SCC to the foundational namespaces (and configurable `scc.extra_namespaces` for software packs) **before** ArgoCD installs, so pods that pin a fixed UID + seccomp (e.g. `argocd-redis`) schedule on the first try. This was the single biggest manual step on OpenShift; it's now automated.
- **OpenShift-aware `InfraSettings`**: `gp3-csi` default storage, no MetalLB, AWS in-tree `nlb` LB annotation (ROSA ships no AWS LB Controller).
- **Keycloak realm-setup OpenShift fix** — the `keycloak-realm-setup` job silently failed under OpenShift's arbitrary UID (kcadm couldn't write its `~/.keycloak` config), so the `nebari` realm was never created and login 404'd. The job now pins `kcadm --config /tmp/kcadm.config` (UID-independent) and fails loudly. No-op on vanilla Kubernetes.
- Example config (`examples/openshift-config.yaml`) + unit tests throughout.

## Validation
Clean from-scratch ROSA HCP build: provider auto-applies SCCs → ArgoCD + full foundational stack (`cert-manager`, Envoy Gateway, Keycloak, postgres, nebari-operator, landing, OTel) reach Healthy, **login works** (realm + OIDC clients provisioned), and a `provenance-collector` software pack syncs Healthy via the app-of-apps.

## Open items for discussion
- `provision` mode kubeconfig retrieval (OCM/rosa) is stubbed; `existing` mode is fully wired. Provision templates are schema-validated but full provision-mode e2e is pending.
- Ingress/TLS on the ephemeral `*.apps` domain still needs OpenShift Routes + cert-acceptance; a real domain + `letsencrypt` is the production path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


brandonrc:
### End-to-end ROSA HCP provisioning wired + validated live

Provision mode now stands up a real ROSA HCP cluster and tears it down cleanly. Pushed in `9177aee`.

**What's added**
- `state.go` — S3 remote state (provision mode), mirroring the aws provider
- `deploy.go` — `deployProvision` (ensure bucket → tofu init/apply → SCC bootstrap) and `destroyProvision` (destroy + orphan-sweep retry)
- `kubeconfig.go` — provision kubeconfig via `rosa create admin` → polled `oc login` (ROSA has no EKS-style token plugin)
- `cleanup.go` — VPC orphan sweep for resources not in tofu state that block `DeleteVpc`
- gated live E2E harness (`-tags ocp_e2e`) + unit tests + `E2E.md` runbook

**Validated against live ROSA HCP (us-east-1) this session**
- ✅ dry-run `tofu plan` (no resources)
- ✅ full lifecycle: provision → kubeconfig → SCC bootstrap → `oc get nodes` (2 workers Ready, OCP 4.20 / k8s v1.33)
- ✅ provision-only + verify, twice
- ✅ **hands-off destroy**: first `tofu destroy` hit the known ROSA teardown snag — `DeleteVpc` `DependencyViolation` from the PrivateLink endpoint's leftover security group (not in tofu state). The new sweep auto-deleted the orphan and retried to completion. Final leak check: 0 VPCs, 0 state buckets, 0 IAM roles, cluster gone.

**Known follow-ups**
- Full Nebari app-stack deploy → reachable landing page is documented as a two-phase runbook in `E2E.md` but not yet run live (the LB-sweep branch of the teardown fix is exercised by that path).
- Account-global `ManagedOpenShift-HCP-*` roles are created per-deploy under a shared prefix and intentionally survive teardown; consider managing them out-of-band.

pkg/providers/cluster/openshift/provider.go
```diff
@@ -0,0 +1,69 @@
...
+		span.RecordError(err)
+		return nil, fmt.Errorf("failed to unmarshal openshift config: %w", err)
+	}
+	return &cfg, nil
+}
+
+// Validate is implemented in validate.go.
+// Deploy/Destroy/Summary are implemented in deploy.go.
+// GetKubeconfig is implemented in kubeconfig.go.
+// InfraSettings is implemented in infra.go.
```
marcelovilla, line 69: Why not leaving the implementations in provider.go? I think it might make sense to split it but none of the other providers do so just flagging the asymmetry

pkg/providers/cluster/openshift/provision_e2e_test.go
marcelovilla, line 1: We need e2e tests for the cluster providers but I'm inclined towards implementing them in our GHA workflows, and have them test not just the cluster provision, but also the repo, dns, certificate, and foundational software. Given that we have not set on a testing strategy yet, I'd like to avoid introducing patterns that won't we followed by other providers.

pkg/providers/cluster/openshift/config.go
```diff
@@ -0,0 +1,122 @@
...
+	ExtraNamespaces []string `yaml:"extra_namespaces,omitempty"`
+}
+
+// Config is the cluster.openshift provider configuration. It is dual-mode: a
+// `mode` discriminator selects between provisioning a ROSA HCP cluster
+// (provision) and targeting an existing OpenShift cluster (existing). Fields not
+// relevant to the active mode are ignored.
+type Config struct {
+	// ModeField selects provision (default) or existing. Exposed via Mode().
+	ModeField string `yaml:"mode,omitempty"`
```
marcelovilla, line 55: I don't think having a `mode` switch is the right approach. Cluster providers are meant to provision infrastructure, and there's already an `existing` provider that is meant for users to deploy the platform (i.e., the foundational software) on top of existing infrastructure.

After looking into what the existing mode does in your PR, I'd say `storage_class` and `load_balancer_annotations` are already handled by the `existing` provider and the new piece is the Security Context Constraints that are granted to the foundational software before ArgoCD's Helm install.

I don't think this should be handled by the cluster provider, but rather by a distribution-aware step between `provider.Deploy` and `argocd.Install`. That way, the `existing` provider would cover OpenShift clusters and a ROSA HCP provider could make use of it. This needs some more thought and discussion, though.

pkg/argocd/templates/manifests/keycloak/realm-setup-job.yaml
marcelovilla, line 1: This seems to me like a valuable addition and I like the approach of failing loudly. Regardless of the ROSA HCP provider, I think this warrants an independent PR.

pkg/providers/cluster/openshift/state.go
marcelovilla, line 1: It seems this is the same logic as in the AWS provider. I'd consider moving it into a shared helper that both providers can call.

pkg/providers/cluster/openshift/kubeconfig.go
```diff
@@ -0,0 +1,256 @@
...
+	out, err := pollOCLogin(ctx, apiURL, adminUsername, password)
+	if err != nil {
+		return nil, err
+	}
+
+	p.kubeconfigMu.Lock()
+	p.kubeconfigCache[cacheKey] = out
+	p.kubeconfigMu.Unlock()
+	return out, nil
+}
```
marcelovilla, line 152: It seems this could be made more straightforward by either using [OpenTofu's rhcs identity provider](https://registry.terraform.io/providers/terraform-redhat/rhcs/latest/docs/resources/identity_provider) or the [OCM Go SDK](https://github.com/openshift-online/ocm-sdk-go) to create the cluster admin credentials and then doing an OAuth token exchange to get the kubeconfig.

That way, it does not rely on two different CLI tools (i.e. `rosa` and `oc`) being available.

marcelovilla (commented):
Thanks for this PR @brandonrc!

I went over your suggested changes and left several comments. As I mentioned, I'm not familiar with OpenShift, nor ROSA HCP so there's a couple of things here that are new to me, such as SCC. Besides the concerns I left in the comments, a quick chat with Claude yielded the following comment:

> `privileged` should be the default, and the grant scope needs tightening.
>
> The code comment says no stock SCC permits fixed-UID + seccomp — but that skips `nonroot-v2` (4.11+), which is `restricted-v2` without the UID restriction: explicit non-root UIDs **and** `runtime/default` seccomp. argocd-redis (999), Keycloak/Postgres (1001), Envoy are all non-root, so `nonroot-v2` should admit them. Can you rerun your anyuid test with it? If it passes, the default drops from node-root-equivalent to near-restricted.
>
> Independent of the SCC chosen: binding to `system:serviceaccounts:<ns>` grants it to *every* SA in eight namespaces plus any `scc.extra_namespaces` — for packs, that's privilege-escalation-by-config-line. The bindings should target the specific ServiceAccounts that pin UIDs (CRBs to not-yet-existing SAs work fine pre-ArgoCD, same as the group bindings). Longer-term, the cleanest fix is unsetting `runAsUser` in our Helm values where charts allow it — then `restricted-v2` needs no grant at all, like your keycloak-job fix.
>
> Also: existing-mode `Destroy` is a no-op, so these cluster-scoped grants are left behind on the user's cluster after teardown — they should be removed on destroy.
>
> If `privileged` stays available as a fallback, it should be explicit opt-in with a loud warning, never the silent default.

Any thoughts?

PR #434, opened 2026-06-26
Docs Generation

### LINKED ISSUE #440: [Doc] Auto-generate NIC CLI and configuration reference docs
Upstream issue nebari-dev/nebari-docs#665 requires that the `nic` CLI reference and NIC configuration schema reference be auto-generated in `nebari-infrastructure-core` so they cannot drift from the code.

### `nic` CLI reference

- Add a `make docs` target that uses `cobra/doc.GenMarkdownTree` to emit one markdown page per command into `docs/reference/cli/`
- Output should cover: `deploy`, `destroy`, `kubeconfig`, `validate`, `version`
- Delete the hand-written `docs/cli-reference.md`
- Wire into CI — build should fail if generated output is stale

### NIC configuration schema reference

- Add a `make config-docs` tarig structs and rendersper-provider reference pages into `docs/configuration/`
- Output should cover: AWS, Azsting-cluster, Cloudflare DNS,git, core
- Wire into CI — build should tale

### Acceptance criteria

- [ ] CLI reference is generatnd-edited CLI pages)
- [ ] Config schema is generated from Go struct tags (no hand-edited field lists)
- [ ] Both generation targets  automatically

### Reference

- nebari-dev/nebari-docs#665

andrewfulton9:
## Closes
closes #440 
addresses NIC relevant parts of [nebari-docs#665](https://github.com/nebari-dev/nebari-docs/issues/665)

## Description

Replaces the hand-written docs/cli-reference.md with two auto-generated documentation sets that stay in sync with the code:

CLI reference (docs/reference/cli/): A hidden __gendocs subcommand in cmd/nic/gendocs.go uses cobra/doc.GenMarkdownTree to emit one markdown page per command. Invoked via make docs.

Configuration reference (docs/configuration/): A standalone cmd/docgen/ tool uses go/ast to parse provider config structs and render per-provider reference pages (AWS, Azure, GCP, Hetzner, local, existing, Cloudflare, git, core). Invoked via make config-docs. This was ported from upstream/feature/config-doc-generator, with source paths updated for the pkg/providers/ refactor (#405) and coverage extended to Hetzner, existing-cluster, and the new TLS certificate structs.

Both targets are wired into CI with staleness checks — the build fails if the generated output is out of date.

Caveats:
- cobra/doc has no built-in option to suppress its Auto generated by spf13/cobra on <date> footer, so stripCobraDateFooters post-processes the output to keep diffs stable.
- Shell completion commands (nic completion bash etc.) are excluded from the generated docs via DisableDefaultCmd = true during doc generation only — they remain available at runtime.
- Config doc quality depends on doc comments in source structs. Fields with no comment will have an empty Description column.

## How to Test

1. Regenerate CLI docs with `make docs`, then `ls docs/reference/cli/`. Expected: `nic.md`, `nic_deploy.md`, `nic_destroy.md`, `nic_validate.md`, `nic_version.md`, `nic_kubeconfig.md` — no `nic_completion_*.md` files.

2. Regenerate config docs with `make config-docs`, then `ls docs/configuration/`. Expected: `README.md`, `core.md`, `aws.md`, `azure.md`, `gcp.md`, `hetzner.md`, `local.md`, `existing.md`, `cloudflare.md`, `git.md`.

3. Run the test suites with `go test ./cmd/nic/... ./cmd/docgen/...`. Expected: all tests pass.

4. Verify the CI staleness checks pass locally — `make docs && git diff --exit-code docs/reference/cli/` and `make config-docs && git diff --exit-code docs/configuration/` should both exit 0.


cmd/nic/gendocs.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func runGendocs(cmd *cobra.Command, args []string) error {
+	if err := os.MkdirAll(gendocsOutputDir, 0750); err != nil {
+		return fmt.Errorf("create output directory: %w", err)
+	}
+
+	// Remove this command from rootCmd so it is excluded from the generated output.
+	rootCmd.RemoveCommand(cmd)
+	rootCmd.CompletionOptions.DisableDefaultCmd = true
```
dcmcand: **Blocker:** This flag is set too late to do anything. cobra's `ExecuteC` calls `InitDefaultCompletionCmd` before dispatching to any `RunE` (command.go:1113 in v1.10.1), so by the time this line runs the completion command is already attached to `rootCmd`, and `GenMarkdownTree` documents it. A fresh `make docs` on this branch emits five untracked `nic_completion*.md` files, and the committed `nic.md` still links `nic_completion.md` (line 18), a page that doesn't exist in the tree. So the committed docs neither reproduce nor link-check.

Fix: remove the completion command here instead of setting the flag:

```go
for _, c := range rootCmd.Commands() {
    if c.Name() == "completion" {
        rootCmd.RemoveCommand(c)
    }
}
```

then regenerate so `nic.md` loses the dangling link.

.github/workflows/ci.yml
```diff
@@ -64,3 +64,13 @@ jobs:
 
       - name: Test binary
         run: ./nic --help || true
+
+      - name: Verify docs are up-to-date
+        run: |
+          make docs
+          git diff --exit-code docs/reference/cli/
```
dcmcand: **Blocker:** `git diff --exit-code` only reports modifications to tracked files, so this gate misses the two failure modes this branch already exhibits: newly generated pages are untracked and invisible (the regenerated `nic_completion*.md` files pass right through, which is how the completion issue shipped green), and pages for deleted commands or providers linger forever because neither make target cleans its output directory first.

Suggested shape:

```yaml
- name: Verify docs are up-to-date
  run: |
    make docs
    git add -N docs/reference/cli
    git diff --exit-code -- docs/reference/cli
```

(`git add -N` makes untracked files visible to `git diff`; same treatment for the config-docs step.) And have `make docs` / `make config-docs` delete `*.md` in their output dirs before regenerating so removals show up as deletions.

cmd/docgen/main.go
```diff
@@ -0,0 +1,288 @@
...
+
+// configFile represents a source file and the structs to extract from it.
+type configFile struct {
+	path     string
+	structs  []string
+	docTitle string
+	docDesc  string
+}
+
+var configFiles = []configFile{
```
dcmcand, line 61: **Blocker:** This list is a second provider registry, maintained by hand in parallel with `pkg/nic/registry.go`, and nothing fails when it's incomplete: a new provider package, or a new struct in an already-listed file, produces no diff in generated output, so the CI gate passes and the docs are quietly missing things. It has already happened on this branch: `KindConfig` exists in `pkg/providers/cluster/local/config.go` and local.md's table references the type `*KindConfig`, but there's no KindConfig section on the page because it isn't in this list. Readers hit a type name with no definition.

The repo's existing generated-artifact check solves this with discovery instead of a list: the lockfile drift workflow finds every `pkg/providers/cluster/**/templates/main.tf` with `find`. The same idea works here: glob `pkg/providers/{cluster,dns}/*/config.go`, include all exported structs by default (with an exclude list if needed), and fail if a provider directory yields no page. That would also let `generateIndex` derive the README from the discovered set instead of being a third hand-maintained copy of this list (line 247). At minimum, add a test asserting this list covers every directory under `pkg/providers/{cluster,dns}/`.

cmd/docgen/parser.go
```diff
@@ -0,0 +1,236 @@
...
+					doc.IsInline = true
+				}
+			}
+		default:
+			doc.YAMLKey = parts[0]
+		}
+
+		// Check for omitempty to determine required status
+		doc.Required = true
+		for _, opt := range parts {
```
dcmcand: **Blocker:** This loop scans every tag part, including `parts[0]`, which is the key itself. A field whose yaml key is literally `inline` gets `IsInline = true` and is silently dropped by the renderer. This is live in the committed output: `TrustBundleConfig.Inline` (`pkg/providers/cluster/aws/config.go:57`, tag `yaml:"inline,omitempty"`) is missing from the aws.md table even though the section prose right above it says exactly one of Path or Inline must be set.

Fix: iterate `parts[1:]` here; the switch above already consumed `parts[0]`.

cmd/docgen/markdown.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// Write table header
+	g.printf("| Field | YAML Key | Type | Required | Description |\n")
+	g.printf("|-------|----------|------|----------|-------------|\n")
+
+	// Write each field
+	for _, field := range visibleFields {
+		required := "No"
+		if field.Required && !strings.HasPrefix(field.GoType, "*") {
```
dcmcand, line 62: **Blocker:** Deriving Required from the absence of `omitempty` documents serialization behavior, not validation, and it's already wrong in the committed pages: aws.md marks `kubernetes_version` as Required: Yes, but the AWS provider's `Validate` explicitly allows it to be empty (`pkg/providers/cluster/aws/provider.go:164`) and the terraform variable defaults to null (`templates/variables.tf:47`). Either add `omitempty` to tags that are genuinely optional so the tags (and this column) tell the truth, or drop/rename the column. A reference page that says a field is required when it isn't is exactly the drift this PR exists to prevent.

cmd/docgen/main.go
```diff
@@ -0,0 +1,288 @@
+//go:generate go run . -output ../../docs/configuration
```
dcmcand: **Blocker:** This directive writes outside the repository. The `-output` value is joined to the detected project root (line 147), not to the package directory `go generate` runs from, so `go generate ./cmd/docgen` resolves to `<repo>/../../docs/configuration`. Verified with `-verbose` against a scratch root. Since the generated README advertises `go generate ./cmd/docgen` as a regen path, this is a broken documented entry point. `go run .` is enough here; the flag default is already `docs/configuration`.

cmd/docgen/main.go
```diff
@@ -0,0 +1,288 @@
...
+
+	if verbose {
+		log.Printf("Writing %s...", outputPath)
+	}
+
+	f, err := os.Create(filepath.Clean(outputPath))
+	if err != nil {
+		return fmt.Errorf("creating output file: %w", err)
+	}
+	defer func() {
```
dcmcand, line 214: **Blocker:** This deferred assignment can never reach the caller: the function's return value is unnamed, so `return err` at the bottom copies the value before the defer runs. `generateIndex` and `stripCobraDateFooters` get this right with a named `(err error)` return; this function needs the same. As written, with `printf` ignoring write errors by design, a write failure (disk full, say) exits 0 and prints the success message over truncated files.

Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: help build test test-unit test-integration test-coverage test-race clean fmt vet lint install pre-commit release-snapshot
+.PHONY: help build test test-unit test-integration test-coverage test-race clean fmt vet lint install pre-commit release-snapshot localkind-up localkind-down docs config-docs
```
dcmcand: **Stylistic:** `localkind-up` and `localkind-down` no longer exist as targets (removed on main in #400), so these `.PHONY` entries declare nothing. Looks like a leftover from the upstream merge; drop them.

cmd/docgen/parser_test.go
```diff
@@ -0,0 +1,400 @@
...
+		want       string
+	}{
+		{
+			name:       "config package",
+			sourcePath: "pkg/config/config.go",
+			want:       "core.md",
+		},
+		{
+			name:       "aws provider",
+			sourcePath: "pkg/provider/aws/config.go",
```
dcmcand: **Stylistic:** These fixture paths (`pkg/provider/aws/config.go`, `pkg/dnsprovider/cloudflare/config.go`) predate the #405 refactor and no longer exist. The test still passes because `generateOutputName` is pure string manipulation, but the fixtures re-encode dead layout; use the current `pkg/providers/cluster/...` and `pkg/providers/dns/...` paths.

cmd/docgen/markdown.go
```diff
@@ -0,0 +1,145 @@
...
+	g.printf("\n")
+}
+
+func (g *MarkdownGenerator) writeFieldRow(field FieldDoc, required string) {
+	// Escape pipe characters in descriptions
+	desc := strings.ReplaceAll(field.Doc, "|", "\\|")
+	// Replace newlines with spaces
+	desc = strings.ReplaceAll(desc, "\n", " ")
+	// Truncate long descriptions
+	if len(desc) > 200 {
```
dcmcand: **Stylistic:** Byte-slicing at 197 can cut a multibyte rune in half and emit invalid UTF-8 into the markdown. Doc comments are mostly ASCII today so it's latent, but slicing runes (or wrapping with `strings.ToValidUTF8`) is cheap insurance.

cmd/nic/gendocs_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/spf13/cobra/doc"
+)
+
+func TestGendocsGeneratesExpectedFiles(t *testing.T) {
+	tmpDir := t.TempDir()
+
+	rootCmd.RemoveCommand(gendocsCmd)
+	t.Cleanup(func() { rootCmd.AddCommand(gendocsCmd) })
+
+	if err := doc.GenMarkdownTree(rootCmd, tmpDir); err != nil {
```
dcmcand: **Stylistic:** Both tests call `doc.GenMarkdownTree` directly, bypassing `runGendocs`, which is exactly why the completion-command regression was invisible here: cobra only injects the completion command during `Execute`, which never runs in these tests. One end-to-end test that executes the real command (`rootCmd.SetArgs([]string{"__gendocs", "--output-dir", tmpDir})` then `rootCmd.Execute()`) and asserts no `nic_completion*.md` files and no `Auto generated by spf13/cobra` lines would cover both the completion fix and `stripCobraDateFooters`, which currently has no coverage.

cmd/nic/gendocs.go
```diff
@@ -0,0 +1,99 @@
...
+
+	gendocsCmd = &cobra.Command{
+		Use:    "__gendocs",
+		Short:  "Generate markdown reference documentation",
+		Hidden: true,
+		RunE:   runGendocs,
+	}
+)
+
+func init() {
```
dcmcand: **Stylistic:** The five existing commands are all registered centrally in `main.go`'s `init()`; this one registers itself from its own file. Worth following the central pattern so there's one place to see the command tree.

Related, while you're in this file: `runGendocs` is the only handler without a `cmd.<name>` span (even `cmd.version` opens one). If build-tooling commands are meant to be exempt from the tracing convention, a line in AGENTS.md saying so would stop future reviews from flagging it; otherwise the standard span keeps the siblings uniform.

cmd/docgen/cli.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+
+	return stripCobraDateFooters(gendocsOutputDir)
+}
+
+// stripCobraDateFooters removes the "Auto generated by spf13/cobra on <date>"
+// footer line from generated markdown files so the output is stable across runs.
+func stripCobraDateFooters(dir string) (err error) {
+	root, err := os.OpenRoot(dir)
```
dcmcand, line 37: Nice touch using `os.OpenRoot` so the rewrite can't escape the output dir, and stripping the date footer is the right call for reproducible output.

cmd/docgen/parser_test.go
```diff
@@ -0,0 +1,400 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestParseTag(t *testing.T) {
```
dcmcand, line 10: Good table-driven coverage across the tag edge cases here and throughout both test files; matches the project convention nicely.

dcmcand (changes requested):
This is the right direction: generating the CLI reference from the live cobra command tree and the config reference from the source structs means the docs can't drift the way the hand-written page did, and wiring staleness checks into CI is exactly what #440 asked for. The parser and renderer are cleanly separated, docgen stays out of the provider packages by parsing with go/ast (so the layering rules hold), and the parser test coverage is table-driven and thorough. Tests pass, lint is clean, and the config pages regenerate byte-identical on this branch.

The blockers cluster around one theme: the sync guarantee this PR promises doesn't fully hold yet. The completion-command exclusion doesn't actually work (and the committed nic.md links a page that doesn't exist in the tree), the CI gate can't see untracked or orphaned files (which is how the completion issue shipped green), and the config generator has a silent-omission failure mode plus two bugs that put wrong information in the committed pages (`TrustBundleConfig.Inline` missing from aws.md; `kubernetes_version` marked required when it isn't). Details are in the inline comments.

Items not tied to diff lines:

**Blocker: dangling links to the deleted page.** README.md:28, README.md:180, README.md:390, and AGENTS.md:417 all link to `docs/cli-reference.md`, which this PR deletes. Point them at `docs/reference/cli/nic.md` and `docs/configuration/README.md`, and consider adding `cmd/docgen/` and the two generated dirs to the repo map in AGENTS.md.

**Question: relationship to ADR-0005.** That ADR sketches a config-schema pipeline for the same upstream issue (a `cmd/schemagen` emitting JSON Schema + YAML under `schemas/` for nebari-docs to consume). Does `cmd/docgen` supersede it, complement it, or coexist with it? A line in the ADR recording which pipeline is authoritative would prevent two config-reference mechanisms from landing independently.

**Question: two mechanisms for one concern.** CLI docs come from a hidden `__gendocs` subcommand compiled into the production binary, config docs from a standalone tool. Understandable since `rootCmd` is unexported, but did you consider a `newRootCmd()` constructor that a doc tool could consume? That would remove the shipped hidden command and the runtime mutation of the live command tree.

Once the mechanics settle, a short "generated documentation" note in AGENTS.md (what's generated, the regen commands, what to do when adding a provider or command) would make the contract explicit instead of something the next contributor has to discover.

andrewfulton9:
> **Question: relationship to ADR-0005.** That ADR sketches a config-schema pipeline for the same upstream issue (a `cmd/schemagen` emitting JSON Schema + YAML under `schemas/` for nebari-docs to consume). Does `cmd/docgen` supersede it, complement it, or coexist with it? A line in the ADR recording which pipeline is authoritative would prevent two config-reference mechanisms from landing independently.

That's a good point. I guess my question would be is whether or not we expect nebari-docs to ingest config schemas from other packages. If so, then the markdown conversion layer can exist there and we use schemagen here. For what it's worth, Nebari-operator already has a config generator that outputs markdown. If we want all the conversions to live here, it should be straight forward to allow docgen to output in multiple formats. I have no preference as to whether that tool is called "docgen" vs "schemagen". One of the issue requirements is "nebari-docs pulls generated markdown at build time" which suggests that the markdown should be gnerated here

> **Question: two mechanisms for one concern.** CLI docs come from a hidden `__gendocs` subcommand compiled into the production binary, config docs from a standalone tool. Understandable since `rootCmd` is unexported, but did you consider a `newRootCmd()` constructor that a doc tool could consume? That would remove the shipped hidden command and the runtime mutation of the live command tree.

I overlooked that but I think that is better then the hidden command. I'll update that.

dcmcand:
Been chewing on the two-mechanism split here and wanted to float a direction before it lands.

Right now the PR generates docs two ways: CLI reference comes from a hidden `__gendocs` subcommand baked into the `nic` binary, and config reference comes from the standalone `cmd/docgen` tool. Both work, but I think we can collapse them into a single tool that isn't part of the shipped `nic` binary at all.

The one thing blocking that today is that `rootCmd` lives in `package main`, so the only way to reflect over the real command tree is from inside the binary - which is why `__gendocs` has to exist. If we extract command construction into an importable `internal/cli` package exposing `NewRootCmd() *cobra.Command`, then:

- `cmd/nic/main.go` becomes a thin shim: `cli.NewRootCmd().Execute()`
- `cmd/docgen` imports `internal/cli`, builds the same tree, and runs `cobra/doc` itself
- the hidden `__gendocs` subcommand and its `RemoveCommand` handling go away entirely
- as a bonus the CLI becomes unit-testable, which it isn't today

Config docs would stay exactly as they are. `go/ast` parsing is the right tool there, because it reads the doc comments that drive the Description column and reflection can't. So `cmd/docgen` ends up with two generators (CLI via cobra, config via AST) behind one `make docs` target.

Net change from where the PR is now: delete `cmd/nic/gendocs.go`, move the cobra doc call into `cmd/docgen`, and do the `internal/cli` extraction. One tool, off the binary, accurate on both sides.

The cost is the `internal/cli` refactor touching every command's registration, but that's a good change to make regardless.

andrewfulton9:
@dcmcand, I updated the branch based on your last comment. Could you take another look. I think the only think not completely addressed is the schema-gen / doc-gen question from ADR-0005

cmd/docgen/cli.go
```diff
@@ -0,0 +1,86 @@
...
+	"strings"
+
+	"github.com/spf13/cobra/doc"
+
+	"github.com/nebari-dev/nebari-infrastructure-core/internal/cli"
+)
+
+// generateCLIDocs writes one markdown page per nic command into outputDir,
+// built from the same command tree the nic binary runs (internal/cli.NewRootCmd).
+// doc.GenMarkdownTree walks the tree directly without ever calling
```
dcmcand, line 18: **Kudos:** excluding the completion command structurally by never calling `Execute` is cleaner than the `RemoveCommand` approach I'd suggested last round. I checked `GenMarkdownTree` against cobra v1.10.1: it only walks `Commands()`/`IsAvailableCommand()` and never reaches `InitDefaultCompletionCmd`, so there's genuinely nothing to filter out. Nice.

cmd/docgen/cli_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	for _, flag := range []string{"--file", "--dry-run", "--timeout", "--regen-apps"} {
+		if !bytes.Contains(content, []byte(flag)) {
+			t.Errorf("flag %s missing from nic_deploy.md", flag)
+		}
+	}
+}
+
+func TestGenerateCLIDocsExcludesCompletionCommandAndFooter(t *testing.T) {
```
dcmcand, line 51: **Kudos:** this is exactly the end-to-end regression test that was missing before. It runs the real `generateCLIDocs` and asserts no `nic_completion*` pages and no cobra footer, so the completion regression can't quietly come back.

cmd/docgen/main.go
```diff
@@ -0,0 +1,334 @@
...
+		docTitle: "Git Repository Configuration",
+		docDesc:  "Configuration options for GitOps repository integration with ArgoCD.",
+	},
+}
+
+// discoverProviderConfigFiles globs pkg/providers/{cluster,dns}/*/config.go
+// and returns one configFile per match, sorted by path for deterministic
+// output. Every match documents all of its exported structs; there is no way
+// for a provider directory to be silently skipped or partially documented.
+func discoverProviderConfigFiles(rootDir string) ([]configFile, error) {
```
dcmcand, line 90: **Kudos:** glob-based discovery (plus the new-provider test) is a real improvement over the hand-maintained list: a new provider directory gets a page automatically, and `KindConfig` shows up now. Deriving the index from the same set is a nice touch too.

.github/workflows/ci.yml
```diff
@@ -64,3 +64,9 @@ jobs:
 
       - name: Test binary
         run: ./nic --help || true
+
+      - name: Verify docs are up-to-date
+        run: |
+          make docs
+          git add -N docs/reference/cli docs/configuration
```
dcmcand, line 71: **Kudos:** the `git add -N` here plus the `rm -f` in the `make docs` target close both gaps from last round: untracked new pages and orphaned pages for removed commands now both fail the gate.

cmd/docgen/main.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+// configFiles lists the non-provider config files that mix documentation-
+// worthy structs with internal ones (e.g. pkg/config/config.go has
+// ValidateOptions and DNSConfig alongside NebariConfig), so an explicit
+// allowlist is still needed here. Provider config files are discovered
+// separately by discoverProviderConfigFiles.
+var configFiles = []configFile{
+	{
+		path: "pkg/config/config.go",
```
dcmcand, line 63: **Blocker:** this entry lists only `NebariConfig`, `CertificateConfig`, `ACMEConfig`, `ExistingSecretRef`, `CertFiles`, `CertEnv`, and `ParseFile` reads just `pkg/config/config.go`. The main merge moved `TrustBundleConfig` into its own file, `pkg/config/trust_bundle.go`, and added it as a top-level `NebariConfig` field. So `core.md` line 30 references `*TrustBundleConfig` but nothing in `docs/configuration/` defines it: a dangling type reference, which is the drift this tool exists to prevent. The CI gate can't catch it because `make docs` is deterministically incomplete rather than stale (I regenerated and got a clean diff). Quick fix: add `pkg/config/trust_bundle.go` as its own `configFile` entry. More durable: parse every `.go` in `pkg/config/` the way `discoverProviderConfigFiles` walks provider dirs, or fail generation when a referenced `*XxxConfig` type has no generated section.

AGENTS.md
```diff
@@ -102,17 +102,24 @@ OTEL_EXPORTER=otlp OTEL_ENDPOINT=localhost:4317 ./nic deploy -f config.yaml
...
+
+internal/
+  └── cli/              # Cobra command tree; imported by cmd/nic (to run) and cmd/docgen (to introspect for docs)
+      ├── root.go        # NewRootCmd() constructor, Execute(ctx), RunError
+      ├── deploy.go      # Deploy command
+      ├── destroy.go     # Destroy command
+      ├── validate.go    # Validate command
+      ├── kubeconfig.go  # Kubeconfig command
+      ├── version.go     # Version command
+      └── config_discovery.go # Resolve config file path
```
dcmcand, line 119: **Blocker (quick, but worth it in a docs-accuracy PR):** you correctly updated this diagram to show `config_discovery.go` under `internal/cli`, but two prose statements didn't move with it. Line 87 still says config resolution lives in `cmd/nic/config_discovery.go` (that path no longer exists), and lines 324 and 454 still say slog is emitted "only in `cmd/nic`", which isn't true now that the handlers log from `internal/cli`. The code is fine (internal/cli is the application layer now), the doc just needs to say so.

internal/cli/root.go
```diff
@@ -0,0 +1,72 @@
...
+// RunError wraps an error that occurred after PersistentPreRun ran - a
+// genuine runtime failure, as opposed to a cobra usage error (bad flag,
+// unknown command, wrong number of args) that cobra already printed itself.
+type RunError struct{ Err error }
+
+func (e *RunError) Error() string { return e.Err.Error() }
+func (e *RunError) Unwrap() error { return e.Err }
+
+// Execute builds a fresh command tree and runs it against ctx.
+func Execute(ctx context.Context) error {
```
dcmcand, line 66: **Stylistic (non-blocking):** `Execute`'s `RunError` classification is the one genuinely new piece of logic here and it has no test (`root_test.go` only checks command names). It's awkward to test as written because there's no way to inject args, cobra falls back to `os.Args`. Consider `Execute(ctx, args []string)` with `root.SetArgs(args)` so both branches are table-testable (RunE-reached vs a usage error like an unknown flag). Related: `reachedRunE` (line 17) is a package global that's only ever set true, never reset. Harmless today since `Execute` runs once, but the same refactor would let you drop the global.

cmd/docgen/main.go
```diff
@@ -0,0 +1,334 @@
...
+	outPath := filepath.Join(*rootDir, *outputDir)
+	if err := os.MkdirAll(outPath, 0750); err != nil {
+		log.Fatalf("Failed to create output directory: %v", err)
+	}
+
+	providerFiles, err := discoverProviderConfigFiles(*rootDir)
+	if err != nil {
+		log.Fatalf("Failed to discover provider config files: %v", err)
+	}
+	if len(providerFiles) == 0 {
```
dcmcand, line 153: **Stylistic (non-blocking):** this guard only fires when `cluster` and `dns` combined yield zero. If the `dns` glob ever matched nothing while `cluster` still had entries, the whole DNS category would disappear silently with no error, same class as the `TrustBundle` gap above. A per-group check (`len(matches) == 0` inside the loop) would catch it.

cmd/docgen/main.go
```diff
@@ -0,0 +1,334 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("computing relative path for %s: %w", match, err)
+			}
+			relPath = filepath.ToSlash(relPath)
+			providerDir := filepath.ToSlash(filepath.Dir(relPath))
+
+			title, desc := providerDocMeta[providerDir].title, providerDocMeta[providerDir].desc
+			if title == "" {
+				name := filepath.Base(providerDir)
+				title = strings.ToUpper(name[:1]) + name[1:] + " Provider Configuration"
```
dcmcand, line 111: **Stylistic (minor):** `name[:1]` slices by byte, same class as the multibyte truncation you already fixed in `markdown.go`. Zero real-world risk for ASCII package names, just `[]rune(name)` for consistency.

PR #439, opened 2026-06-29
Extract git_repository into repository provider abstractions

### LINKED ISSUE #117: Refactor git_repository to git_provider + git_config pattern
## Description

As part of standardizing the configuration format (#116), the git repository configuration should follow the same `<type>_provider` + `<type>_config` pattern being adopted for cluster and DNS providers.

Currently git uses a one-off pattern:

```yaml
git_repository:
  url: "git@github.com:my-org/my-repo.git"
  branch: main
  path: "clusters/my-nebari"
  auth:
    ssh_key_env: GIT_SSH_PRIVATE_KEY
```

This should become:

```yaml
git_provider: local    # default
git_config:
  url: "git@github.com:my-org/my-repo.git"
  branch: main
  path: "clusters/my-nebari"
  auth:
    ssh_key_env: GIT_SSH_PRIVATE_KEY
```

### Why a provider pattern for git?

The current git logic uses go-git to clone, commit, and push directly from the NIC process. This is effectively a "local" git strategy. But other strategies are plausible:

- **local** (default, current behavior) - NIC uses go-git to clone/commit/push the GitOps repo directly
- **github** - use the GitHub API to create commits/PRs instead of raw git operations
- **gitlab** - use the GitLab API similarly
- **none** - skip GitOps repo management entirely (user manages it externally)

Even if we only implement `local` initially, having the provider pattern in place means adding alternatives later doesn't require a config format change.

### Changes required

**Config struct** (`pkg/config/config.go`):
```go
// Before
GitRepository *git.Config `yaml:"git_repository,omitempty"`

// After
GitProvider string         `yaml:"git_provider,omitempty"`
GitConfig   map[string]any `yaml:"git_config,omitempty"`
```

**Git provider interface** - create `pkg/gitprovider/provider.go`:
```go
type GitProvider interface {
    Name() string
    Push(ctx context.Context, gitConfig map[string]any, manifests map[string][]byte) error
}
```

**Local git provider** - move current go-git logic from `pkg/git/` into `pkg/gitprovider/local/`:
- The existing `git.Config` struct becomes the local provider's config
- The existing clone/commit/push logic becomes the local provider's `Push()` implementation

**Registry** - add a git provider registry in `cmd/nic/main.go`, same pattern as cluster and DNS registries.

**Default behavior** - if `git_provider` is omitted, default to `"local"` (preserving current behavior for existing configs that specify `git_config`).

### Resulting config format

After this change plus #116, all three pluggable subsystems follow the same pattern:

```yaml
project_name: my-nebari
domain: nebari.example.com

cluster_provider: aws
cluster_config:
  region: us-west-2
  kubernetes_version: "1.34"
  node_groups:
    general:
      instance: m5.2xlarge

dns_provider: cloudflare
dns_config:
  zone_name: example.com

git_provider: local
git_config:
  url: "git@github.com:my-org/my-repo.git"
  branch: main
  path: "clusters/my-nebari"
  auth:
    ssh_key_env: GIT_SSH_PRIVATE_KEY

certificate:
  type: letsencrypt
  acme:
    email: admin@example.com
```

### Related

- #116 - Config format inconsistency between cloud provider and DNS provider
- #40 - Generate config documentation from Go structs
- #57 - Configuration documentation generator

marcelovilla:
## Closes

Closes #117

## Description

This PR refactors our git repository logic to have its own place as a new provider type: *repository*. Just like the other two provider types (e.g. cluster and dns), the repository provider has an interface that concrete providers can implement with their own logic. This PR moves the logic we had for existing and local repositories, which was scattered across the codebase and tightly coupled to the git client, to their own repo provider implementations. Furthermore, to avoid that coupling, it refactors the git client a bit to split joint high level functions into independent ones. In particular:

- `Init` was used to clone or initialize a local repository. Now there are two methods: `Init` and `Clone`.
- `CommitAndPush` was used to commit and push files in a single method. Now, there are two methods: `Commit` and `Push`.

This allows to have a clean workflow, where `Clone` and `Push` are only called for remote repositories, while `Init` is called for local repositories and `Commit` for both.

## How to Test

To test this PR, you'll want to make sure both repo providers are working as expected.

### Local repository provider

Deploy a local cluster with a local repository provider:

```yaml
project_name: local-repo-test
domain: nebari.local

certificate:
  type: selfsigned

cluster:
  local:
    kind: {}
    node_selectors:
      general:
        kubernetes.io/os: linux
      user:
        kubernetes.io/os: linux
      worker:
        kubernetes.io/os: linux

# Will use the default path for local gitops
repository:
  local: {}
```

### Existing repository provider

Deploy any cluster with an existing repo provider (make sure to export `GIT_SSH_PRIVATE_KEY` or `GIT_TOKEN`):

```yaml
project_name: existing-repo-test

# your cluster config
cluster:
  ...

# your certificate config
certificate:
  ...

# your dns config:
dns:
  ...

repository:
  existing:
    url: "git@github.com:foo/bar.git"
    branch: main
    path: "clusters/test"
    auth:
      ssh:
        env: GIT_SSH_PRIVATE_KEY
 
      # Alternatively, use a token
      # token:
      #   env: GIT_TOKEN
```


pkg/config/config.go
```diff
@@ -156,6 +158,74 @@ func (c *ClusterConfig) ProviderConfig() map[string]any {
...
+	}
+	return nil
+}
+
+// DefaultLocalRepoPath returns the host directory NIC manages for a project's
+// local GitOps repository when the local repo provider is used without an
+// explicit path.
+func DefaultLocalRepoPath(projectName string) string {
+	return filepath.Join(os.TempDir(), fmt.Sprintf("nebari-gitops-%s", projectName))
+}
```
marcelovilla: Not a huge fan of having this here, but I did not find a better place. The reason we need a shared helper function to get the default local repo path is that the kind provider needs it to create the folder and mount it. The provider is not aware of the `repository` config block and while we could have had the user have to specify the path manually in the `repository` config block and as an `extra_mount` in the `cluster: local` config block, this keeps the original UX where it works out of the box for non-custom paths.

pkg/argocd/writer.go
```diff
@@ -261,15 +260,14 @@ func WriteAll(ctx context.Context, fn func(appName string) (io.WriteCloser, erro
 	return nil
 }
 
-// WriteAllToGit writes all templates (apps and manifests) to the git repository.
+// WriteAllToGit writes all templates (apps and manifests) into workDir.
 // Templates are processed with Go template syntax for dynamic values.
-func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.NebariConfig, gitConfig *git.Config, settings cluster.InfraSettings) error {
+func WriteAllToGit(ctx context.Context, workDir string, cfg *config.NebariConfig, src repository.Source, settings cluster.InfraSettings) error {
```
dcmcand: Narrowing this from a `git.Client` interface to a plain `workDir string` is exactly right: you only needed the one field, and it let the tests drop the `mockGitClient` scaffolding entirely. Nice cleanup.

pkg/providers/repository/provider.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (RemoteSource) isSource()           {}
+func (s RemoteSource) RepoURL() string   { return s.URL }
+func (s RemoteSource) GetBranch() string { return s.Branch }
+func (s RemoteSource) RepoPath() string  { return s.Path }
+
+// ArgoCDAuth returns the credential ArgoCD should use to read the repository,
+// falling back to PushAuth when ReadAuth is not set.
+func (s RemoteSource) ArgoCDAuth() Auth {
```
dcmcand, line 98: The sealed `Source`/`Auth` unions (unexported `isSource()`/`isAuth()`) are a clean way to model this, and they keep the type switches in `deploy.go` and `argocd` safe without a `default`. The `RemoteSource` read-vs-push credential split here is a nice touch.

pkg/config/config.go
```diff
@@ -343,10 +413,11 @@ func (c *NebariConfig) Validate(opts ValidateOptions) error {
 		}
 	}
 
-	if c.GitRepository != nil {
-		if err := c.GitRepository.Validate(); err != nil {
-			return fmt.Errorf("invalid git_repository: %w", err)
-		}
+	if c.Repository == nil {
+		return fmt.Errorf("repository field is required")
```
dcmcand, line 438: **Blocker:** `repository` is required here, but it was optional before this PR: cloud clusters without `git_repository` skipped GitOps bootstrap, and local/kind auto-created a default directory. Every existing config without a `repository:` block now fails validation. Issue #117 lists "default behavior when provider omitted should preserve current behavior" as an acceptance criterion, so this needs either a documented breaking-change/migration note or a way to opt out.

**Question:** Was dropping the "deploy with no GitOps repository at all" mode intentional? A bring-your-own (`existing`) cluster with no `git_repository` previously got ArgoCD installed without a repo bootstrap, which is now rejected outright. If that's a deliberate call (GitOps is always required now, add a `none` provider later), it's reasonable, I just can't tell from the diff whether it was decided or fell out of making `repository` symmetric with `cluster`/`dns`. The answer decides whether the fix here is a migration note or a `none`/optional escape hatch.

pkg/config/config.go
```diff
@@ -29,8 +30,9 @@ type NebariConfig struct {
 	// Only one provider can be configured at a time.
 	DNS *DNSConfig `yaml:"dns,omitempty"`
 
-	// GitRepository configures the GitOps repository for ArgoCD bootstrap (optional)
-	GitRepository *git.Config `yaml:"git_repository,omitempty"`
+	// Repository configures the GitOps repository provider (optional).
```
dcmcand: The doc comment says `(optional)`, but `Validate` (line 417) makes it required. Worth flipping to `(required)` to match, the same way `existing-config.yaml` was updated in this PR.

pkg/git/client.go
```diff
@@ -1,39 +1,426 @@
...
+		return err
+	}
+	_, err := c.stageAndCommit(ctx, message)
+	return err
+}
+
+// Push pushes committed changes to the origin remote configured by Clone, using
+// the credentials captured then. Returns nil when the remote is already up to
+// date, and an error when no remote is configured (e.g. a local repository).
+func (c *Client) Push(ctx context.Context) error {
```
dcmcand, line 300: The doc comment says `Push` "returns...an error when no remote is configured (e.g. a local repository)," but the only guard in the body is `c.repo == nil`. After `Init()` on a local repo, `c.repo` is non-nil, so this wouldn't catch that case; the push would fail deeper inside go-git with a less obvious error. Either soften the comment to say the no-remote error surfaces from go-git, or add an explicit remote/auth check so the function's contract matches its doc.

pkg/nic/deploy.go
```diff
@@ -366,58 +360,52 @@ func (c *Client) lookupEndpointAndProvisionDNS(ctx context.Context, cfg *config.
...
-	}
-
-	// Clone/pull the repository
-	if err := gitClient.Init(ctx); err != nil {
-		span.RecordError(err)
-		return fmt.Errorf("failed to initialize git repository: %w", err)
+	// Acquire the working copy. Local sources are opened in place and remote
+	// sources are authenticated and cloned, then pushed after commit.
+	remote := false
+	switch s := src.(type) {
```
dcmcand, line 474: This type switch on `repository.Source` has no `default`. It's safe today because `Source` is sealed, but a nil `Source` (a caller bug, or a future library consumer calling in directly) would silently skip both branches and leave `gitClient`'s `workDir` empty rather than failing clearly. A `default: return fmt.Errorf(...)` makes that a clear error for free.

pkg/providers/repository/existing/provider.go
```diff
@@ -0,0 +1,126 @@
...
+
+	span.SetAttributes(attribute.String("repository.url", existingCfg.URL))
+
+	pushAuth, err := existingCfg.Auth.resolve()
+	if err != nil {
+		span.RecordError(err)
+		return nil, fmt.Errorf("resolve push credentials: %w", err)
+	}
+
+	readAuth, err := existingCfg.GetArgoCDAuth().resolve()
```
dcmcand: `GetArgoCDAuth().resolve()` runs even when `ArgoCDAuth` is unset, re-reading the same env var `Auth.resolve()` just read for `PushAuth`. It's harmless, but it means `ReadAuth` is never nil coming out of this provider, so `RemoteSource.ArgoCDAuth()`'s documented "nil falls back to PushAuth" branch (`provider.go:95`) is unreachable from the one production caller. Worth knowing when you decide what that branch's tests should look like.

pkg/providers/repository/existing/provider.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/config"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/providers/repository"
+)
+
+// ProviderName is the registry key and config block name for this provider.
+const ProviderName = "existing"
+
+// defaultBranch is used when the config does not specify a branch.
+const defaultBranch = "main"
```
dcmcand, line 19: `"main"` as the default branch is now defined here, in `local/provider.go`, and in `git/client.go`, where it used to be a single `git.DefaultBranch`. All three agree today, so it isn't a bug, but a future change to the default would need to touch three spots.

dcmcand (changes requested):
Thanks for this, Marcelo. It's a clean refactor: the repository provider slots in exactly parallel to cluster and dns, and it removes coupling rather than adding it (config.go no longer imports pkg/git, and the git client no longer knows anything about repository config). I traced the `Init`/`Clone` and `Commit`/`Push` split end to end and it's wired correctly, including the empty-remote bootstrap and the re-deploy idempotency paths. Build, vet, the full race-enabled test suite, and golangci-lint are all green.

I'm marking this request-changes for two reasons. Both are completeness gaps, not design problems:

**1. The three new provider packages have no tests.** `go test -cover` reports 0.0% for `pkg/providers/repository`, `.../local`, and `.../existing`. That logic is the deliverable of this PR, and it's the kind that bites later: the exactly-one-of token/ssh rule, env-var resolution and empty-value rejection, the push/read-auth fallback, and the default-path derivation. Mirroring the existing `TestDNSConfig*` table tests for `RepositoryConfig` would also cover its invalid- and duplicate-provider validation branches, which nothing exercises today (even indirectly, since `ValidateOptions` in the config tests never populates `RepositoryProviders`).

**2. `repository` is now required, which changes behavior for existing configs.** See the inline note at `config.go:417`. Issue #117 lists "default behavior when provider omitted should preserve current behavior" as an acceptance criterion, so this is worth a deliberate decision plus a migration note.

Two general notes, neither blocking:
- `pkg/providers/cluster/local/kind.go` still mounts a default local GitOps directory into every kind cluster regardless of which repository provider is configured. It's pre-existing (only the import path moved in this diff), but a kind cluster paired with a remote `existing` repo now gets a stray empty mount. Might be worth a follow-up issue.
- `repository.Provider.Validate` is never called by any code path (only `Provision` runs, and it re-validates internally). That mirrors the already-unused `cluster.Provider.Validate`, so it isn't something this PR introduced. Just flagging it.

**Docs:** ARCHITECTURE.md has "Cluster Provider System" and "DNS Provider System" sections but no repository one, and ADR-0001 still describes the old flat `git_repository` block and the monolithic clone-or-init / `CommitAndPush` client this PR splits apart. I don't think missing docs should hold up the code, so your call: I'm happy to open a separate docs PR (an ARCHITECTURE section plus an ADR marking 0001 superseded, or a follow-up ADR), or if you'd rather keep it in one place, fold it into this PR. Whichever you prefer.

Specifics are in the inline comments.

pkg/config/config.go
marcelovilla: Having`repository` be required was an intentional change. The repository is what ArgoCD uses for the GitOps approach, and without it, foundational software is not deployed, rendering the platform useless. I'm also not a fan of having a default that works locally (a local git repo) but breaks for any other provider, including `existing`. I think it's better to be explicit in this case, and developers testing locally just need to add a small block to their configs.

I'm not worried about existing configs as long as we properly communicate the change. We still don't have a stable config schema and our user base is small enough that this breaking change should not have a big impact.

Where would you document this? I think release notes would probably be the best place at this time.

pkg/config/config.go
marcelovilla: Good catch. Addressed in 4b0ca55039531c44d582f7a2d5b375b7607d9112

pkg/git/client.go
marcelovilla: Good catch. Addressed in a9b585fd353527e4df908d6d3f53ac2ef56352b0

pkg/nic/deploy.go
marcelovilla: Addressed in 39f3c2fa36903f38c20dc25fefff7f6231ea35a5

pkg/providers/repository/existing/provider.go
marcelovilla: Good catch. I think it's cleaner if it runs only when `ArgoCAuthD` is set. Addressed in https://github.com/nebari-dev/nebari-infrastructure-core/pull/439/commits/2f2da5b0fec6b500ac63a7cc6832963f982c9e7a

pkg/providers/repository/existing/provider.go
marcelovilla: This is intentional. I considered having it in the `git` package, like it was before, but decided against it because we're planning to have out-of-tree providers and that means providers need to either import `pkg/git` just to get that constant (pulling extra unneeded deps along the way), or need to redefine theirs. That's exactly why it's defined in three places now. Do you have any strong feelings? Would you prefer seeing this in a single place until we move the providers out-of-tree? Even in the same codebase, I think it's good to have the git client and the repository providers be isolated.

marcelovilla:
Thanks for the review @dcmcand.

I addressed your inline comments (see my replies to each of them) and I'll address the general ones here:

> The three new provider packages have no tests

I had Claude generate tests for them. See 8c8be90795c63788a6631f6d3404c517082083ca

> Repository is now required, which changes behavior for existing configs

I answered this here: https://github.com/nebari-dev/nebari-infrastructure-core/pull/439#discussion_r3558123959

> `pkg/providers/cluster/local/kind.go` still mounts a default local GitOps directory into every kind cluster regardless of which repository provider is configured.

This is a good call. I'll wait to see where https://github.com/nebari-dev/nebari-infrastructure-core/pull/448 lands and then open a follow-up issue for it.

> `repository.Provider.Validate` is never called by any code path (only Provision runs, and it re-validates internally).

Another good call. I want to fix this as part of https://github.com/nebari-dev/nebari-infrastructure-core/issues/72

> Docs: ARCHITECTURE.md has "Cluster Provider System" and "DNS Provider System" sections but no repository one, and ADR-0001 still describes the old flat git_repository block and the monolithic clone-or-init / CommitAndPush client this PR splits apart.

I just updated ARCHITECTURE.md in 4c22814d3bc25f0221257c90ccd73269fdacaef6 and can open a follow-up PR with an ADR that superseeds ADR-0001.

ARCHITECTURE.md
```diff
@@ -247,6 +250,53 @@ func (p *Provider) someFunction(ctx context.Context, clients *Clients, cfg *conf
 
 **Known limitation:** Changing the `domain` in config and redeploying creates new records but does not clean up the old domain's records. See [DNS Provider Architecture](docs/design-doc/implementation/09-dns-provider-architecture.md#orphaned-records-on-domain-change).
 
+## Repository Provider System (pkg/providers/repository/)
+
+**Location:** `pkg/providers/repository/`
+
+**Purpose:** GitOps repository provider abstraction. Providers resolve (or create) the repository that ArgoCD syncs foundational software from, and return a typed `Source` describing how to reach it. The `repository:` config block is required and follows the same provider-name-as-key pattern as `cluster:` and `dns:`. See [ADR-0007](docs/adr/0007-repository-provider-abstraction.md) for the design rationale.
```
dcmcand: Stylistic (non-blocking): this links to `docs/adr/0007-repository-provider-abstraction.md`, which doesn't exist yet (the superseding ADR is the promised follow-up PR). Until that lands this is a dangling link. Either land ADR-0007 alongside this PR or soften the reference so we don't ship a broken link. (Unrelated: the pre-existing `docs/architecture/06-stateless-operation.md` link elsewhere in this file is also dead, if you want to sweep it while you're here.)

marcelovilla:
I'm waiting on https://github.com/nebari-dev/nebari-infrastructure-core/pull/448 to be merged before merging this PR. The reason is that #448 touches several parts that this PR refactored, so I'll need to update it once that lands so @oldsj does not have to deal with that on his PR.

dcmcand:
@marcelovilla #448 is now merged, so you can unblock this pr

marcelovilla:
Even though this PR is (almost) ready for a re-review (need to fix new conflicts), I'm marking it as blocked until https://github.com/nebari-dev/nebari-infrastructure-core/pull/446 lands and we can run automated deployment tests for this.

I've tested this already manually on different providers but it's a nice opportunity to have the tests run for a big refactor.

dcmcand (dismissed):
Re-reviewed at `ca93109`. Since my last pass the delta is two merges of main plus a dependency bump, so I focused on merge integrity and on what the conflict resolutions had to hand-write.

**The merge work is the hard part of this PR and it came through clean.** Two merges spanning about 35 main commits, including four modify/delete conflicts where `pkg/git/client_impl.go`, `client_impl_test.go`, `config.go` and `config_test.go` were deleted on this branch while main was actively changing them. That is exactly the shape where upstream fixes disappear silently. I checked and they did not: #379's host-key verification is correctly ported into `pkg/git/auth.go`, #448's permission repair into `pkg/git/local.go`, and #499, #473, #481, #535, #443/#509/#523/#527, #489, #504 and #518 are all intact. The changed-file set is exactly 64/64 against the merge base, so nothing drifted in outside the diff. `pkg/git/dependency_floors_test.go` survived the package rewrite. Build, vet, race-enabled suite and golangci-lint are all green locally, and CI is green.

Everything from my previous two rounds is still resolved. The `default` arm and nil guard in `deploy.go`, the `ArgoCDAuth` fix, the `Push` doc comment, and the provider test coverage all held up through the merges.

One blocker and a batch of documentation fixes.

### Blocker: `nic validate` no longer catches malformed repository config, and `--dry-run` doesn't either

The per-provider `Config.Validate()` methods are correct, but nothing on the validate path calls them. `NebariConfig.Validate` reaches only `RepositoryConfig.Validate` (`pkg/config/config.go:185-200`), which checks exactly one provider key, a registered name, and a mapping value. It never dispatches to the provider. `existing.Provider.Validate` and `local.Provider.Validate` have no production callers at all: removing `Validate` from the `repository.Provider` interface and renaming both implementations still compiles the entire production tree, and only the two providers' own test files break.

So the real checks are first reached inside `Provision`, at `deploy.go:187`, which runs after `clusterProvider.Deploy` has already returned at line 166.

Five checks that main performs today are lost:

| Config defect | main `nic validate` | this branch |
|---|---|---|
| `existing` with no `url` | exit 1 | **exit 0, "valid"** |
| `existing` with both `token` and `ssh` | exit 1 | **exit 0, "valid"** |
| `existing` with no `auth` block | exit 1 | **exit 0, "valid"** |
| `local` with a relative `path` | exit 1 | **exit 0, "valid"** |
| `file://` path resolving to a non-directory | exit 1 | no equivalent check at any stage |

`--dry-run` is not a fallback: `deploy.go:186` gates the whole repository branch behind `if !opts.DryRun`, so a config with no URL and both auth methods set produces "Deployment completed successfully" and exit 0. On main the same config fails dry-run with exit 1, because the check lives in `cfg.Validate`.

Two things make this worse than the equivalent gap elsewhere. First, `repository:` is now mandatory, so the population carrying an unvalidated block goes from "users who opted in" to everyone. Second, on aws, azure and hetzner the failure now surfaces after `tofu apply` has created real infrastructure.

The fix pattern already exists in the same file: `validate.go:31` calls `validateDNSProvider` "so misconfigurations surface before any infrastructure is provisioned", and `validate.go:44` consults `InfraSettings` for the Longhorn check "at validate time rather than mid-deploy". Both repository providers' `Validate` methods are side-effect free (the only `os.Getenv` calls in the tree are in `AuthConfig.resolve()`, reached only from `Provision`), so they are safe to call offline. Roughly ten lines mirroring `validateDNSProvider` restores all of it.

I raised the unreachable `Validate` last round as a non-blocking note and accepted deferring it to #72. That was my error: I framed it as an unused method mirroring `cluster.Provider.Validate`, and missed that on main these checks are reachable through `config.Validate`, which makes it a regression rather than a pre-existing gap. I do think the general fix (wiring provider `Validate` into `Client.Validate` for all three categories) is the right shape and belongs in #72, but the repository half should land here since it is coverage this PR removes.

While confirming this I also noticed the local/cloud-cluster incompatibility check at `deploy.go:195` runs after `Deploy`. That one is not a regression, main has no such check at all, so it is a genuine improvement. It would just be worth more at validate time, and it comes free with the same fix.

### Documentation

- `docs/adr/0007-repository-provider-abstraction.md` collides with `docs/adr/0007-cloudnativepg-managed-databases.md`, already on main and already indexed. Every reference to "ADR-0007" is now ambiguous. Next free number is 0015. (The repo hit this before with the duplicate 0005, renumbered in #483.)
- The new ADR is not in `docs/adr/README.md`, whose own step 4 is "Update the index table above".
- ADR-0007:7 declares "Supersedes ADR-0001", but ADR-0001 still reads `Proposed` with no back-link, `README.md:13` still lists it as `Proposed`, and its body at line 97 still documents the `git_repository:` block this PR deletes. `README.md:30` defines `Superseded` as a status requiring a link to the replacement.
- ADR-0007 states the wrong default path twice. Line 58 says "defaults to a per-project temp dir" and line 70 repeats it, but `config.go:237-242` returns `~/.nic/gitops/<project_name>` with the temp dir only as a fallback when the home directory cannot be resolved. The other three places that document this are correct, so the ADR is the outlier.
- `AGENTS.md:13` still says "two provider categories in tree" and the `Registry` snippet at lines 188-192 shows two fields where `pkg/registry/registry.go` now has three.

### Suggestions, none blocking

- `pkg/providers/repository/provider.go:11-12` and ADR-0007:83 tell implementations they "must not depend on pkg/git or go-git", but `local/provider.go` does exactly that (and reasonably so, a repository provider using go-git is fine). Worth rewording now, while the contract is still being written, so it doesn't misdirect the first out-of-tree provider author.
- ADR-0007:113 says "sealed kinds keep dispatch exhaustive", but Go doesn't check this and `exhaustive` isn't enabled in `.golangci.yml`. `install.go:128` and `:190` silently skip the hostPath mount and the credentials Secret on an unknown kind, and `gitAuth`'s `default` at `deploy.go:339` silently returns anonymous auth. Unreachable today with two kinds; a third would produce a cluster with no credentials and no error. Either give those three sites erroring defaults, as `bootstrapGitOps` already does, or soften the claim.
- AGENTS.md has "Adding a New Cluster Provider" and "Adding a New DNS Provider" recipes but no repository one, and the pre-commit checklist item 7 still scopes the boundary rule to cluster and dns.
- `local.ResolveDir` is at 0% coverage; it is the whole of the package's drop from 83.3% to 76.9%.
- The kind-mount follow-up issue never got filed. #448 merged on 2026-07-17, and `kind.go:77-86` still creates and mounts `~/.nic/gitops/<project>` even when the repository provider is `existing`.
- `pkg/nic/registry.go` registers cluster and DNS providers with string literals but repository providers with `ProviderName` consts. The const form is better and the consts already exist for `cluster/aws`, `cluster/local` and `cluster/existing`.
- This will want one more merge of main before it lands: #550 added `pkg/nic/destroy_test.go`, which did not exist at your merge base and which tests `reportRetainedGitOpsDir`, the function this PR rewrites. A clean textual merge won't surface that.
- The breaking change still has no release-notes home, and there is no `CHANGELOG.md`. ADR-0007:88 records it, which helps. A user upgrading will hit `repository field is required` with no mention of `git_repository`; naming the old key in that error would be a cheap migration aid.

Nice work on the merges, and on the custom-path documentation: the `extra_mounts` requirement for a non-default `repository.local.path` is the sharp edge here and you documented it in three separate places.

pkg/nic/deploy.go
```diff
@@ -554,34 +578,19 @@ func (c *Client) writeConfigToRepo(ctx context.Context, cfg *config.NebariConfig
...
-// configured bundle is rewritten to its resolved inline form; this also keeps
-// the committed config self-contained and reflecting the deployed value.
-func scrubbedConfig(cfg *config.NebariConfig, gitConfig *git.Config, trustBundlePEM string) *config.NebariConfig {
+// committedConfig returns the config as it should be recorded in the GitOps
+// repo. A path:-based trust_bundle references a file on the operator's
+// machine, which is meaningless (and leaks a local path) in the committed
+// record, so any configured bundle is rewritten to its resolved inline form
+// (trustBundlePEM, already resolved by the caller; empty when unset). This
+// keeps the committed config self-contained and reflecting the deployed value.
+func committedConfig(cfg *config.NebariConfig, trustBundlePEM string) *config.NebariConfig {
```
oldsj, line 592: **Blocker**, though not for the reason it first looks like — I'll argue myself down on the disclosure and up on the process.

`main`'s `scrubbedConfig` deliberately zeroed the auth block before writing `nic-config.yaml`:

```go
gitRepo.Auth = git.AuthConfig{} // value type with no omitempty: zero the env-var names
gitRepo.ArgoCDAuth = nil        // *AuthConfig with omitempty: nil → omitted
```

`committedConfig` only rewrites the trust bundle, so the `repository:` block now goes in verbatim. Probe test at this head, using `main`'s own forbidden-strings list:

```yaml
repository:
    existing:
        argocd_auth:
            token:
                env: ARGOCD_TOKEN
        auth:
            ssh:
                env: MY_SSH_KEY
        url: git@github.com:org/repo.git
```

**On the disclosure itself: I think committing the names is fine, and arguably better than what `main` did.** Names are identifiers, not secrets; `GIT_SSH_PRIVATE_KEY` and `GIT_TOKEN` already ship in `examples/` so the realistic leak is one bit from a set of two well-known strings; SSH-vs-token is already inferable from the `url:` scheme, which neither branch strips; and the same file deliberately inlines the resolved trust-bundle PEM (`deploy.go:595-596`), a far larger disclosure that both branches accepted. `main`'s scrub also produced an actively *wrong* record — `auth: {ssh_key_env: "", token_env: ""}`, asserting empty auth with nothing marking it redacted — which is worse for a file whose stated purpose is to reflect the deployed value. So: no objection to the new behaviour.

**What I'd like fixed before merge is that the change is invisible.** A tested, commented control was removed and nothing records it:

1. `main` had **9 auth-specific assertions**, including `// Sensitive env-var names and the ArgoCDAuth block must not appear.` over `{"MY_SSH_KEY", "MY_TOKEN", "ARGOCD_TOKEN", "argocd_auth"}`. They were deleted, not inverted — `TestCommittedConfig` at head has four subtests, all trust-bundle. **Nothing now pins either behaviour**, in either direction.
2. `docs/design-doc/implementation/10-foundational-software.md:51` still labels the file `# Scrubbed copy of nebari-config.yaml`. That is now false, and it's the last line in the tree describing `main`'s behaviour.
3. The operator-visible signal disappeared: `main` sent `"Wrote NIC config to repository (auth fields scrubbed)"`, head sends `"Wrote NIC config to repository"`.
4. No artifact a reviewer reads — the comment at `deploy.go:563`, ADR-0015's Consequences, or the PR body — mentions that a prior deliberate decision was reversed. ADR-0015:85 is the near miss: it lands under "Consequences → Good" and reads as a security *improvement*.

Roughly a ten-line fix: invert `TestCommittedConfig` to assert the names **are** present with a comment saying why that's acceptable (so the next person can't silently flip it back either way), fix the `10-foundational-software.md:51` label, and add a line to ADR-0015 recording the reconsideration.

<sub>🤖 claude-opus-5 (xhigh) · `review-pr` skill · reviewed by @oldsj</sub>

ARCHITECTURE.md
```diff
@@ -247,6 +250,53 @@ func (p *Provider) someFunction(ctx context.Context, clients *Clients, cfg *conf
...
+
+### Local Provider (pkg/providers/repository/local/)
+
+**Location:** `pkg/providers/repository/local/`
+
+Provisions a directory on disk as a `LocalSource` - the zero-dependency, no-network option for local/dev clusters. NIC commits to it in place and ArgoCD's repo-server reads it via a hostPath mount, so it requires a cluster provider with `SupportsLocalGitOps` (e.g. kind); `pkg/nic` rejects incompatible pairings before the GitOps bootstrap runs.
+
+| File | Purpose |
+|------|---------|
+| `provider.go` | `Provision` creates the directory (default: per-project dir under the OS temp dir) and returns a `LocalSource` |
```
oldsj: **Blocker (docs):** wrong default, and this line is new in this PR.

`config.DefaultLocalRepositoryPath` (`pkg/config/config.go:237-243`) returns `~/.nic/gitops/<project_name>`; the OS temp dir is only the fallback for when the home directory can't be resolved.

This matters more than a typo because the two locations differ in durability — a reader who believes their GitOps repo lives under `$TMPDIR` may reasonably assume it's disposable.

Worth noting the same claim was caught in the ADR last round and fixed there, so this is the last surviving copy. The other four places that document it are correct (`docs/adr/0015-repository-provider-abstraction.md:70`, `pkg/providers/repository/local/config.go:10-13`, `docs/design-doc/appendix/16-configuration-reference.md:562`, `docs/design-doc/implementation/07-configuration-design.md:139`).

<sub>🤖 claude-opus-5 (xhigh) · `review-pr` skill · reviewed by @oldsj</sub>

ARCHITECTURE.md
```diff
@@ -49,7 +52,7 @@ nebari-infrastructure-core/
 
 ### Key Patterns:
 
-- **Provider Registration:** `main.go` explicitly registers all cloud and DNS providers using registry pattern
+- **Provider Registration:** `main.go` explicitly registers all cloud, DNS, and repository providers using registry pattern
```
oldsj: Stylistic: this PR edited the line to add "repository" but the `main.go` claim was already false and stays false — registration happens in `defaultRegistry` at `pkg/nic/registry.go:53`. `grep -rn Register cmd/nic/` finds no registration calls at all, only the two `Printf`s in `version.go`.

Same stale claim appears at `ARCHITECTURE.md:467` and `:546-547`. Not introduced here, but this is the line the PR touched, so it's the cheap moment to fix it.

<sub>🤖 claude-opus-5 (xhigh) · `review-pr` skill · reviewed by @oldsj</sub>

pkg/nic/validate.go
```diff
@@ -68,3 +81,35 @@ func validateDNSProvider(ctx context.Context, cfg *config.NebariConfig, reg *reg
...
+	return nil
+}
+
+// ensureLocalRepositorySupported rejects the local repository provider on a
+// cluster that cannot host its directory (only local/kind clusters can mount
+// one). The name-based check catches this at validate time; deploy re-checks
+// the provisioned source kind, which also covers out-of-tree providers that
+// return a LocalSource.
+func ensureLocalRepositorySupported(cfg *config.NebariConfig, supportsLocalGitOps bool) error {
+	if cfg.Repository != nil && cfg.Repository.ProviderName() == repositorylocal.ProviderName && !supportsLocalGitOps {
```
oldsj, line 111: Question: this is a provider-name switch in `pkg/nic`, which is what the checklist item this PR widened forbids — `AGENTS.md:488`: "no provider-name switches outside `pkg/providers/cluster/`, `pkg/providers/dns/`, or `pkg/providers/repository/`". `pkg/nic` also imports the concrete provider package to do it (`validate.go:11`).

I'm raising this as a Question rather than a Blocker because there's no wrong outcome today: with only `local` and `existing` in tree the check is exactly right, and your comment above it already concedes the out-of-tree gap and points at the deploy-time source-kind re-check at `deploy.go:214` as the backstop. So the behaviour is fine and documented — it's the precedent I'd like to settle, since this PR is also the thing that widened the rule to cover repository providers.

The codebase already has the shape that would avoid it: `backupRoleARNResolver` at `deploy.go:606-624` type-asserts an optional interface on the registry-resolved provider instead of comparing names. An optional `interface{ ProvisionsLocalSource() bool }` would do the same job here.

Happy for the answer to be "deliberate, and the general fix belongs in #72" — that's consistent with how the `Validate`-wiring discussion landed. Mostly I want it to be a recorded choice rather than drift.

<sub>🤖 claude-opus-5 (xhigh) · `review-pr` skill · reviewed by @oldsj</sub>

pkg/nic/destroy.go
```diff
@@ -182,29 +183,19 @@ func (c *Client) Destroy(ctx context.Context, cfg *config.NebariConfig, opts Des
...
-		// Fall back to the same auto-generated local config deploy uses, but
-		// only for providers that manage a local GitOps directory.
-		if !clusterProvider.InfraSettings(cfg.Cluster).SupportsLocalGitOps {
-			return
-		}
-		gitConfig = defaultGitConfig(cfg.ProjectName)
-	}
-
-	if !gitConfig.IsLocalPath() {
+	if cfg.Repository == nil || cfg.Repository.ProviderName() != repositorylocal.ProviderName {
```
oldsj, line 216: Same as the `validate.go:111` note — name switch plus a concrete call into the provider package (`repositorylocal.ResolveDir` at `:198`).

Lower stakes than the validate one: the consequence of an out-of-tree local-ish provider falling through here is a missing "directory was retained" reminder, which is cosmetic. An optional `interface{ RetainedDir(ctx, project, cfg) (string, bool) }` would cover both this and the validate site.

<sub>🤖 claude-opus-5 (xhigh) · `review-pr` skill · reviewed by @oldsj</sub>

pkg/providers/repository/existing/config.go
```diff
@@ -0,0 +1,115 @@
...
+	InsecureSkipHostKeyVerification bool `yaml:"insecure_skip_host_key_verification,omitempty" json:"insecure_skip_host_key_verification,omitempty"`
+}
+
+// EnvRef names the environment variable a secret is read from.
+type EnvRef struct {
+	Env string `yaml:"env" json:"env"`
+}
+
+// Validate checks that the configuration is well-formed.
+func (c *Config) Validate() error {
```
oldsj, line 64: Stylistic, and explicitly **not** a regression — `main` behaved identically. But this PR creates the natural home for the fix, so flagging it here.

Nothing validates `Path`, and a `..` component escapes the clone. `pkg/git/client.go:167` does `c.workDir = filepath.Join(tempDir, c.path)`, and `Join` cleans, so:

```
path: ""                -> <tempDir>            (fine)
path: "/"               -> <tempDir>            (fine)
path: "clusters/x"      -> <tempDir>/clusters/x (fine)
path: "../.."           -> /
path: "../../../etc"    -> /etc
```

With an escaped `workDir`, manifests and `nic-config.yaml` are written outside the clone, and `removeStaleTemplate`'s `os.RemoveAll` at `pkg/argocd/writer.go:482` operates on escaped paths. Self-inflicted — the operator has to write a hostile path into their own config — so it's a footgun rather than a vulnerability.

Now that this `Validate()` actually runs at validate time, rejecting it is about three lines:

```go
if c.Path != "" {
    if filepath.IsAbs(c.Path) {
        return fmt.Errorf("path must be relative to the repository root, got: %s", c.Path)
    }
    if cleaned := filepath.Clean(c.Path); cleaned == ".." || strings.HasPrefix(cleaned, "../") {
        return fmt.Errorf("path must not escape the repository root, got: %s", c.Path)
    }
}
```

<sub>🤖 claude-opus-5 (xhigh) · `review-pr` skill · reviewed by @oldsj</sub>

pkg/nic/validate.go
```diff
@@ -68,3 +81,35 @@ func validateDNSProvider(ctx context.Context, cfg *config.NebariConfig, reg *reg
...
+	}
+	return nil
+}
+
+// ensureLocalRepositorySupported rejects the local repository provider on a
+// cluster that cannot host its directory (only local/kind clusters can mount
+// one). The name-based check catches this at validate time; deploy re-checks
+// the provisioned source kind, which also covers out-of-tree providers that
+// return a LocalSource.
+func ensureLocalRepositorySupported(cfg *config.NebariConfig, supportsLocalGitOps bool) error {
```
oldsj, line 110: Minor, and possibly intentional: this checks local-repo-on-cloud-cluster, but not the reverse. `repository.existing` on a `local`/kind cluster validates clean. That's probably correct — a kind cluster can reach a remote repo — but worth confirming it's deliberate rather than an unconsidered asymmetry, since the function name reads as though it covers pairing in general.

<sub>🤖 claude-opus-5 (xhigh) · `review-pr` skill · reviewed by @oldsj</sub>

marcelovilla:
@dcmcand / @oldsj can I get a final review here? Tests with both [a local repository (on a local cluster)](https://github.com/nebari-dev/nebari-infrastructure-core/actions/runs/32031367696/job/95392648036) and with [a remote repository (on a Hetzner cluster)](https://github.com/nebari-dev/nebari-infrastructure-core/actions/runs/32031447219/job/95392870672) are passing.

oldsj (approved):
Re-reviewed at `4723eae`. Everything from my last round is resolved, and the two behavioural fixes I checked by execution rather than by reading:

- The committed-config reversal is now recorded rather than silent — `TestCommittedConfig` asserts the env-var names are present, and I mutation-checked it (nulling `out.Repository` in `committedConfig` turns exactly that subtest red). The `10-foundational-software.md` label and ADR-0015's Consequences both match the behaviour now.
- Path escaping is rejected at validate time: `path: "../.."` and `path: "/etc"` both exit 1 on `nic validate` and `nic deploy --dry-run`, `clusters/../other` correctly still passes, and all 9 examples plus the 5 CI fixtures validate clean.
- `nic version` prints the repository category, and that line survived main's `version.go` rewrite in the merge.
- The docs fixes landed, [#585](https://github.com/nebari-dev/nebari-infrastructure-core/issues/585) captures the bootstrap-marker ordering with a better fix shape than the one I sketched, and [#117](https://github.com/nebari-dev/nebari-infrastructure-core/issues/117) now explains the shape deviation.

Your calls to defer the migration-error wording, the release-notes home, and the provider-name switches to the plugin-contract work are all reasonable and consistently argued — no objection from me.

One leftover, not worth another round: `ARCHITECTURE.md:46` still says `main.go` "Registers providers". It's the last copy of the claim you fixed four lines below at `:55`, and it predates this PR — fold it in if you touch the file again.

Verified at `4723eae`: build, vet, `go test -short ./...`, and `golangci-lint` (0 issues) all green; CI green. Not run: a real cloud deploy — your Hetzner and local deployment-test runs cover that.

@dcmcand — your `CHANGES_REQUESTED` from `ca93109` is still formally standing and is what's holding the merge (`mergeStateStatus: BLOCKED`). The blocker behind it (unreachable repository `Validate`) has been fixed since `0b79196`; worth dismissing when you get a moment.

<sub>🤖 claude-opus-5 (medium) · `review-pr` skill · reviewed by @oldsj</sub>

PR #443, opened 2026-07-06
feat(backups): native Longhorn backups (#431)

tylerpotts:
## Summary

Brings Longhorn backup functionality natively into NIC and deprecates the standalone [`nebari-longhorn-backup-pack`](https://github.com/nebari-dev/nebari-longhorn-backup-pack) chart. Longhorn snapshot/backup scheduling and the S3/azblob `BackupTarget` are now a first-class, config-driven part of `nic deploy`.

Closes https://github.com/nebari-dev/nebari-infrastructure-core/issues/431

### What's included
- **Config:** new top-level `backups.longhorn` block (S3 or azblob target, schedules, retention, `allow_recurring_job_while_volume_detached`, `retain_on_destroy`), validated at `nic validate`.
- **Credential Secret:** built imperatively by NIC from env-var-sourced keys (never the config file) + optional CA from a referenced Secret/ConfigMap (`AWS_CERT`), applied to `longhorn-system` during foundational bootstrap.
- **GitOps:** a foundational ArgoCD Application renders `BackupTarget/default`, the two `RecurringJob`s, and the detached-volume `Setting` from config; skipped entirely when backups are disabled.
- **Bucket provisioning (in the provider modules):** the backup S3 bucket / storage account + container now live in the provider Terraform modules — **companion PRs** nebari-dev/terraform-aws-eks-cluster#39 and nebari-dev/terraform-azurerm-aks-cluster#4 — behind optional `longhorn_backup_*` variables. NIC's shim stays a pure module caller: it passes `create`/`name`/`force_destroy` into the module and reads the module's backup output.
- **`retain_on_destroy` (default true):** `nic destroy` preserves backups by removing the backup resources from Terraform state before destroy (shared helper + signal-safe `Show`/`StateRm` wrappers); `false` deletes them with the cluster.
- **Third-party S3:** `endpoint` → `AWS_ENDPOINTS`, `virtual_hosted_style` → `VIRTUAL_HOSTED_STYLE`, `ca_cert` → `AWS_CERT`.
- **Safety guard:** backups require the cluster's storage layer to be Longhorn; enabling them where Longhorn isn't installed (Azure, or AWS with Longhorn disabled) fails cleanly at `nic validate`/`nic deploy`.
- **Docs:** `docs/longhorn-backups.md` + `examples/longhorn-backups-config.yaml`.

### ⚠️ Temporary git-ref module source — must flip before merge
The two shim `main.tf` files temporarily point their module `source` at the companion feature **branches** (`git::…?ref=…`) because the backup variables aren't in a released module version yet. Once the companion PRs merge and releases are cut, revert each `source` to the registry form + a pinned `version` (there's a `TODO(#431)` at each site). **This PR should not merge with the git-ref source in place.**

### Scope notes
- **Azure azblob is dormant scaffolding.** NIC does not install Longhorn on Azure (`managed-csi`), so the azblob path is guarded off until a separate "Longhorn on Azure" prerequisite lands.
- **GCP/local** have no Longhorn/TF module and are excluded from cloud-native bucket creation (external `endpoint` only).

## Test plan
- [x] `make check` (fmt, vet, golangci-lint, unit tests) passes
- [x] Race tests pass across affected packages
- [x] `tofu validate` passes on both shims after `init` fetches the git-branch modules (verifies the cross-repo variable/output wiring)
- [x] `examples/longhorn-backups-config.yaml` parses and validates
- [x] MinIO integration test (`-tags=integration`): mapping assertions run; live round-trip documented + skip-gated
- [ ] Companion module PRs merged + released; shim `source` flipped back to pinned `version`
- [x] Live AWS deploy (keyless Pod Identity): RecurringJobs + `BackupTarget` healthy (`available=true`), backups `Completed` and objects land in S3
- [ ] Verify `nic destroy` with `retain_on_destroy: true` leaves the bucket + backups intact

## TODO (merge order & follow-ups)

**Blocking — before this PR merges:**
- [ ] Merge https://github.com/nebari-dev/terraform-aws-eks-cluster/pull/39 (optional S3 bucket + Longhorn Pod Identity) and cut a release
- [ ] Merge https://github.com/nebari-dev/terraform-azurerm-aks-cluster/pull/4 (optional storage account + container) and cut a release
- [ ] Flip both shim `main.tf` `source` from the `git::…?ref=…` branch ref back to the registry form + a pinned `version` (remove the `TODO(#431)` markers at each site)

**Follow-ups — out of band:**
- [ ] Archive https://github.com/nebari-dev/nebari-longhorn-backup-pack and point its README at NIC's docs
- [ ] Implement Longhorn-on-Azure to activate the azblob backup path (currently dormant scaffolding)


tylerpotts:
## How to validate Longhorn backups are working

Verified end-to-end on an EKS cluster with keyless (Pod Identity) auth. `$KUBECONFIG` points at the target cluster; `<bucket>` is `backups.longhorn.s3.bucket` from your config.

### 1. Auth verdict — the single most important check

```bash
kubectl -n longhorn-system get backuptarget default \
  -o jsonpath='url={.spec.backupTargetURL}{"\n"}available={.status.available}{"\n"}{.status.conditions[0].message}{"\n"}'
```

- `available=true` ⇒ Longhorn made a real authenticated `ListBucket` against the target. Done.
- `available=false` ⇒ read the message. Two common causes:
  - `missing [AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY] in the secret` — the credential secret lacks both static keys **and** `AWS_IAM_ROLE_ARN`. On keyless AWS this means the role ARN wasn't written (see step 2).
  - `AccessDenied ... s3:ListBucket on resource "arn:aws:s3:::<name>"` — the role is assumed but the **BackupTarget URL and the IAM policy point at different buckets**. Almost always config drift: the bucket was renamed but the GitOps repo wasn't regenerated (see "Gotcha" below).

### 2. Credential secret

```bash
kubectl -n longhorn-system get secret longhorn-backup-credentials -o jsonpath='{.data}' | tr ',' '\n'
```

- **Keyless (AWS, recommended):** exactly one key — `AWS_IAM_ROLE_ARN`, no `AWS_ACCESS_KEY_ID`. NIC resolves the ARN from the EKS Pod Identity association at deploy time.
- **Static keys:** `AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY`, no role ARN.

Keyless only: confirm Pod Identity creds are actually injected into the pods that do the S3 I/O:

```bash
POD=$(kubectl -n longhorn-system get pods -o name | grep instance-manager | head -1)
kubectl -n longhorn-system exec "${POD#pod/}" -- env | grep AWS_CONTAINER_CREDENTIALS_FULL_URI
# expect: AWS_CONTAINER_CREDENTIALS_FULL_URI=http://169.254.170.23/v1/credentials
```

### 3. Schedules were created

```bash
kubectl -n longhorn-system get recurringjobs
kubectl -n longhorn-system get setting allow-recurring-job-while-volume-detached -o jsonpath='{.value}{"\n"}'
```

Expect one RecurringJob per configured schedule (snapshot + backup) with the cron/retain/concurrency from your config.

### 4. A backup actually completes and lands in S3

Backups fire on the `backup` schedule's cron and only run against existing Longhorn volumes (a fresh cluster's foundational services — e.g. the landing-page Redis PVC — provide some). Wait for the next tick, then:

```bash
kubectl -n longhorn-system get backups.longhorn.io \
  -o custom-columns=NAME:.metadata.name,VOL:.status.volumeName,STATE:.status.state
#   want: STATE=Completed

aws s3 ls s3://<bucket>/backupstore/ --recursive | head
#   want: backupstore/volumes/.../backups/backup_*.cfg  + blocks/*.blk
```

### Gotcha: renaming the bucket / reusing a GitOps repo

The BackupTarget is synced from the GitOps repo, which NIC does **not** regenerate once `.bootstrapped` is set. If you change `s3.bucket` (or reuse the same GitOps repo across cluster rebuilds), the live BackupTarget keeps the old bucket while the IAM policy/created bucket move to the new one → `AccessDenied`. Fix: redeploy with `--regen-apps` (or use a fresh GitOps path). Confirm the git manifest matches your config:

```bash
# in the GitOps repo, on the target branch/path:
grep backupTargetURL manifests/storage/longhorn-backup/backuptarget.yaml
```

viniciusdc:
Hey @tylerpotts, took this for a spin on a live AWS EKS cluster. Couple of things.

`tofu plan` fails right away:

```
Error: Unsupported argument
  An argument named "longhorn_backup_pod_identity_enable" is not expected here.
  main.tf:45, in module "eks_cluster"
```

The shim passes `longhorn_backup_pod_identity_enable`, but the module branch it points at (`terraform-aws-eks-cluster@feat/longhorn-backup-bucket`) calls it `enable_longhorn_backup_pod_identity`. The other three backup args match up, so it's just this one that drifted. 

The module follows the `enable_*_pod_identity` naming (same as the LB controller and autoscaler ones), so I think the module name is the canonical one and the shim needs the rename.

I patched the shim locally to get past it, and the feature itself works end to end on keyless Pod Identity: the module creates the bucket and the pod identity association, `BackupTarget/default` goes `available`, and an on-demand backup completes with the blocks landing in S3. No static keys anywhere, really nice.

The other one is more of an existing-cluster thing than something this PR introduced. A full `nic deploy` never gets to the ArgoCD phase here because the node groups roll on every deploy (latest AMI, #410) and the roll fails with `PodEvictionFailure` on the storage and user groups, so tofu errors out before the BackupTarget/RecurringJobs ever render. On an existing cluster the backup config just never lands, I had to apply it by hand to finish the test. Might be worth calling out that the config phase depends on the node roll completing, otherwise it silently no-ops on upgrades.

tylerpotts:
Thanks for the live-cluster run @viniciusdc — both points addressed in e5015a0:

1. **Shim arg renamed** to `enable_longhorn_backup_pod_identity`. Confirmed against `variables.tf` on the module's `feat/longhorn-backup-bucket` branch; the other three `longhorn_backup_bucket_*` args already matched. The module's `enable_*_pod_identity` convention is the canonical one, as you said.

2. **Docs**: added an "Enabling backups on an existing cluster" section to `docs/longhorn-backups.md` calling out that the configuration phase only runs after the infrastructure phase (including node rolls) completes — a `PodEvictionFailure` mid-roll means the backup config silently never lands, and a re-run after resolving the eviction is required. Includes a quick `kubectl get backuptargets,recurringjobs` check to verify it landed.

The node-roll-on-every-deploy behavior itself (https://github.com/nebari-dev/nebari-infrastructure-core/issues/410) is out of scope here, but agreed it makes upgrades fragile for any config-phase feature, not just backups.

viniciusdc:
Hey @tylerpotts, circling back after running this on an EKS cluster (`nic-nebari-aws`). The backup mechanism holds up nicely: keyless Pod Identity backups, incrementals, retention pruning, detached-volume handling, and a cross-cluster restore all worked as expected.

Nothing to change in this PR; just flagging that we recently talked about possibly not covering the dot files present on the user storage since not only most of it is regenerable from JupyterHub, but it also includes unnecessary files like the npm/conda/pixi/pip and other caching files... just commenting here for later follow-up.

PR #446, opened 2026-07-07
Use custom deploy action to test local deployment in CI

marcelovilla:


.github/fixtures/deploy/azure-config.yaml
```diff
@@ -0,0 +1,50 @@
...
+    node_groups:
+      general:
+        instance: Standard_D4_v3
+        min_nodes: 1
+        max_nodes: 3
+        mode: System
+
+      user:
+        instance: Standard_D8_v3
+        min_nodes: 1
```
dcmcand, line 45: **Blocker:** with `general` at `min_nodes: 1` (D4_v3, 4 vCPU) plus `user` at `min_nodes: 1` (D8_v3, 8 vCPU) this fixture needs 12 vCPU minimum, and run 30400098663 failed here with `ErrCode_InsufficientVCPUQuota: left regional vcpu quota 6, requested quota 8`.

`aws-config.yaml` sets its user pool to `min_nodes: 0`; this one does not. Setting `min_nodes: 0` here, choosing a smaller SKU, or raising the eastus quota would all work, but as-is the Azure job is permanently red.

.github/actions/deploy/action.yml
```diff
@@ -0,0 +1,65 @@
...
+    required: false
+    default: ''
+  wait:
+    description: Wait for all Argo CD Applications to be Healthy after deploy.
+    required: false
+    default: 'true'
+  wait-timeout:
+    description: Seconds to wait for Applications to converge.
+    required: false
+    default: '600'
```
dcmcand: **Blocker:** 600s is too short for a cloud deploy and no cloud job overrides it. In run 30400098663 the AWS infrastructure came up cleanly and the job still failed on `Applications did not converge within 600s`, against a `timeout-minutes: 120` budget. Worth either raising the default or setting a much larger value on the three cloud jobs.

.github/actions/deploy/action.yml
```diff
@@ -0,0 +1,65 @@
...
+  destroy:
+    description: Destroy the deployment in the post step when the job ends.
+    required: false
+    default: 'true'
+  force:
+    description: Pass --force to nic destroy so teardown continues past individual resource failures.
+    required: false
+    default: 'true'
```
dcmcand: **Blocker:** defaulting `force` to true means teardown failures are reported as success.

`post.ts:22` always appends `--force`, and in `pkg/nic/destroy.go:150-160` a provider destroy error under `Force` is downgraded to a warning, after which `"Destruction completed successfully"` is sent unconditionally and the function returns `nil`. So one stuck ENI or LoadBalancer-owned security group (both routine on EKS teardown) leaves a cluster, VPC, and NAT gateway running with a green job and no signal anywhere. There is no scheduled reaper in the repo to catch it later.

Suggest not defaulting this to true for the cloud providers, or making `--force` continue past failures but still exit non-zero.

.github/actions/deploy/src/nic.ts
```diff
@@ -0,0 +1,263 @@
...
+        .split("\n")
+        .filter(Boolean)
+        .map((line) => {
+          const [name, sync, health] = line.split(" ");
+          return { name, sync: sync || "Unknown", health: health || "Unknown" };
+        });
+    }
+
+    const notReady = apps.filter((a) => a.health !== "Healthy");
+    if (apps.length > 0 && notReady.length === 0) {
```
dcmcand: **Blocker:** this gate can pass before Argo CD has created the Applications.

There is no expected set or count and no stability window, and the first poll runs immediately (`sleep(10)` is at the bottom of the loop). The one Application guaranteed to exist at that moment is the root app, and Argo CD deliberately excludes child Applications from parent health, so `nebari-root` reports `Healthy` regardless of its children. That is visible in this PR's own failure dump, where `nebari-root Synced Healthy` sits alongside `certificates`, `gateway-config`, and `httproutes` all Degraded. `nic deploy` only waits for cluster nodes and Argo CD's own Deployments (`pkg/argocd/install.go:118,174`), and `WaitForApplication` in `pkg/argocd/applications.go:98` has no callers at all.

What masks it today is incidental: `nic deploy` spends ~100s on load balancer and DNS lookup first, but only `if cfg.Domain != ""`. Every config here happens to set a domain. Drop the domain (a supported config) and the first poll reports `All 1 Applications are Healthy`.

Fix needs an expected app count or name list, a stability window of N consecutive all-healthy polls, or gating on the root app being `Synced` plus child health.

.github/workflows/deployment-tests.yml
```diff
@@ -0,0 +1,290 @@
...
+          GIT_TOKEN: ${{ secrets.NIC_CI_GITOPS_TOKEN }}
+        with:
+          config: .github/fixtures/deploy/local-config.yaml
+          nic-binary: nic
+
+  aws:
+    name: Deploy (AWS)
+    needs: build
+    if: >-
+      github.event_name == 'release' || github.event_name == 'pull_request' ||
```
dcmcand: **Blocker:** `github.event_name == 'pull_request'` appears in all five jobs' `if:` (lines 54, 90, 145, 202, 246), so marking any PR ready for review fires the whole fan-out, not just local. Run 30400098663 confirms it: Local, Existing, AWS, Azure, and Hetzner all executed on a `pull_request` event.

The PR description says the trigger is "to run deployment tests for the local provider", so I think this is just not what was intended.

.github/workflows/deployment-tests.yml
```diff
@@ -0,0 +1,290 @@
...
+          - aws
+          - azure
+          - hetzner
+          - existing
+        default: local
+  release:
+    # published covers both stable and pre-releases
+    types: [published]
+  pull_request:
+    types: [ready_for_review]
```
dcmcand: **Blocker (same item as the `if:` conditions):** no `branches:` filter here, so the five-provider fan-out also fires for PRs against any base branch, not only main. `ci.yml` scopes its `pull_request` trigger with `branches: [main]`.

.github/workflows/deployment-tests.yml
```diff
@@ -0,0 +1,290 @@
...
+          nic-binary: nic
+
+  aws:
+    name: Deploy (AWS)
+    needs: build
+    if: >-
+      github.event_name == 'release' || github.event_name == 'pull_request' ||
+      inputs.provider == 'aws' || inputs.provider == 'all'
+    runs-on: ubuntu-latest
+    environment: aws
```
dcmcand, line 99: **Blocker:** this reads as a gate but currently is not one. The `aws`, `azure`, and `hetzner` environments all have `protection_rules: []` and `deployment_branch_policy: null`, while `release` has `required_reviewers`. Run 30400098663 has a real deployment record against `aws` created from a `pull_request` event with nothing interposed.

Two things worth being precise about, since they change the fix. This is not a fork exposure (forks get no secrets and no `id-token: write`); it is a same-repo-branch exposure, where both the workflow and the vendored `dist/` come from the PR head. And adding environment reviewers would not close it, because `CLOUDFLARE_API_TOKEN` and `NIC_CI_GITOPS_TOKEN` are repository secrets consumed by the two jobs that declare no environment at all. Gating the trigger is what actually fixes this; a `deployment_branch_policy` is still worth adding for the `workflow_dispatch` path.

.github/workflows/deployment-tests.yml
```diff
@@ -0,0 +1,290 @@
...
+    # Only one AWS deploy test should run at a time and triggering a new one should
+    # not cancel the previous one
+    concurrency:
+      group: deploy-test-aws
+      cancel-in-progress: false
+    timeout-minutes: 120
```
dcmcand, line 105: Related to the `force` default: teardown on a cancelled job is capped by GitHub's 5-minute cancellation window, and `nic destroy` measurably took about 14.5 minutes in run 30400098663 (deploy failed 21:47:39, job ended 22:02:14). A job that trips this 120-minute timeout gets force-terminated mid-destroy. Lowering it to leave real headroom would help.

.github/workflows/ci.yml
```diff
@@ -10,6 +10,24 @@ permissions:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Set up Go
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version-file: go.mod
+
+      - name: Run golangci-lint
```
dcmcand, line 26: **Blocker:** splitting lint into its own job is the right structure, but the required status checks on `main` are exactly `Test` and `Build`, so from the moment this merges a lint failure fails only an unrequired job. This needs a branch-protection update landing with the merge.

Worth fixing the wider drift at the same time: `Workflow pins & release config` and `Vulnerabilities` are already unrequired, and this PR adds two more unrequired jobs (`Lint`, `Deploy (local)`).

On a related note, dropping `needs: test` from `build` is *not* a second gate regression, since both are independently required. Its only real cost is that `Deploy (local)` now spins up a kind cluster even when unit tests are failing.

.github/actions/deploy/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "deploy-nic-action",
+  "version": "0.1.0",
+  "private": true,
+  "description": "GitHub Action that deploys a NIC platform from a config file and destroys it when the job ends",
+  "scripts": {
+    "typecheck": "tsc --noEmit",
+    "build": "ncc build src/main.ts -o dist/main --license licenses.txt && ncc build src/post.ts -o dist/post --license licenses.txt"
```
dcmcand: **Blocker:** nothing in CI runs either of these scripts, so neither `build` nor `typecheck` ever executes and nothing verifies that `dist/` still corresponds to `src/`.

I did confirm the committed bundles are currently byte-identical to a fresh `npm ci && npm run build`, which is genuinely good. But the next contributor who edits `src/nic.ts` and forgets to rebuild ships CI that silently runs the old logic, and the drift is invisible in review because `dist/index.js` is 27,897 lines. A job running `npm ci && npm run build && git diff --exit-code -- .github/actions/deploy/dist` would close it, and it is worth having before the action is extracted to its own repo.

.github/workflows/deployment-tests.yml
```diff
@@ -0,0 +1,290 @@
...
+          branch: existing
+          token: ${{ secrets.NIC_CI_GITOPS_TOKEN }}
+
+
+      # Disable Traefik as NIC will deploy its own Envoy gateway LoadBalancer.
+      # Also, pass --no-lb to drop k3d's external proxy container, which is 
+      # unnecessary on a Linux runner where the node IP is directly routable.
+      - name: Create k3d cluster
+        run: |
+          curl -fsSL https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | TAG=v5.9.0 bash
```
dcmcand: The pin checker only matches `uses:`, so anything inside a `run:` block is unpinned by construction. Two instances here: this installer comes from the mutable `main` branch (`TAG=v5.9.0` pins only the binary the script fetches), and line 278 uses the mutable tag `rancher/k3s:v1.36.2-k3s1` rather than a digest. Both run in the same job as steps holding `NIC_CI_GITOPS_TOKEN`.

In a repo that SHA-pins every third-party action this is worth closing. Extending `check-action-pins.sh`'s glob to `.github/actions/**/action.yml` would also future-proof the two new composite actions.

.github/actions/deploy/src/nic.ts
```diff
@@ -0,0 +1,263 @@
...
+  const actual = crypto
+    .createHash("sha256")
+    .update(fs.readFileSync(tarPath))
+    .digest("hex");
+  if (!expected || expected !== actual) {
+    throw new Error(
+      `checksum mismatch for ${tarball} (expected: ${expected || "<missing from checksums.txt>"}, actual: ${actual})`,
+    );
+  }
+  core.info(`checksum verified (${actual})`);
```
dcmcand: `docs/operations/verifying-releases.md:16-19` treats identity-pinned cosign verification as a distinct authenticity step from the checksum, and releases do publish `checksums.txt.sigstore.json` plus provenance. Anything able to serve a tampered tarball can serve a matching `checksums.txt`, so this verifies transport integrity only.

Not urgent, since no workflow here sets `nic-version` so this path is unreached, but it becomes a real gap the moment the action is published for external repos.

.github/actions/deploy/src/post.ts
```diff
@@ -0,0 +1,33 @@
...
+
+// Post step destroys the deployment when the job ends, and runs even when
+// the job failed or was cancelled. Skipped when deploy never started or the
+// caller set destroy: false.
+function post(): void {
+  if (core.getState("deployStarted") !== "true") {
+    core.info("Deploy never started; nothing to destroy.");
+    return;
+  }
+  if (core.getState("destroy") !== "true") {
```
dcmcand: `destroy` round-trips as a raw string (`main.ts:43`) and is compared with `!== "true"` here, so `destroy: "True"`, `"TRUE"`, `yes`, or `1` all skip teardown with only a warning, whereas `wait` uses `core.getBooleanInput` (`main.ts:59`) and fails loudly on a malformed boolean.

The reachable hazard set is narrow and the default path is fine, so this is really about the asymmetry: the input that costs money when misread is the one that fails open.

.github/actions/deploy/src/nic.ts
```diff
@@ -0,0 +1,263 @@
...
+      "kubectl",
+      [
+        "get",
+        "applications.argoproj.io",
+        "-n",
+        "argocd",
+        "-o",
+        'jsonpath={range .items[*]}{.metadata.name}{" "}{.status.sync.status}{" "}{.status.health.status}{"\\n"}{end}',
+      ],
+      { encoding: "utf8", env },
```
dcmcand: No `else` branch here, `res.error` is never inspected, and `res.stderr` is discarded, so a bad kubeconfig, an unreachable API server, a missing CRD, and "not ready yet" are all indistinguishable for the full timeout. The diagnostics dump at the deadline does eventually surface it, just 600 seconds late. A `core.warning` on the first non-zero exit would cost two lines.

.github/actions/deploy/src/nic.ts
```diff
@@ -0,0 +1,263 @@
...
+
+function curl(url: string, dest: string | null, token: string): string {
+  const args = ["-fsSL", "-H", `Authorization: Bearer ${token}`, url];
+  if (dest) args.push("-o", dest);
+  return capture("curl", args);
+}
+
+function releaseArchName(): string {
+  const os = process.platform;
+  const arch = process.arch === "arm64" ? "arm64" : "x86_64";
```
dcmcand: Two gaps on this line and the next: `process.platform` returns `win32` where the release assets say `windows`, and line 70 hardcodes `.tar.gz` where Windows assets are `.zip`. Separately, `process.arch === "arm64" ? "arm64" : "x86_64"` maps every other architecture to `x86_64`, so a non-x64 runner downloads an x86_64 tarball, passes both the checksum and the `isExecutable` check, then fails cryptically at exec time. Rejecting unsupported platforms with a clear message would beat a 404 or a bad exec.

.github/actions/deploy/src/main.ts
```diff
@@ -0,0 +1,70 @@
...
+  const kubeconfig = path.join(
+    process.env.RUNNER_TEMP || "/tmp",
+    `nic-kubeconfig-${process.env.GITHUB_ACTION || "deploy"}`,
+  );
+  run(nic, ["kubeconfig", "-f", config, "-o", kubeconfig]);
+  core.exportVariable("KUBECONFIG", kubeconfig);
+  core.setOutput("kubeconfig", kubeconfig);
+
+  if (core.getBooleanInput("wait")) {
+    const timeout = parseInt(core.getInput("wait-timeout"), 10) || 600;
```
dcmcand: `|| 600` turns an explicit `0` into 600, and `parseInt` also silently truncates `300s` to 300 and accepts a negative value that leaves the deadline already past. `Number.isFinite` plus an explicit range check would cover all three.

.github/actions/deploy/src/main.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  // Save teardown state before deploying so the post step can destroy a
+  // partially created deployment even when `nic deploy` fails mid-way.
+  core.saveState("nicBinary", nic);
+  core.saveState("config", config);
+  core.saveState("destroy", core.getInput("destroy"));
+  core.saveState("force", core.getInput("force"));
+  core.saveState("deployStarted", "true");
+
+  core.startGroup("nic deploy");
```
dcmcand: `run()` throws on a non-zero exit, so a failing `nic deploy` never reaches `endGroup()` and the failure output renders inside a collapsed, never-closed group. Same pattern in `post.ts:24-26`. A `try/finally` would fix both, and it matters most in exactly the case where the log is the only record.

.github/workflows/deployment-tests.yml
```diff
@@ -0,0 +1,290 @@
...
+        run: make build
+
+      - name: Upload binary
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: nic
+          path: nic
+          retention-days: 1
+
+  local:
```
dcmcand, line 49: `local` and `existing` have no `concurrency:` group while the three cloud jobs do, and all five force-reset a shared branch of `nic-ci-gitops`, so two overlapping runs would wipe a branch out from under an in-flight deploy. Low blast radius (the branch is reset every run and these are the two cheap jobs), but it is a copy-paste away.

One thing worth knowing when you add them: with `cancel-in-progress: false` GitHub keeps at most one *pending* run per group, so a third trigger cancels the queued one rather than queueing behind it. The comment at lines 94-95 holds for the in-progress run but not for a queued one.

.github/fixtures/deploy/existing-config.yaml
```diff
@@ -0,0 +1,18 @@
...
+  branch: existing
+  path: "clusters/existing-ci"
+  auth:
+    token_env: GIT_TOKEN
+
+cluster:
+  existing:
+    context: k3d-existing-ci
+    storage_class: local-path
+
```
dcmcand: Nit: trailing-whitespace line and no final newline. The other four fixtures end cleanly.

.github/workflows/deployment-tests.yml
```diff
@@ -0,0 +1,290 @@
...
+
+      - name: Reset gitops branch
+        uses: ./.github/actions/reset-gitops-branch
+        with:
+          branch: existing
+          token: ${{ secrets.NIC_CI_GITOPS_TOKEN }}
+
+
+      # Disable Traefik as NIC will deploy its own Envoy gateway LoadBalancer.
+      # Also, pass --no-lb to drop k3d's external proxy container, which is
```
dcmcand: Nit: trailing whitespace here, and a double blank line at 269-270.

.github/actions/deploy/action.yml
```diff
@@ -0,0 +1,65 @@
...
+  or git ref (nic-version). Exports KUBECONFIG so the steps after this action
+  run against the deployed cluster, and by default waits for all Argo CD
+  Applications to become Healthy.
+
+inputs:
+  config:
+    description: >-
+      Path to the NIC config file, relative to the workspace. When unset, the
+      action deploys its built-in default: a local kind cluster with an
+      auto-created local gitops repository (see default-config.yaml in the
```
dcmcand: Nit: the closing paren is misplaced, so the parenthetical swallows the following sentence: "(see default-config.yaml in the action directory. Copy it into your repo as a starting point for a custom config)." Closing after "directory" reads as intended.

.github/actions/deploy/src/main.ts
```diff
@@ -0,0 +1,70 @@
...
+    binary: core.getInput("nic-binary"),
+    version: core.getInput("nic-version"),
+    token: core.getInput("token"),
+  });
+  run(nic, ["version"]);
+  core.setOutput("nic-binary", nic);
+
+  // Save teardown state before deploying so the post step can destroy a
+  // partially created deployment even when `nic deploy` fails mid-way.
+  core.saveState("nicBinary", nic);
```
dcmcand: Nice: saving teardown state *before* `nic deploy` runs is the detail that makes the post step able to clean up a partially created deployment. I traced every statement preceding this and none of them touch a cloud API, so there is no path that creates infrastructure without recorded teardown state. Combined with no `post-if` on the post step (so it defaults to `always()`), this handles the failure mode that actually leaks money.

.github/actions/reset-gitops-branch/action.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Reset branch to an empty orphan commit
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.token }}
+        REPO: ${{ inputs.repository }}
+        BRANCH: ${{ inputs.branch }}
+      run: |
+        # The empty tree is a well-known git object present in every
+        # repository; an orphan commit on it is a fully empty branch state.
+        tree=$(git hash-object -t tree /dev/null)
```
dcmcand, line 34: Nice: the empty-tree orphan commit is an elegant way to get a guaranteed-empty branch, with a clean fallback for the first run when the ref does not exist. The description explaining *why* this is needed (the bootstrap marker, and why `--regen-apps` was rejected) is the kind of context that saves the next reader a lot of time.

.github/fixtures/deploy/aws-config.yaml
```diff
@@ -0,0 +1,45 @@
+project_name: aws-ci
+domain: aws-ci.nebari.dev
+
+certificate:
+  type: letsencrypt
+  acme:
+    email: ci@nebari.dev
+    server: https://acme-staging-v02.api.letsencrypt.org/directory
```
dcmcand, line 8: Nice: using the Let's Encrypt staging directory across all three cloud fixtures, so CI cannot burn production ACME rate limits. Worth remembering that staging certs chain to an untrusted root if an HTTPS assertion step is ever added.

PR #448, opened 2026-07-08
Fix Gitops file permissions and host path

oldsj:
## Description

I was running into issues getting the cluster to come up locally with `./nic deploy -f examples/local-config.yaml` on mac.

This PR 
- changes the local gitops dir path to ~/.nic/gitops/<project> (this may not strictly be necessary but keeps the path consistent cross platform and in a stable non-tmp location. Happy to drop this change or move the dir somewhere else) see https://github.com/golang/go/issues/21318 and https://github.com/nebari-dev/nebari-infrastructure-core/issues/374
- fixes file permissions on the dir and files so that non-root argo containers can read them
- logs a message on destroy that local gitops repo will remain on disk

## How to Test

```
./nic destroy -f examples/local-config.yaml
./nic deploy -f examples/local-config.yaml

ls -l ~/.nic/gitops/<project> 

```

On destroy nic logs about the remaining local gitops dir if present

```
./nic destroy -f examples/local-config.yaml
...
"The local GitOps directory was left in place and can be removed manually: /Users/jameswork/.nic/gitops/my-nebari-local","resource":"gitops","path":"/Users/jameswork/.nic/gitops/my-nebari-local"
```

Re-deploying with an existing local gitops dir works as expected

```
ls -l ~/.nic/gitops/<project> 

drwxr-xr-x@ 16 jameswork  staff  512 Jul 10 12:23 apps
drwxr-xr-x@  7 jameswork  staff  224 Jul 10 12:23 manifests
-rw-r--r--@  1 jameswork  staff  471 Jul 10 12:23 nic-config.yaml

./nic deploy -f examples/local-config.yaml

<success>
```

Argo can read future writes to the gitops dir by default

```
echo 'test123' >~/.nic/gitops/my-nebari-local/test.txt
kubectl exec -n argocd argocd-repo-server-768b5c5f68-kt6tg -- cat /Users/jameswork/.nic/gitops/my-nebari-local/test.txt

test123
```

You should see the cluster come up and not hang on provisioning a load balancer.
`kubectl get pods -n envoy-gateway-system` shows healthy pods

### DIFF
--- a/docs/local-kind-development.md
+++ b/docs/local-kind-development.md
@@ -24,7 +24,7 @@ The Makefile reads `examples/local-config.yaml` and automatically handles three 
 
 | Config | What happens |
 |--------|-------------|
-| No `git_repository` section | Auto-creates `~/.nebari/gitops/{project_name}` and mounts it into the cluster |
+| No `git_repository` section | Auto-creates `~/.nic/gitops/{project_name}` and mounts it into the cluster |
 | `url: "file:///path/to/repo"` | Mounts that local path into the cluster |
 | `url: "git@github.com:..."` | No mount — ArgoCD pulls from the remote repo directly |
 
--- a/docs/local-kind-development.md
+++ b/docs/local-kind-development.md
@@ -24,11 +24,13 @@ The Makefile reads `examples/local-config.yaml` and automatically handles three 
 
 | Config | What happens |
 |--------|-------------|
-| No `git_repository` section | Auto-creates `/tmp/nebari-gitops-{project_name}` and mounts it into the cluster |
+| No `git_repository` section | Auto-creates `~/.nebari/gitops/{project_name}` and mounts it into the cluster |
 | `url: "file:///path/to/repo"` | Mounts that local path into the cluster |
 | `url: "git@github.com:..."` | No mount — ArgoCD pulls from the remote repo directly |
 
 For local `file://` repos, the path is mounted into both the Kind node and the ArgoCD repo-server pod so ArgoCD can read manifests directly from your filesystem.
+
+If an existing Kind cluster was created with a different local GitOps path, recreate it with `make localkind-down` followed by `make localkind-up`; Kind mounts are fixed at cluster creation time.
 
 > **Note:** `file://` repos only work when the cluster nodes can access the local path (Kind, k3s, bare metal). For cloud providers (AWS, GCP, Azure), use a remote git repository since Kubernetes nodes don't have access to your local filesystem.
 
--- a/docs/local-kind-development.md
+++ b/docs/local-kind-development.md
@@ -25,10 +25,12 @@ The Makefile reads `examples/local-config.yaml` and automatically handles three 
 | Config | What happens |
 |--------|-------------|
 | No `git_repository` section | Auto-creates `~/.nic/gitops/{project_name}` and mounts it into the cluster |
-| `url: "file:///path/to/repo"` | Mounts that local path into the cluster |
+| `url: "file:///path/to/repo"` | Uses the matching `cluster.local.kind.extra_mounts` entry supplied by the user |
 | `url: "git@github.com:..."` | No mount — ArgoCD pulls from the remote repo directly |
 
 For local `file://` repos, the path is mounted into both the Kind node and the ArgoCD repo-server pod so ArgoCD can read manifests directly from your filesystem.
+
+NIC normalizes local GitOps repositories to directory mode `0755` and file mode `0644` so the non-root ArgoCD repo-server can read them. Other `extra_mounts` are user-managed, and NIC does not recursively change their permissions.
 
 If an existing Kind cluster was created with a different local GitOps path, recreate it with `make localkind-down` followed by `make localkind-up`; Kind mounts are fixed at cluster creation time.
 
--- a/pkg/argocd/writer.go
+++ b/pkg/argocd/writer.go
@@ -360,9 +360,6 @@ func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.Nebari
 		if err := os.MkdirAll(filepath.Dir(destPath), git.LocalGitOpsDirMode); err != nil {
 			return fmt.Errorf("failed to create directory for %s: %w", destPath, err)
 		}
-		if err := os.Chmod(filepath.Dir(destPath), git.LocalGitOpsDirMode); err != nil {
-			return fmt.Errorf("failed to set directory permissions for %s: %w", destPath, err)
-		}
 
 		// Write processed content
 		if err := os.WriteFile(destPath, processed, git.LocalGitOpsFileMode); err != nil {
--- a/pkg/argocd/writer.go
+++ b/pkg/argocd/writer.go
@@ -338,10 +338,10 @@ func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.Nebari
 		destPath := filepath.Join(workDir, relPath)
 
 		if d.IsDir() {
-			if err := os.MkdirAll(destPath, git.LocalGitOpsDirMode); err != nil {
+			if err := os.MkdirAll(destPath, git.GitOpsDirMode); err != nil {
 				return err
 			}
-			return os.Chmod(destPath, git.LocalGitOpsDirMode)
+			return os.Chmod(destPath, git.GitOpsDirMode)
 		}
 
 		// Read template content
@@ -357,15 +357,15 @@ func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.Nebari
 		}
 
 		// Ensure parent directory exists
-		if err := os.MkdirAll(filepath.Dir(destPath), git.LocalGitOpsDirMode); err != nil {
+		if err := os.MkdirAll(filepath.Dir(destPath), git.GitOpsDirMode); err != nil {
 			return fmt.Errorf("failed to create directory for %s: %w", destPath, err)
 		}
 
 		// Write processed content
-		if err := os.WriteFile(destPath, processed, git.LocalGitOpsFileMode); err != nil {
+		if err := os.WriteFile(destPath, processed, git.GitOpsFileMode); err != nil {
 			return fmt.Errorf("failed to write %s: %w", destPath, err)
 		}
-		if err := os.Chmod(destPath, git.LocalGitOpsFileMode); err != nil {
+		if err := os.Chmod(destPath, git.GitOpsFileMode); err != nil {
 			return fmt.Errorf("failed to set file permissions for %s: %w", destPath, err)
 		}
 
--- a/pkg/argocd/writer.go
+++ b/pkg/argocd/writer.go
@@ -315,8 +315,10 @@ func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.Nebari
 		destPath := filepath.Join(workDir, relPath)
 
 		if d.IsDir() {
-			// Create directory
-			return os.MkdirAll(destPath, 0750)
+			if err := os.MkdirAll(destPath, git.LocalGitOpsDirMode); err != nil { //nolint:gosec // Generated GitOps manifests must be pod-readable.
+				return err
+			}
+			return os.Chmod(destPath, git.LocalGitOpsDirMode) //nolint:gosec // Generated GitOps manifests must be pod-readable.
 		}
 
 		// Read template content
@@ -332,13 +334,19 @@ func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.Nebari
 		}
 
 		// Ensure parent directory exists
-		if err := os.MkdirAll(filepath.Dir(destPath), 0750); err != nil {
+		if err := os.MkdirAll(filepath.Dir(destPath), git.LocalGitOpsDirMode); err != nil { //nolint:gosec // Generated GitOps manifests must be pod-readable.
 			return fmt.Errorf("failed to create directory for %s: %w", destPath, err)
+		}
+		if err := os.Chmod(filepath.Dir(destPath), git.LocalGitOpsDirMode); err != nil { //nolint:gosec // Generated GitOps manifests must be pod-readable.
+			return fmt.Errorf("failed to set directory permissions for %s: %w", destPath, err)
 		}
 
 		// Write processed content
-		if err := os.WriteFile(destPath, processed, 0600); err != nil {
+		if err := os.WriteFile(destPath, processed, git.LocalGitOpsFileMode); err != nil { //nolint:gosec // Generated GitOps manifests are non-secret and must be pod-readable.
 			return fmt.Errorf("failed to write %s: %w", destPath, err)
+		}
+		if err := os.Chmod(destPath, git.LocalGitOpsFileMode); err != nil { //nolint:gosec // Generated GitOps manifests are non-secret and must be pod-readable.
+			return fmt.Errorf("failed to set file permissions for %s: %w", destPath, err)
 		}
 
 		return nil
--- a/pkg/argocd/writer.go
+++ b/pkg/argocd/writer.go
@@ -315,10 +315,10 @@ func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.Nebari
 		destPath := filepath.Join(workDir, relPath)
 
 		if d.IsDir() {
-			if err := os.MkdirAll(destPath, git.LocalGitOpsDirMode); err != nil { //nolint:gosec // Generated GitOps manifests must be pod-readable.
+			if err := os.MkdirAll(destPath, git.LocalGitOpsDirMode); err != nil {
 				return err
 			}
-			return os.Chmod(destPath, git.LocalGitOpsDirMode) //nolint:gosec // Generated GitOps manifests must be pod-readable.
+			return os.Chmod(destPath, git.LocalGitOpsDirMode)
 		}
 
 		// Read template content
@@ -334,18 +334,18 @@ func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.Nebari
 		}
 
 		// Ensure parent directory exists
-		if err := os.MkdirAll(filepath.Dir(destPath), git.LocalGitOpsDirMode); err != nil { //nolint:gosec // Generated GitOps manifests must be pod-readable.
+		if err := os.MkdirAll(filepath.Dir(destPath), git.LocalGitOpsDirMode); err != nil {
 			return fmt.Errorf("failed to create directory for %s: %w", destPath, err)
 		}
-		if err := os.Chmod(filepath.Dir(destPath), git.LocalGitOpsDirMode); err != nil { //nolint:gosec // Generated GitOps manifests must be pod-readable.
+		if err := os.Chmod(filepath.Dir(destPath), git.LocalGitOpsDirMode); err != nil {
 			return fmt.Errorf("failed to set directory permissions for %s: %w", destPath, err)
 		}
 
 		// Write processed content
-		if err := os.WriteFile(destPath, processed, git.LocalGitOpsFileMode); err != nil { //nolint:gosec // Generated GitOps manifests are non-secret and must be pod-readable.
+		if err := os.WriteFile(destPath, processed, git.LocalGitOpsFileMode); err != nil {
 			return fmt.Errorf("failed to write %s: %w", destPath, err)
 		}
-		if err := os.Chmod(destPath, git.LocalGitOpsFileMode); err != nil { //nolint:gosec // Generated GitOps manifests are non-secret and must be pod-readable.
+		if err := os.Chmod(destPath, git.LocalGitOpsFileMode); err != nil {
 			return fmt.Errorf("failed to set file permissions for %s: %w", destPath, err)
 		}
 
--- a/pkg/argocd/writer_test.go
+++ b/pkg/argocd/writer_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 
 	"github.com/nebari-dev/nebari-infrastructure-core/pkg/config"
+	"github.com/nebari-dev/nebari-infrastructure-core/pkg/git"
 	"github.com/nebari-dev/nebari-infrastructure-core/pkg/providers/cluster"
 )
 
@@ -628,8 +629,21 @@ func TestWriteAllToGit_IncludesRedirectRoute(t *testing.T) {
 	}
 
 	redirectPath := filepath.Join(tmpDir, "manifests", "networking", "routes", "http-to-https-redirect.yaml")
-	if _, err := os.Stat(redirectPath); os.IsNotExist(err) {
+	redirectInfo, err := os.Stat(redirectPath)
+	if os.IsNotExist(err) {
 		t.Error("WriteAllToGit did not write http-to-https-redirect.yaml")
+	} else if err != nil {
+		t.Fatalf("stat redirect route: %v", err)
+	} else if got := redirectInfo.Mode().Perm(); got != git.LocalGitOpsFileMode {
+		t.Errorf("redirect route mode = %v, want %v", got, git.LocalGitOpsFileMode)
+	}
+
+	routesInfo, err := os.Stat(filepath.Dir(redirectPath))
+	if err != nil {
+		t.Fatalf("stat routes directory: %v", err)
+	}
+	if got := routesInfo.Mode().Perm(); got != git.LocalGitOpsDirMode {
+		t.Errorf("routes directory mode = %v, want %v", got, git.LocalGitOpsDirMode)
 	}
 
 	content, err := os.ReadFile(redirectPath) //nolint:gosec // path is t.TempDir() + constant
--- a/pkg/argocd/writer_test.go
+++ b/pkg/argocd/writer_test.go
@@ -730,16 +730,16 @@ func TestWriteAllToGit_IncludesRedirectRoute(t *testing.T) {
 		t.Error("WriteAllToGit did not write http-to-https-redirect.yaml")
 	} else if err != nil {
 		t.Fatalf("stat redirect route: %v", err)
-	} else if got := redirectInfo.Mode().Perm(); got != git.LocalGitOpsFileMode {
-		t.Errorf("redirect route mode = %v, want %v", got, git.LocalGitOpsFileMode)
+	} else if got := redirectInfo.Mode().Perm(); got != git.GitOpsFileMode {
+		t.Errorf("redirect route mode = %v, want %v", got, git.GitOpsFileMode)
 	}
 
 	routesInfo, err := os.Stat(filepath.Dir(redirectPath))
 	if err != nil {
 		t.Fatalf("stat routes directory: %v", err)
 	}
-	if got := routesInfo.Mode().Perm(); got != git.LocalGitOpsDirMode {
-		t.Errorf("routes directory mode = %v, want %v", got, git.LocalGitOpsDirMode)
+	if got := routesInfo.Mode().Perm(); got != git.GitOpsDirMode {
+		t.Errorf("routes directory mode = %v, want %v", got, git.GitOpsDirMode)
 	}
 
 	content, err := os.ReadFile(redirectPath) //nolint:gosec // path is t.TempDir() + constant
--- a/pkg/git/client_impl.go
+++ b/pkg/git/client_impl.go
@@ -3,6 +3,7 @@ package git
 import (
 	"context"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
@@ -263,6 +264,11 @@ func (c *ClientImpl) initLocalPath(ctx context.Context) error {
 	_, span := tracer.Start(ctx, "git.initLocalPath")
 	defer span.End()
 
+	if err := EnsureLocalGitOpsDir(ctx, c.repoPath); err != nil {
+		span.RecordError(err)
+		return err
+	}
+
 	// Ensure working directory exists
 	if c.cfg.Path != "" {
 		if err := os.MkdirAll(c.workDir, LocalGitOpsDirMode); err != nil {
@@ -395,6 +401,54 @@ func (c *ClientImpl) pull(ctx context.Context) error {
 // WorkDir returns the local working directory path.
 func (c *ClientImpl) WorkDir() string {
 	return c.workDir
+}
+
+// NormalizeLocalPermissions normalizes a local file:// GitOps repository tree,
+// including .git, so ArgoCD's non-root repo-server can read refs and generated
+// manifests through a read-only kind hostPath mount. Symlinks are skipped.
+func (c *ClientImpl) NormalizeLocalPermissions(ctx context.Context) error {
+	tracer := otel.Tracer(tracerName)
+	_, span := tracer.Start(ctx, "git.NormalizeLocalPermissions")
+	defer span.End()
+	span.SetAttributes(attribute.String("git.repo_path", c.repoPath))
+
+	if c.cfg == nil || !c.cfg.IsLocalPath() {
+		span.SetAttributes(attribute.Bool("git.local_path", false))
+		return nil
+	}
+	span.SetAttributes(attribute.Bool("git.local_path", true))
+
+	rootDir, err := os.OpenRoot(c.repoPath)
+	if err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("open local gitops root %s: %w", c.repoPath, err)
+	}
+	defer func() {
+		_ = rootDir.Close()
+	}()
+
+	err = fs.WalkDir(rootDir.FS(), ".", func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.Type()&fs.ModeSymlink != 0 {
+			return nil
+		}
+
+		mode := LocalGitOpsFileMode
+		if d.IsDir() {
+			mode = LocalGitOpsDirMode
+		}
+		if err := rootDir.Chmod(path, mode); err != nil {
+			return fmt.Errorf("set permissions on %s: %w", path, err)
+		}
+		return nil
+	})
+	if err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("set local gitops permissions under %s: %w", c.repoPath, err)
+	}
+	return nil
 }
 
 // CommitAndPush stages all changes, commits, and pushes to remote.
--- a/pkg/git/client_impl.go
+++ b/pkg/git/client_impl.go
@@ -265,11 +265,11 @@ func (c *ClientImpl) initLocalPath(ctx context.Context) error {
 
 	// Ensure working directory exists
 	if c.cfg.Path != "" {
-		if err := os.MkdirAll(c.workDir, LocalGitOpsDirMode); err != nil { //nolint:gosec // Local GitOps path must be readable by ArgoCD's non-root repo-server.
+		if err := os.MkdirAll(c.workDir, LocalGitOpsDirMode); err != nil {
 			span.RecordError(err)
 			return fmt.Errorf("failed to create subdirectory %s: %w", c.cfg.Path, err)
 		}
-		if err := os.Chmod(c.workDir, LocalGitOpsDirMode); err != nil { //nolint:gosec // Local GitOps path must be readable by ArgoCD's non-root repo-server.
+		if err := os.Chmod(c.workDir, LocalGitOpsDirMode); err != nil {
 			span.RecordError(err)
 			return fmt.Errorf("failed to set subdirectory permissions %s: %w", c.cfg.Path, err)
 		}
@@ -538,11 +538,11 @@ func (c *ClientImpl) WriteBootstrapMarker(ctx context.Context) error {
 
 	content := fmt.Sprintf("bootstrapped_at: %s\n", time.Now().UTC().Format(time.RFC3339))
 
-	if err := os.WriteFile(markerPath, []byte(content), LocalGitOpsFileMode); err != nil { //nolint:gosec // Bootstrap marker is non-secret and must be pod-readable.
+	if err := os.WriteFile(markerPath, []byte(content), LocalGitOpsFileMode); err != nil {
 		span.RecordError(err)
 		return fmt.Errorf("failed to write bootstrap marker: %w", err)
 	}
-	if err := os.Chmod(markerPath, LocalGitOpsFileMode); err != nil { //nolint:gosec // Bootstrap marker is non-secret and must be pod-readable.
+	if err := os.Chmod(markerPath, LocalGitOpsFileMode); err != nil {
 		span.RecordError(err)
 		return fmt.Errorf("failed to set bootstrap marker permissions: %w", err)
 	}
--- a/pkg/git/client_impl.go
+++ b/pkg/git/client_impl.go
@@ -271,11 +271,11 @@ func (c *ClientImpl) initLocalPath(ctx context.Context) error {
 
 	// Ensure working directory exists
 	if c.cfg.Path != "" {
-		if err := os.MkdirAll(c.workDir, LocalGitOpsDirMode); err != nil {
+		if err := os.MkdirAll(c.workDir, GitOpsDirMode); err != nil {
 			span.RecordError(err)
 			return fmt.Errorf("failed to create subdirectory %s: %w", c.cfg.Path, err)
 		}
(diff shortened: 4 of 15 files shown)


dcmcand (commented):
Thanks for this, James, and welcome to the repo. Clean, well-scoped fix, and the "local cluster hangs on the load balancer" symptom is a genuinely annoying one to track down.

A quick note on how I've labeled the inline comments so nothing reads as heavier than it is:
- **Kudos** are things you did well that I want to call out.
- **Stylistic** comments are non-blocking. For each one you can address it now, open a follow-up issue, or decide it isn't worth it and leave it. Your call. I flag the one exception below.
- No blockers, and no open questions.

What I verified rather than assumed:
- **No secrets leak into the now world-readable (0644) files.** Foundational secrets (Keycloak, Postgres, Redis, ArgoCD OIDC) are created straight against the Kubernetes API and never written to git, and `nic-config.yaml` is scrubbed of auth fields before it's written. The 0644/0755 loosening is safe for this local-dev path. (The permission levels themselves are correct and match what the team agreed on, so nothing to answer there.)
- **`ensureLocalGitOpsPermissions` does what it should:** it chmods the root and the `.git` tree, skips symlinks, and 0644 on git objects does not break ArgoCD's reads.
- Build is clean and the changed-package tests pass.

The one comment I'd weight more heavily than the rest is the package-placement note on `ensureLocalGitOpsPermissions`. It's still technically stylistic, but it's a design/maintainability concern rather than a nitpick: the "make a NIC-managed local repo readable" logic now lives in three packages. I'd encourage either handling it here or opening a follow-up so it doesn't get lost, rather than ignoring it.

Two small test gaps are also worth a look (the walk helper has no direct test, and the `DefaultLocalPath` fallback is uncovered). Good follow-up candidates, not blockers. Nice work overall.

pkg/git/config.go
```diff
@@ -15,15 +15,25 @@ import (
...
 
 // DefaultLocalPath returns the host directory NIC manages for a project's
-// local gitops repository when no git_repository is configured. It is a pure
-// function so the deploy orchestrator and providers that mount the directory
-// (e.g. the local kind provider) can derive the same path independently
-// without threading it between them.
+// local gitops repository when no git_repository is configured. It lives under
+// the user's home directory so the repo is durable and stays on host paths that
+// kind/Docker Desktop can mount reliably.
 func DefaultLocalPath(projectName string) string {
```
dcmcand, line 37: **Kudos:** both `defaultGitConfig` and the kind provider derive the path from this one function, so the home-dir path and the `os.TempDir()` fallback are always computed the same way in both places. That structurally rules out the failure mode where the mount and the writer disagree about which directory to use.

pkg/nic/deploy.go
```diff
@@ -491,17 +508,65 @@ func (c *Client) writeConfigToRepo(ctx context.Context, cfg *config.NebariConfig
...
+// ensureLocalGitOpsPermissions normalizes NIC-managed local GitOps repositories
+// so ArgoCD's non-root repo-server can read refs and generated manifests through
+// a read-only kind hostPath mount.
+func ensureLocalGitOpsPermissions(ctx context.Context, root string) error {
+	tracer := otel.Tracer("nebari-infrastructure-core")
+	_, span := tracer.Start(ctx, "nic.ensureLocalGitOpsPermissions")
+	defer span.End()
+	span.SetAttributes(attribute.String("local_path", root))
+
+	rootDir, err := os.OpenRoot(root)
```
dcmcand: **Kudos:** `os.OpenRoot` + `rootDir.Chmod` is a nice choice here. It's path-traversal-safe, and the symlink skip in the walk keeps you from following links back out of the repo. Clean.

pkg/nic/deploy.go
```diff
@@ -428,6 +432,12 @@ func (c *Client) bootstrapGitOps(ctx context.Context, cfg *config.NebariConfig,
 	}
 
 	if bootstrapped && !regenApps {
```
dcmcand, line 462: **Kudos:** good call running the normalization on the already-bootstrapped skip path too. A repo created by an older nic (0600/0750) gets repaired on the next deploy without forcing a regen, and the two call sites together cover every branch.

pkg/providers/cluster/local/provider.go
```diff
@@ -147,7 +147,7 @@ func (p *Provider) Deploy(ctx context.Context, projectName string, clusterConfig
 		return err
 	}
 	if exists {
-		status.Send(ctx, status.NewUpdate(status.LevelInfo, fmt.Sprintf("Kind cluster %s already exists, reusing it (changes to kind settings such as node_image or extra_mounts only take effect on a recreate)", projectName)).
+		status.Send(ctx, status.NewUpdate(status.LevelInfo, fmt.Sprintf("Kind cluster %s already exists, reusing it (changes to kind settings such as node_image, extra_mounts, or the default local GitOps mount path only take effect on a recreate)", projectName)).
```
dcmcand, line 150: **Kudos:** good UX touch. Calling out that the GitOps mount path only changes on recreate (matching the new note in `docs/local-kind-development.md`) means someone hitting a stale mount after an upgrade gets a pointer instead of a silent failure.

pkg/argocd/writer_test.go
```diff
@@ -628,8 +629,21 @@ func TestWriteAllToGit_IncludesRedirectRoute(t *testing.T) {
 	}
 
 	redirectPath := filepath.Join(tmpDir, "manifests", "networking", "routes", "http-to-https-redirect.yaml")
-	if _, err := os.Stat(redirectPath); os.IsNotExist(err) {
+	redirectInfo, err := os.Stat(redirectPath)
+	if os.IsNotExist(err) {
 		t.Error("WriteAllToGit did not write http-to-https-redirect.yaml")
+	} else if err != nil {
+		t.Fatalf("stat redirect route: %v", err)
+	} else if got := redirectInfo.Mode().Perm(); got != git.LocalGitOpsFileMode {
```
dcmcand: **Kudos:** appreciate that the tests assert the actual mode value, not just that the file exists (same in the git and nic tests). That's what makes these regression-proof.

pkg/nic/deploy.go
```diff
@@ -491,17 +508,65 @@ func (c *Client) writeConfigToRepo(ctx context.Context, cfg *config.NebariConfig
...
+	}
 	status.Send(ctx, status.NewUpdate(status.LevelInfo, "Wrote NIC config to repository (auth fields scrubbed)").
 		WithMetadata("path", configDest))
 	return nil
 }
 
+// ensureLocalGitOpsPermissions normalizes NIC-managed local GitOps repositories
+// so ArgoCD's non-root repo-server can read refs and generated manifests through
+// a read-only kind hostPath mount.
+func ensureLocalGitOpsPermissions(ctx context.Context, root string) error {
```
dcmcand: **Stylistic (but weightier than a nitpick):** the logic for making a NIC-managed local repo readable is now split across three packages: the `LocalGitOps*` constants and chmods in `pkg/git`, the per-file chmods in `pkg/argocd/writer.go`, and this whole-tree walk plus the chmods in `getOrCreateGitConfig`/`writeConfigToRepo` here in `pkg/nic`. Filesystem ownership of the local repo would be easier to reason about in one place. Consider moving this helper into `pkg/git` as a method on the git client, so there's a single owner of "normalize this local repo so the mount can read it." I'd lean toward doing this here or filing a follow-up rather than leaving it, since it tends to get worse quietly as more writers are added.

pkg/providers/cluster/local/kind.go
```diff
@@ -74,18 +74,22 @@ func createKindCluster(ctx context.Context, kp *cluster.Provider, name string, k
...
+	}
 	mounts = append(mounts, v1alpha4.Mount{
 		HostPath:      defaultGitOps,
 		ContainerPath: defaultGitOps,
 		Readonly:      true,
 	})
 
 	for _, m := range kindCfg.ExtraMounts {
-		if err := os.MkdirAll(m.HostPath, 0o750); err != nil {
+		if err := os.MkdirAll(m.HostPath, git.LocalGitOpsDirMode); err != nil {
```
dcmcand: **Stylistic:** the default gitops dir gets the 0755 `MkdirAll` plus a follow-up chmod (and `ensureLocalGitOpsPermissions` later), but `extra_mounts` only get the `MkdirAll`. Since `MkdirAll` is masked by umask, on a restrictive umask (e.g. 077) a fresh extra_mount dir lands at 0700, and because `ensureLocalGitOpsPermissions` is gated on `cfg.GitRepository == nil` it won't touch a user-supplied `file://` repo. So a custom local repo could still hit the exact "non-root ArgoCD can't read it" problem this PR fixes for the default. extra_mounts are user-managed so this may be acceptable, but a short comment or doc note about the asymmetry would save the next person some confusion.

pkg/argocd/writer.go
```diff
@@ -332,14 +334,20 @@ func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.Nebari
 		}
 
 		// Ensure parent directory exists
-		if err := os.MkdirAll(filepath.Dir(destPath), 0750); err != nil {
+		if err := os.MkdirAll(filepath.Dir(destPath), git.LocalGitOpsDirMode); err != nil {
 			return fmt.Errorf("failed to create directory for %s: %w", destPath, err)
 		}
+		if err := os.Chmod(filepath.Dir(destPath), git.LocalGitOpsDirMode); err != nil {
```
dcmcand: **Stylistic (minor/harmless):** this re-creates and re-chmods the parent dir, which was already handled when that directory was visited as its own entry a few lines up, and `ensureLocalGitOpsPermissions` re-chmods the whole tree again afterward on the auto-generated path. Three layers doing the same work. Not a bug, arguably defense-in-depth, just flagging in case you want to trim it.

pkg/nic/deploy.go
```diff
@@ -491,17 +508,65 @@ func (c *Client) writeConfigToRepo(ctx context.Context, cfg *config.NebariConfig
...
+	rootDir, err := os.OpenRoot(root)
+	if err != nil {
+		span.RecordError(err)
+		return fmt.Errorf("open local gitops root %s: %w", root, err)
+	}
+	defer func() {
+		_ = rootDir.Close()
+	}()
+
+	err = fs.WalkDir(rootDir.FS(), ".", func(path string, d fs.DirEntry, err error) error {
```
dcmcand: **Stylistic (test coverage):** this walk is the most intricate new piece and nothing exercises it directly (the `.git` inclusion, the symlink skip, the root `.` chmod, the `OpenRoot` error path). A table-driven test that builds a nested tree with a `.git` dir and a symlink, then asserts the resulting modes, would guard a non-obvious, ArgoCD-critical invariant. As it stands, a future refactor that "helpfully" skipped `.git` would break ArgoCD silently.

pkg/git/config_test.go
```diff
@@ -2,12 +2,26 @@ package git
...
 import (
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
 	"github.com/go-git/go-git/v5/plumbing/transport"
 )
 
+func TestDefaultLocalPath(t *testing.T) {
```
dcmcand, line 17: **Stylistic (test coverage):** this covers the home-dir-available case but `t.Skip`s otherwise, so the `os.TempDir()` fallback branch has no coverage. Testing it cleanly would mean injecting the home lookup, so it's a judgment call whether it's worth it here.

pkg/git/client_impl.go
```diff
@@ -393,6 +399,60 @@ func (c *ClientImpl) WorkDir() string {
...
+// objects the repo-server cannot read. It runs on the write path (after commit),
+// which is the only point new content is produced; we intentionally do not
+// re-walk an already-bootstrapped repo on the skip path.
+func (c *ClientImpl) normalizeLocalPermissions(ctx context.Context) error {
+	tracer := otel.Tracer(tracerName)
+	_, span := tracer.Start(ctx, "git.normalizeLocalPermissions")
+	defer span.End()
+	span.SetAttributes(attribute.String("git.repo_path", c.repoPath))
+
+	if c.cfg == nil || !c.cfg.IsLocalPath() {
```
dcmcand: **Blocker:** since 1aeb246c this gate is any local `file://` repo, not just the NIC-managed auto-generated one, and the walk below force-chmods every non-symlink file in the tree to 0644, `.git` included. For an OPTION 2 user pointing `git_repository.url` at their own pre-existing repo, that has two concrete consequences. A tracked script at 0755 gets rewritten to 0644 on disk after the first commit, and the next commit cycle stages the mode-only diff and bakes `100644` into their history (go-git tracks the executable bit as part of the tree mode). I reproduced this end to end against this branch: the first `CommitAndPush` rewrote the file to 0644, and a second stage-and-commit picked up the mode change and committed it. The walk also strips `+x` from `.git/hooks`, so a user's pre-commit hook silently stops firing with no warning.

Suggested direction: scope the recursive walk to content NIC owns (the auto-generated repo). The files NIC writes into a user-supplied repo already set their own modes in `writer.go` and `writeConfigToRepo`, and `.git` object repair is only needed where NIC's own process created those objects under the user's umask.

pkg/git/client_impl.go
```diff
@@ -393,6 +399,60 @@ func (c *ClientImpl) WorkDir() string {
...
 
+// normalizeLocalPermissions normalizes a local file:// GitOps repository tree,
+// including .git, so ArgoCD's non-root repo-server can read refs and generated
+// manifests through a read-only kind hostPath mount. Symlinks are skipped.
+//
+// This is load-bearing, not defensive: git/go-git create .git contents under the
+// deploying user's umask, so a restrictive umask (e.g. 077) yields 0600/0700
+// objects the repo-server cannot read. It runs on the write path (after commit),
+// which is the only point new content is produced; we intentionally do not
+// re-walk an already-bootstrapped repo on the skip path.
```
dcmcand: **Blocker:** this intent misses the primary way people use a local GitOps repo: hand-editing manifests and letting ArgoCD pick them up (the PR's own test plan does exactly that). A file edited or added by hand under a restrictive umask (corporate `umask 077` is common) lands at 0600. On the next `nic deploy` without `--regen-apps`, `bootstrapGitOps` hits the already-bootstrapped skip path and returns before `CommitAndPush`, the only remaining call site for this walk, so the file is never repaired: the repo-server can't read it and deploy still exits successfully. That's the exact symptom this PR fixes, back through a different door, and it's a regression from the last revision I reviewed, where the skip path repaired stale permissions.

This interacts with my comment on the gate above: scope the walk to NIC-managed repos and it becomes safe to run it on the skip path again, which is roughly the shape the branch had at my first review. To be clear, the comment itself is excellent and unexporting the method was a real cleanup; it's the dropped skip-path repair I'd block on.

docs/local-kind-development.md
```diff
@@ -24,12 +24,16 @@ The Makefile reads `examples/local-config.yaml` and automatically handles three
...
 |--------|-------------|
-| No `git_repository` section | Auto-creates `/tmp/nebari-gitops-{project_name}` and mounts it into the cluster |
-| `url: "file:///path/to/repo"` | Mounts that local path into the cluster |
+| No `git_repository` section | Auto-creates `~/.nic/gitops/{project_name}` and mounts it into the cluster |
+| `url: "file:///path/to/repo"` | Uses the matching `cluster.local.kind.extra_mounts` entry supplied by the user |
 | `url: "git@github.com:..."` | No mount — ArgoCD pulls from the remote repo directly |
 
 For local `file://` repos, the path is mounted into both the Kind node and the ArgoCD repo-server pod so ArgoCD can read manifests directly from your filesystem.
 
+NIC normalizes local GitOps repositories to directory mode `0755` and file mode `0644` so the non-root ArgoCD repo-server can read them. Other `extra_mounts` are user-managed, and NIC does not recursively change their permissions.
```
dcmcand: **Blocker (falls out of the skip-path fix):** as written this reads as an unconditional guarantee, but it only holds on a commit-producing deploy (first bootstrap or `--regen-apps`). A redeploy that hits the already-bootstrapped skip path normalizes nothing. If the skip-path repair comes back, this sentence becomes true as written; otherwise it needs scoping so someone hand-editing the repo doesn't rely on a repair that won't happen. The matching sentence in `examples/local-config.yaml` has the same issue.

pkg/nic/destroy.go
```diff
@@ -156,9 +157,52 @@ func (c *Client) Destroy(ctx context.Context, cfg *config.NebariConfig, opts Des
...
 }
 
+// reportRetainedGitOpsDir logs a reminder that the local GitOps directory is
+// left in place after a destroy so the user knows it exists and where to find
+// it. Cluster teardown does not remove this directory: it may hold local
+// commits or edits the user still wants, and it is cheap to delete manually.
+// Only the NIC-managed local (file://) directory is reported; remote git
+// repositories are the user's to manage. Nothing is logged when the directory
+// no longer exists on disk.
+func reportRetainedGitOpsDir(ctx context.Context, cfg *config.NebariConfig, clusterProvider cluster.Provider) {
```
dcmcand, line 175: **Stylistic:** new `pkg/` function without a span; the convention in this project is `tracer.Start`/`span.RecordError` for new functions, and the siblings added in this same PR (`EnsureLocalGitOpsDir`, `normalizeLocalPermissions`) both follow it. The architecture pass rated the omission design-gating for consistency; I'm leaving it stylistic since the pre-existing `destroyDNS` in this file follows the same span-less pattern, but it's a two-minute add and I'd do it. While you're in here: the `os.Stat` branch treats any error as "directory is gone", including a permission error on an ancestor where the directory still exists. Distinguishing `os.IsNotExist` would make the comment match the behavior.

pkg/git/client_impl.go
```diff
@@ -412,7 +472,15 @@ func (c *ClientImpl) CommitAndPush(ctx context.Context, message string) error {
 	// For local paths, only commit (no push)
 	if c.cfg.IsLocalPath() {
 		span.SetAttributes(attribute.Bool("git.local_path", true))
-		return c.commitLocal(ctx, message)
+		if err := c.commitLocal(ctx, message); err != nil {
+			span.RecordError(err)
+			return err
+		}
+		if err := c.normalizeLocalPermissions(ctx); err != nil {
```
dcmcand: **Stylistic:** `CommitAndPush`'s doc on the `git.Client` interface still reads "stages all changes, commits, and pushes to remote", but for local paths it now also recursively chmods the whole tree, `.git` included, and that applies to user-supplied `file://` repos too. ADR-0001 treats that doc block as the interface contract, so one sentence disclosing the normalization side effect (pointing at the impl comment above, which explains it well) would keep the contract honest.

pkg/nic/deploy_test.go
```diff
@@ -11,6 +14,36 @@ import (
 	"github.com/nebari-dev/nebari-infrastructure-core/pkg/git"
 )
 
+func TestWriteConfigToRepoUsesGitOpsPermissions(t *testing.T) {
```
dcmcand: **Stylistic:** deleting `TestBootstrapGitOpsNormalizesExistingLocalRepository` was the right call given the behavior it pinned was deliberately removed. But once the skip path repairs permissions again (per the blockers), bring back a test shaped like the deleted one so the fix has regression coverage; right now the only artifact describing this behavior boundary is a code comment.

pkg/git/client_impl.go
```diff
@@ -393,6 +399,60 @@ func (c *ClientImpl) WorkDir() string {
 	return c.workDir
 }
 
+// normalizeLocalPermissions normalizes a local file:// GitOps repository tree,
```
dcmcand: **Kudos:** folding normalization into `CommitAndPush` retires the concrete-type escape hatch from the last revision: `pkg/nic` goes through the `git.Client` interface for everything again, and the producer of the umask-masked bytes owns the fix. This doc comment explaining why the walk is load-bearing rather than defensive is exactly the kind of comment that survives refactors.

pkg/nic/destroy.go
```diff
@@ -156,9 +157,52 @@ func (c *Client) Destroy(ctx context.Context, cfg *config.NebariConfig, opts Des
...
+	gitConfig := cfg.GitRepository
+	if gitConfig == nil {
+		// Fall back to the same auto-generated local config deploy uses, but
+		// only for providers that manage a local GitOps directory.
+		if !clusterProvider.InfraSettings(cfg.Cluster).SupportsLocalGitOps {
```
dcmcand, line 184: **Kudos:** gating the fallback on `InfraSettings().SupportsLocalGitOps` and reusing `defaultGitConfig` is exactly how provider-shaped behavior should reach library code: no provider-name switch, and deploy and destroy can't drift on what "the default local repo" means. The notice going through the status channel instead of slog is right too.

docs/local-kind-development.md
```diff
@@ -24,12 +24,16 @@ The Makefile reads `examples/local-config.yaml` and automatically handles three
 
 | Config | What happens |
 |--------|-------------|
-| No `git_repository` section | Auto-creates `/tmp/nebari-gitops-{project_name}` and mounts it into the cluster |
-| `url: "file:///path/to/repo"` | Mounts that local path into the cluster |
+| No `git_repository` section | Auto-creates `~/.nic/gitops/{project_name}` and mounts it into the cluster |
+| `url: "file:///path/to/repo"` | Uses the matching `cluster.local.kind.extra_mounts` entry supplied by the user |
```
dcmcand, line 28: **Kudos:** this table fix corrects a real inaccuracy; the old wording implied an automatic mount that never happens for a custom `file://` path. Same for the `LocalGitOps*` to `GitOps*` rename: those constants already applied to remote repo content, so the old name was misleading before this PR.

pkg/git/client_impl.go
oldsj: Fixed. The repair no longer walks the working tree or assigns `0644`/`0755` to existing paths. It is scoped to the repository root and Git-serving data under `.git`, skips symlinks and private metadata, and ORs in only the required read/traverse bits while preserving executable and special bits. Because tracked files are never chmodded, the mode-only diff from your reproduction cannot occur; hooks are excluded rather than rewritten. The tests cover a tracked executable remaining `0755`, a restrictive working-tree file remaining `0600`, private Git metadata remaining unchanged, and the worktree staying clean after repair.

pkg/git/client_impl.go
oldsj: Addressed, with the repair scoped to what ArgoCD actually reads. The live-cluster check confirmed that an uncommitted working-tree file is not an ArgoCD source: the repo-server resolves `.git` and builds its own checkout. Even so, stale or partially repaired `.git` data must be fixed before the skip path. Local `Init` now performs the additive root/`.git` repair before `IsBootstrapped`, and `CommitAndPush` repeats it after creating new objects. `TestClientInitRepairsExistingLocalRepository` covers an already-bootstrapped repository with restrictive `.git` permissions while asserting that its `0600` working-tree manifest and private hook remain unchanged.

docs/local-kind-development.md
oldsj: Fixed. Both now describe the actual invariant: initialization and local commits repair the repository root and Git-serving data under `.git` for any local `file://` repository. They also state that ArgoCD creates its own checkout and that NIC does not change source working-tree files, hooks, reflogs, the index, or unrelated `extra_mounts`.

pkg/nic/destroy.go
oldsj: Done. `reportRetainedGitOpsDir` now starts a span and records `GetLocalPath` and non-`IsNotExist` stat errors. A permission failure is no longer silently treated as if the directory does not exist.

pkg/git/client_impl.go
oldsj: Done. The `git.Client` contract now documents the additive root/`.git` permission repair for local repositories, including the working-tree/private-metadata exclusions. I also documented the initialization-time repair on `Init`, since that is now part of the local repository contract.

pkg/nic/deploy_test.go
oldsj: Done at the package that owns the behavior. `TestClientInitRepairsExistingLocalRepository` recreates an already-bootstrapped local repository with restrictive root and `.git` modes, reopens it through `Init`, and verifies the Git-serving data is repaired before the skip decision while the working-tree manifest and private hook retain their original modes. The lower-level table test separately covers special-bit preservation, nested private metadata, symlinks, and the missing-root error path.

oldsj:
Thanks for the thorough re-review. The two blockers exposed that the earlier implementation was changing much more of the host repository than ArgoCD actually needs, so I traced the local `file://` path end to end and re-scoped the fix.

The background is that the Kind-mounted repository is read by a non-root ArgoCD repo-server. I verified on a live cluster that ArgoCD resolves the revision from the mounted repository's `.git` data, creates its own checkout under `/tmp/_argocd-repo`, and does not read manifests from the source working tree. The repository root still has to be readable/traversable, and go-git creates loose objects at `0600`, so `.git` remains the load-bearing permission boundary.

The permission repair now lives entirely in `pkg/git` and is limited to the repository root plus Git-serving data under `.git`. It never walks the source working tree. Existing modes are preserved by OR-ing in only group/other read bits (`044`) and, for directories, read/traverse bits (`055`); executable and special bits are never cleared. Symlinks and private checkout metadata such as hooks, reflogs, worktree metadata, the index, and transient Git files are skipped, including in nested repositories under `.git/modules`.

The repair runs for any configured local `file://` repository during `Init`, before the already-bootstrapped skip decision, and again after NIC creates a local commit. That repairs stale Git objects and makes a failed post-commit repair retryable without changing user working-tree files. Separately, newly created, non-secret files and directories that NIC owns request conventional `0644`/`0755` modes, existing working-tree modes are not chmodded.

I tested this on the live local cluster by deploying with restricted source working-tree permissions and confirming ArgoCD still read the committed manifests and deployed the foundational software. Unit coverage now includes fresh and user-supplied local repositories, initialization-time repair of an already-bootstrapped repository, unchanged working-tree and executable modes, preservation of special bits, private Git metadata and symlink exclusions, `.git/objects` readability, error paths, and the `DefaultLocalPath` fallback. `go test ./pkg/git ./pkg/nic`, `make lint`, `make vet`, and `git diff --check` pass.

pkg/git/client_impl.go
```diff
@@ -393,8 +409,138 @@ func (c *ClientImpl) WorkDir() string {
...
+	// (b) Walk the .git subtree, excluding metadata that is private to the
+	// source checkout and is not used to serve commits to ArgoCD.
+	err = fs.WalkDir(rootDir.FS(), ".git", func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		// Match by entry name so nested repositories under .git/modules keep
+		// their private checkout metadata private too.
+		switch d.Name() {
+		case "hooks", "logs", "worktrees":
```
dcmcand: **Blocker.** This matches on `d.Name()` alone, with no check on where the entry sits in the tree, so it over-matches into the refs namespace. A branch named e.g. `hooks/pre-commit-fix` creates `.git/refs/heads/hooks/pre-commit-fix`; the walk hits `.git/refs/heads/hooks` (a directory), matches `case "hooks"`, and `SkipDir`s the whole subtree, so that ref file never gets the group/other read bit. The non-root repo-server then can't read that branch, which shows up as a silent ArgoCD sync error rather than a NIC error. A ref named exactly `index`/`COMMIT_EDITMSG`/`MERGE_MSG`/`ORIG_HEAD`/`FETCH_HEAD` hits the same thing via the second case.

I reproduced it against the branch: `refs/heads/main` was upgraded to 0644 but `refs/heads/hooks/pre-commit-fix` stayed 0600. The default NIC repo uses `main` so it's unaffected; the trigger is an OPTION 2 user-supplied repo with a colliding ref name, which this path explicitly supports. Suggest matching by path position (only `.git/hooks|logs|worktrees` as direct children, and `.git/modules/<name>/...`) rather than the basename anywhere in the tree.

pkg/argocd/writer_test.go
```diff
@@ -724,8 +725,21 @@ func TestWriteAllToGit_IncludesRedirectRoute(t *testing.T) {
...
 
 	redirectPath := filepath.Join(tmpDir, "manifests", "networking", "routes", "http-to-https-redirect.yaml")
-	if _, err := os.Stat(redirectPath); os.IsNotExist(err) {
+	redirectInfo, err := os.Stat(redirectPath)
+	if os.IsNotExist(err) {
 		t.Error("WriteAllToGit did not write http-to-https-redirect.yaml")
+	} else if err != nil {
+		t.Fatalf("stat redirect route: %v", err)
+	} else if got := redirectInfo.Mode().Perm(); got != git.GitOpsFileMode {
+		t.Errorf("redirect route mode = %v, want %v", got, git.GitOpsFileMode)
```
dcmcand: **Blocker.** With the post-write `os.Chmod` gone from `writer.go` (correct: working-tree modes don't matter to ArgoCD now), this assertion's mode is just whatever `WriteFile` produces after the umask is applied. `umask 022` is the macOS/Linux default, so this passes for everyone and in CI, but relying on the default is fragile: under `umask 027` it fails (redirect route 0640 != 0644, routes dir 0750 != 0755, reproduced). More to the point, the assertion now checks an invariant the code intentionally no longer provides. Suggest dropping the exact-mode assertions on these working-tree paths, or pinning the umask in the test if you still want a mode guarantee. Same issue in `deploy_test.go`.

pkg/nic/deploy_test.go
```diff
@@ -11,6 +14,36 @@ import (
...
+	if err := client.writeConfigToRepo(context.Background(), cfg, gitConfig, workDir, ""); err != nil {
+		t.Fatalf("writeConfigToRepo() error: %v", err)
+	}
+
+	dirInfo, err := os.Stat(workDir)
+	if err != nil {
+		t.Fatalf("stat workDir: %v", err)
+	}
+	if got := dirInfo.Mode().Perm(); got != git.GitOpsDirMode {
+		t.Errorf("workDir mode = %v, want %v", got, git.GitOpsDirMode)
```
dcmcand: **Blocker.** Same umask-dependent assertion as in `writer_test.go` (see that comment): under `umask 027` this fails (workDir 0750 != 0755, nic-config.yaml 0640 != 0644). The working-tree mode is no longer an invariant the code guarantees, so this passes only by ambient default. Drop the exact-mode assertions or pin the umask.

pkg/git/client_impl_test.go
```diff
@@ -822,6 +830,533 @@ func TestClientCommitLocalOnFreshRepo(t *testing.T) {
...
+			},
+			check: func(t *testing.T, root string) {
+				t.Helper()
+				// 0o755 | 0o044 = 0o755 — group/other read was already set,
+				// so the OR is a no-op and +x is never at risk of removal.
+				assertPathMode(t, filepath.Join(root, ".git", "hooks-sample.sh"), 0o755)
+			},
+		},
+		{
+			name: "private git metadata left untouched",
```
dcmcand, line 1269: **Stylistic.** This case covers `hooks`/`logs`/`worktrees`/`index` and the nested `modules/example` variants, but `COMMIT_EDITMSG`/`MERGE_MSG`/`ORIG_HEAD`/`FETCH_HEAD` (four of the nine excluded names) are never exercised. Low risk since they share the tested `index` path, but a couple of extra table rows would fully pin this switch given how central it is to the correctness story.

pkg/git/client_impl_test.go
```diff
@@ -822,6 +830,533 @@ func TestClientCommitLocalOnFreshRepo(t *testing.T) {
...
+// TestClientCommitAndPushUpgradesLocalPermissions verifies that CommitAndPush
+// on a local repo makes the repo root and .git (including
+// .git/objects) group/other readable. go-git writes loose objects via a
+// temp-file-then-rename that is always created at 0600 regardless of the
+// process umask, so without this fix a non-root reader (e.g. ArgoCD's
+// repo-server) can't read a fresh commit. It also verifies the upgrade leaves
+// the working tree untouched — argo reads .git and rebuilds its own checkout,
+// so a 0600 working-tree file keeps its mode — and that the upgrade is
+// additive: a tracked script that's already executable keeps exactly its
+// original mode, with no diff staged, so a hand-picked +x bit (e.g. a git
```
dcmcand: **Stylistic.** The script this comment refers to lives in the working tree, which the upgrade never walks, so this assertion is trivially true regardless of the additive-OR logic: it proves "working tree untouched", not "+x survives the OR". The real proof is the ".git file already executable is unchanged" table case. Worth rewording the comment to claim only what the test shows, or adding an executable file directly under `.git` to the integration flow.

pkg/git/client_impl.go
```diff
@@ -393,8 +409,138 @@ func (c *ClientImpl) WorkDir() string {
...
+// exactly as-is. The .git/objects NIC writes are NIC's own commit artifacts,
+// so making them readable is just finishing the job of the commit.
+//
+// Permissions are only ever OR'd in, never replaced: a path that already has
+// the needed bits is left untouched, and special bits such as setgid and sticky
+// are preserved. This cannot strip an existing executable bit or produce a
+// tracked mode-only diff. It runs when a local repository is initialized and
+// after every local commit so stale or partially repaired repositories recover
+// before the already-bootstrapped skip path. Symlinks are skipped.
+func (c *ClientImpl) upgradeLocalPermissions(ctx context.Context) error {
```
dcmcand, line 466: **Stylistic.** `upgradeLocalPermissions` dropped its internal `IsLocalPath()` guard, so the "local `file://` only" precondition now lives implicitly in the three callers (all correctly guarded today). A one-line doc note here saying callers must only invoke this for local repos would make the moved precondition explicit for the next editor.

pkg/git/client_impl_test.go
```diff
@@ -822,6 +830,533 @@ func TestClientCommitLocalOnFreshRepo(t *testing.T) {
...
+	if !status.IsClean() {
+		t.Errorf("existing worktree status after NIC commit = %v, want clean", status)
+	}
+}
+
+// TestClientInitRepairsExistingLocalRepository verifies that Init repairs
+// stale Git-serving permissions before bootstrapGitOps can take its
+// already-bootstrapped skip path. Working-tree and private Git metadata remain
+// untouched.
+func TestClientInitRepairsExistingLocalRepository(t *testing.T) {
```
dcmcand, line 1037: **Kudos:** this is a genuinely good integration test. It recreates a stale, umask-restricted, already-bootstrapped repo and confirms `Init()` alone repairs the `.git`-serving data before the skip decision, with the working-tree manifest and private hook left untouched. That directly closes the skip-path gap from the last round.

pkg/git/client_impl_test.go
oldsj: Done in 8b36a63. The integration flow now creates an executable file directly under .git at 0700 and verifies the additive repair produces 0744, which proves the executable bits are preserved on a path the repair actually walks. The working-tree script assertion and comments now claim only that the working tree remains untouched.

pkg/nic/deploy_test.go
oldsj: Fixed in 8b36a63. I removed TestWriteConfigToRepoUsesGitOpsPermissions entirely: after dropping the invalid exact-mode assertions, the remaining existence-only check did not provide meaningful coverage. The serialization and auth/trust-bundle scrubbing contract is already covered directly by TestScrubbedConfig. The changed packages pass under umask 027.

pkg/git/client_impl_test.go
oldsj: Done in 8b36a63. The private-metadata case now exercises all five excluded files: index, COMMIT_EDITMSG, MERGE_MSG, ORIG_HEAD, and FETCH_HEAD, and confirms each remains 0600 after the repair.

pkg/git/client_impl.go
oldsj: Fixed in 8b36a63. The exclusions are now matched by path position, so private names are skipped only as direct children of .git (or .git/modules/<name>), while colliding names in the refs namespace are still repaired. Added regression coverage for refs/heads/hooks/pre-commit-fix and refs/tags/index, asserting they become 0755/0644 as appropriate.

pkg/git/client_impl.go
oldsj: Done in 8b36a63. The upgradeLocalPermissions doc comment now explicitly states that callers must establish the configured repository is a local file:// path before invoking it.

pkg/argocd/writer_test.go
oldsj: Fixed in 8b36a63. I removed the exact working-tree file/directory mode assertions from this test because those modes are umask-dependent and are no longer part of the ArgoCD-readability contract. The test still verifies that the redirect route is generated and has the expected content. The changed packages pass under both the normal umask and umask 027.

pkg/git/client_impl.go
```diff
@@ -393,8 +410,177 @@ func (c *ClientImpl) WorkDir() string {
...
+	defer func() {
+		_ = rootDir.Close()
+	}()
+
+	// isGitDir reports whether p (a slash-separated path relative to the repo
+	// root) is a Git directory whose direct children may carry private
+	// metadata: the top-level .git or a submodule's .git/modules/<name>.
+	// Keep this as a closure inside the traced operation rather than a separate
+	// uninstrumented package function.
+	isGitDir := func(p string) bool {
```
dcmcand, line 486: **Kudos:** the path-position `isGitDir` closure is exactly the right fix for the refs-namespace over-match. Scoping the private-metadata skip to direct children of `.git` (and one level of `.git/modules/<name>`) means a branch or tag named `hooks/...` or a ref named `index` is repaired normally, while genuine private metadata still stays private. Keeping it as a closure inside the traced op rather than a bare package function is a nice touch too.

pkg/git/client_impl.go
```diff
@@ -393,8 +410,177 @@ func (c *ClientImpl) WorkDir() string {
...
+	// position, not by basename. Matching the basename anywhere in the tree
+	// over-matches into the refs namespace: a branch or tag named e.g.
+	// "hooks/..." creates .git/refs/heads/hooks, and a ref named exactly
+	// "index"/"ORIG_HEAD"/etc. collides too, leaving those refs unreadable to
+	// the non-root repo-server.
+	err = fs.WalkDir(rootDir.FS(), ".git", func(entryPath string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if isGitDir(path.Dir(entryPath)) {
```
dcmcand, line 559: **Kudos:** matching on `path.Dir(entryPath)` position rather than `d.Name()` anywhere in the tree is the crux of the fix, and the comment above it documents precisely why the old basename match was wrong. That's the kind of comment that stops this from regressing.

pkg/git/client_impl_test.go
```diff
@@ -822,6 +827,594 @@ func TestClientCommitLocalOnFreshRepo(t *testing.T) {
...
+				for _, name := range []string{"index", "COMMIT_EDITMSG", "MERGE_MSG", "ORIG_HEAD", "FETCH_HEAD"} {
+					assertPathMode(t, filepath.Join(root, ".git", name), 0o600)
+				}
+				assertPathMode(t, filepath.Join(root, ".git", "modules", "example", "hooks"), 0o700)
+				assertPathMode(t, filepath.Join(root, ".git", "modules", "example", "logs"), 0o700)
+				assertPathMode(t, filepath.Join(root, ".git", "modules", "example", "index"), 0o600)
+			},
+		},
+		{
+			name: "ref names colliding with private metadata are still upgraded",
```
dcmcand, line 1313: **Kudos:** this regression case pins the exact bug from last round - `refs/heads/hooks/pre-commit-fix` and `refs/tags/index` upgraded rather than skipped. Reproduced it green locally.

pkg/git/client_impl_test.go
```diff
@@ -822,6 +827,594 @@ func TestClientCommitLocalOnFreshRepo(t *testing.T) {
...
+		t.Fatalf("failed to write script: %v", err)
+	}
+	// Chmod explicitly so the 0o755 assertion below is umask-independent.
+	if err := os.Chmod(scriptPath, 0o755); err != nil { //nolint:gosec // Deliberately executable to prove the working tree is left untouched.
+		t.Fatalf("failed to chmod script: %v", err)
+	}
+	// An executable file directly under .git IS walked (unlike the working-tree
+	// script above), so it proves the OR adds the missing group/other read bits
+	// without stripping the executable bit: 0o700 | 0o044 = 0o744.
+	gitExecPath := filepath.Join(client.WorkDir(), ".git", "executable-serving-file")
```
dcmcand, line 867: **Kudos:** the `.git/executable-serving-file` (0700 -> 0744) case is the genuine proof the additive OR preserves `+x`, on a path the walk actually touches - much stronger than the old working-tree assertion, which only showed the working tree was left alone.

PR #455, opened 2026-07-10
feat: optional CloudNativePG operator install via database config block

tylerpotts:
## Summary

Adds an opt-in `database:` config block that installs the [CloudNativePG](https://cloudnative-pg.io) operator as a GitOps-managed foundational service, per the discussion in https://github.com/nebari-dev/nebari-infrastructure-core/issues/303. This gives software packs a platform substrate for managed PostgreSQL instead of each pack bundling paywalled Bitnami sub-charts.

```yaml
database:
  enabled: true
  # provider: cloudnativepg   # default and only supported value
```

- **`pkg/config`**: new `DatabaseConfig` (opt-in, default off, `cloudnativepg` the only accepted provider; typos rejected even when disabled) wired into `NebariConfig.Validate`
- **`pkg/argocd`**: static ArgoCD Application template `apps/cloudnative-pg.yaml` (chart `cloudnative-pg` 0.29.0 = operator 1.30.0, namespace `cnpg-system`, sync-wave 3, `ServerSideApply=true` because the CNPG CRDs overflow client-side apply), gated by an `isDatabasePath` skip predicate in `WriteAllToGit`, mirroring the trust-manager pattern
- **Mechanism choice**: ADR-0006 prescribes provider-driven Helm only for software conditional on provider/cluster state; this toggle is user config and provider-agnostic, so it stays on the GitOps path
- **Deliberate teardown semantics** (documented in the example config, `DatabaseConfig` doc comment, and design spec, and pinned by a test): `enabled: false` only prevents installation on a fresh bootstrap and never uninstalls an installed operator. Uninstalling means deleting `apps/cloudnative-pg.yaml` from the gitops repo, which prunes the operator and its CRDs and destroys every CNPG-managed database. A config toggle must not be able to destroy databases.
- Design spec and implementation plan included under `docs/superpowers/`

## Test Plan

- [x] `go test -short ./...` green (new table-driven config tests; writer gating tests incl. enabled/disabled/absent and a disable-never-deletes pinning test; template validity/pinning assertions)
- [x] `golangci-lint run`, `gofmt -l`, `go vet` clean
- [x] Chart coordinates verified against the live repo index (`cloudnative-pg` 0.29.0 -> appVersion 1.30.0, latest release)
- [ ] Manual validation on a Hetzner cluster (steps below)

## Validation steps (Hetzner)

Prerequisite: this feature is delivered through GitOps, and Hetzner does not support local gitops directories, so the config **must** have a remote `git_repository` block (with working auth env vars). Without one, NIC skips the GitOps bootstrap and no ArgoCD apps (including this one) are installed.

**1. Config validation catches bad input**

```bash
make build
# Start from your Hetzner config (see examples/hetzner-config.yaml) and add:
#   database:
#     enabled: true
#     provider: not-a-real-provider
./nic validate -f hetzner-config.yaml
# EXPECT: failure: invalid database provider "not-a-real-provider", must be: cloudnativepg

# Fix it: drop the provider line (defaults to cloudnativepg) or set it explicitly
./nic validate -f hetzner-config.yaml
# EXPECT: validation passes
```

**2. Deploy with the toggle on**

```bash
export HETZNER_TOKEN=...        # plus your git auth env var(s), e.g. GIT_SSH_PRIVATE_KEY
./nic deploy -f hetzner-config.yaml
```

If this cluster's gitops repo was already bootstrapped before this branch, add `--regen-apps` — a plain deploy skips manifest generation and the app will not appear.

**3. The app file lands in the gitops repo**

Pull your gitops repo and confirm `apps/cloudnative-pg.yaml` exists (under `git_repository.path`, if set), pinned to chart `0.29.0`.

**4. The Application syncs and the operator comes up**

```bash
./nic kubeconfig -f hetzner-config.yaml -o /tmp/nic-hetzner-kubeconfig
export KUBECONFIG=/tmp/nic-hetzner-kubeconfig

kubectl get application cloudnative-pg -n argocd
kubectl wait --for=jsonpath='{.status.health.status}'=Healthy application/cloudnative-pg -n argocd --timeout=300s

kubectl rollout status deployment/cloudnative-pg -n cnpg-system --timeout=300s
kubectl get crds | grep postgresql.cnpg.io
# EXPECT: backups, clusters, poolers, scheduledbackups (+ others) .postgresql.cnpg.io
```

**5. The operator actually provisions a database (smoke test)**

The end-state consumer of this substrate is the NebariApp CRD (`spec.database`, issue https://github.com/nebari-dev/nebari-infrastructure-core/issues/303), but the NebariApp schema at the operator version NIC deploys (`v0.1.0-alpha.20`, and nebari-operator `main` as of this PR) has no `database` field yet, so a NebariApp-based smoke test is not possible today. Until that lands, exercise the operator with a raw CNPG `Cluster`, which is exactly what the future NebariApp reconciler will create:

```bash
kubectl apply -f - <<'YAML'
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
metadata:
  name: smoke-test
spec:
  instances: 1
  storage:
    size: 1Gi
YAML
# Uses the cluster default StorageClass (longhorn on Hetzner unless disabled)
kubectl wait --for=condition=Ready pod/smoke-test-1 --timeout=300s
kubectl exec smoke-test-1 -- psql -U postgres -c 'select version();'
# EXPECT: PostgreSQL version banner
kubectl delete cluster smoke-test
```

Once nebari-operator ships `spec.database`, the equivalent smoke test becomes (shape from the issue, not runnable yet):

```yaml
apiVersion: reconcilers.nebari.dev/v1
kind: NebariApp
metadata:
  name: smoke-test
spec:
  database:
    enabled: true
    provider: cloudnativepg
    instances: 1
    size: 1Gi
# then verify the operator creates the CNPG Cluster and the
# smoke-test-db-credentials secret (host/port/username/password/database/uri)
```

**6. Disable is install-only (never uninstalls)**

```bash
# Flip enabled: false in the config, then:
./nic deploy -f hetzner-config.yaml --regen-apps
```

Pull the gitops repo: `apps/cloudnative-pg.yaml` must still be present, and `kubectl get application cloudnative-pg -n argocd` must still show Healthy. Regen skips writing but never deletes (pinned by `TestWriteAllToGit_Database/disable_after_enable_leaves_committed_file_untouched`).

**7. Gating for fresh bootstraps (optional on Hetzner)**

A config without the `database:` block on a fresh bootstrap must not write the app file or install anything. This is covered by unit tests (`TestWriteAllToGit_Database` skipped-when-absent/disabled) and is cheaper to reproduce on kind (`make localkind-up`) than with a second Hetzner cluster.

**8. Teardown**

```bash
./nic destroy -f hetzner-config.yaml
```

## Follow-ups (out of scope here)

- NebariApp CRD `database:` support lives in nebari-operator (the consumer of these CRDs)
- NIC's own foundational `postgresql.yaml` (Keycloak's DB) still pulls the frozen Bitnami chart from `charts.bitnami.com`; deserves its own issue, this PR builds the escape route
- Consider amending ADR-0006 to record the third category explicitly (user-config-gated, provider-agnostic -> GitOps + writer skip predicate; instances: trust-manager, cloudnative-pg)
- Bidirectional disable semantics (regen removing skip-predicated apps) would apply to trust-manager/MetalLB too and was deliberately not implemented


pkg/config/database.go
```diff
@@ -0,0 +1,49 @@
...
+	Enabled bool `yaml:"enabled"`
+
+	// Provider selects the database operator. Empty defaults to
+	// "cloudnativepg", the only supported value today.
+	Provider string `yaml:"provider,omitempty"`
+}
+
+// EffectiveProvider returns the configured provider, defaulting to
+// DatabaseProviderCloudNativePG when unset. Nil-safe.
+func (d *DatabaseConfig) EffectiveProvider() string {
```
dcmcand: **Stylistic:** `EffectiveProvider()` has no caller outside tests. `NewTemplateData` in `writer.go` gates on `cfg.Database != nil && cfg.Database.Enabled` and never reads the provider, and the Application template is static regardless of the value. Since `cloudnativepg` is the only accepted value, this is speculative generality for a second provider that does not exist yet. If the toggle survives at all, wire it into a real call site when a second provider lands or drop it until then. Moot if CNPG becomes foundational (the whole config type goes away).

pkg/config/database.go
```diff
@@ -0,0 +1,49 @@
...
+// delete it from the GitOps repo. Deleting apps/cloudnative-pg.yaml (under
+// git_repository.path, if set) from the repo is what uninstalls the
+// operator, and with it the CNPG CRDs, which deletes every CNPG-managed
+// database and its data.
+type DatabaseConfig struct {
+	Enabled bool `yaml:"enabled"`
+
+	// Provider selects the database operator. Empty defaults to
+	// "cloudnativepg", the only supported value today.
+	Provider string `yaml:"provider,omitempty"`
```
dcmcand: **Stylistic:** the field name `Provider` overloads the term the codebase reserves for cloud providers (aws/gcp/azure/local/hetzner). `CertificateConfig` uses `Type` for the analogous enum-selector role. Worth renaming if the field survives.

pkg/argocd/cloudnative_pg_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	source, _ := spec["source"].(map[string]any)
+	if source["chart"] != "cloudnative-pg" {
+		t.Errorf("chart = %v, want cloudnative-pg", source["chart"])
+	}
+	if source["repoURL"] != "https://cloudnative-pg.github.io/charts" {
+		t.Errorf("repoURL = %v, want the cloudnative-pg chart repo", source["repoURL"])
+	}
+	if rev, _ := source["targetRevision"].(string); rev == "" {
+		t.Error("targetRevision must pin a chart version")
```
dcmcand: **Stylistic:** this test is named `PinsChartAndTarget`, but the version check only asserts `targetRevision != ""`. A typo, a wrong version, or an unpinned `*` wildcard would all pass. The chart, repoURL, and namespace checks a few lines above assert exact values; the version deserves the same `== "0.29.0"` treatment so the test delivers what its name promises.

pkg/argocd/writer.go
```diff
@@ -330,6 +335,14 @@ func WriteAllToGit(ctx context.Context, gitClient git.Client, cfg *config.Nebari
 			return nil
 		}
 
+		// Skip the database-operator app unless a database block is enabled
+		if isDatabasePath(relPath) && !data.DatabaseEnabled {
```
dcmcand: **Stylistic (harmless):** the `d.IsDir()` branch inside this skip block is unreachable. `isDatabasePath` matches only the single file `apps/cloudnative-pg.yaml`, which `fs.WalkDir` never presents as a directory entry, so `fs.SkipDir` can never fire here. It is copy-pasted from the `isMetalLBPath`/`isTrustBundlePath` blocks, which do match directory prefixes, so it is consistent with its siblings; flagging only so a future reader does not assume the branch is load-bearing.

pkg/argocd/templates/apps/cloudnative-pg.yaml
```diff
@@ -0,0 +1,52 @@
...
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
+      allowEmpty: false
+    syncOptions:
+      - CreateNamespace=true
+      # CNPG CRDs are large enough to overflow the client-side
+      # last-applied-configuration annotation; server-side apply is required.
+      - ServerSideApply=true
```
dcmcand, line 46: **Kudos:** correct use of `ServerSideApply=true` with an accurate rationale (the CNPG CRDs overflow the client-side last-applied annotation) and a test that asserts it. Easy to get wrong, and the comment means the next person will not silently drop it.

pkg/argocd/cloudnative_pg_test.go
```diff
@@ -0,0 +1,129 @@
...
+		if !strings.Contains(string(got), "chart: cloudnative-pg") {
+			t.Errorf("rendered app missing chart reference, got:\n%s", got)
+		}
+	})
+
+	// Freeze the deliberate contract that the skip predicate only prevents
+	// writing: a rerun with the toggle off must not delete a previously
+	// written app file. Deleting it would let a config toggle cascade into
+	// CNPG CRD removal and destroy every managed database.
+	t.Run("disable after enable leaves committed file untouched", func(t *testing.T) {
```
dcmcand: **Kudos:** pinning 'disable never deletes the committed file' with a dedicated test and a comment explaining why (a config toggle must never cascade into CNPG CRD deletion) is exactly how a dangerous invariant should be frozen. If the toggle goes away this test goes with it, but the instinct is the right one.

tylerpotts:
Reworked per your review: CNPG is now unconditional foundational infrastructure, emitted the same way postgresql.yaml and keycloak.yaml are. The DatabaseConfig type, writer gating, and disable-never-deletes machinery are gone; the branch was rebuilt as a single commit (template + tests) and force-pushed, so the diff is now 2 files / ~130 lines.

On the inline notes: the pinning test now asserts targetRevision == "0.29.0" exactly. The rest (EffectiveProvider, the Provider field name, the unreachable SkipDir branch) went away with the toggle.

Your caveat about the prune hazard is now stated in the PR body so nobody reads foundational as safer than it is. Keycloak cutover noted as the follow-up PR.

